PR #36, opened 2018-12-03
Remove maven publish plugin from sample modules.

zach-klippenstein:
Blocking 0.4.0 release.


zach-klippenstein:
I'm an idiot, these are supposed to be published in 0.4.0.

PR #39, opened 2018-12-03
Add configuration to publish viewbuiler artifacts.

zach-klippenstein:
PR #36 was bad. This should be the last blocker to 0.4.0.

### DIFF
--- a/viewbuilder-android/build.gradle
+++ b/viewbuilder-android/build.gradle
@@ -15,6 +15,10 @@
  */
 apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
+apply plugin: 'com.vanniktech.maven.publish'
+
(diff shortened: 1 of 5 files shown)


viewbuilder-android/build.gradle
```diff
@@ -15,6 +15,10 @@
  */
 apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
+apply plugin: 'com.vanniktech.maven.publish'
+
+sourceCompatibility = JavaVersion.VERSION_1_7
```
rjrjr: Out of scope, but should we be bumping this to 1_8?

rjrjr: I guess we have no real reason to do so.

zach-klippenstein: Yea, probably.

PR #43, opened 2018-12-05
Introduce a coroutine builder-style function for workflows, and implement Reactor in terms of it.

zach-klippenstein:
This allows different implementations of workflows besides the state-value-for-loop that is `Reactor`. For example, a unit test could create a stub workflow that sends a specific sequence of states.


zach-klippenstein:
TODO: Unit tests

workflow-core/src/main/java/com/squareup/workflow/Reactor.kt
```diff
@@ -179,19 +182,57 @@ interface Reactor<S : Any, E : Any, out O : Any> : WorkflowPool.Launcher<S, E, O
...
+ * Use this only to create a top-level workflow. If you're implementing [Launcher.launch], use
+ * [doLaunch].
  */
-fun <S : Any, E : Any, O : Any> Reactor<S, E, O>.doLaunch(
+fun <S : Any, E : Any, O : Any> Reactor<S, E, O>.startRootWorkflow(
   initialState: S,
-  workflows: WorkflowPool,
   context: CoroutineContext = EmptyCoroutineContext
-): Workflow<S, E, O> = ReactorWorkflow(this, initialState, workflows, context)
+): Workflow<S, E, O> = doLaunch(initialState, WorkflowPool(), context)
```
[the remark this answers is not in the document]
rjrjr: (This is why I ❌'d instead of ✔️'d)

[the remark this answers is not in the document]
zach-klippenstein: You don't. This helper doesn't know anything about `WorkflowPool`. `WorkflowPool` is only a `Reactor`-level concept, it assists with writing reactors. All the work it does to provide the next state (but only the next one), that doesn't make sense here, it's only needed because reactors see the world one state at a time.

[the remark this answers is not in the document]
rjrjr: Sorry, confused myself w/the reordering in here.

You know what? startRootWorkflow() is garbage, used only in the integration test. We should get rid of it before it causes harm. I'll get that.

zach-klippenstein:
This PR causes `ShellReactor` to get into an infinite loop when trying to start a new game. not sure why, but think it has something to do with using `Unconfined` dispatchers for everything and the workflow coroutine now emitting its `FinishWith` via returning from `async` instead of explicitly completing a `CompletableDeferred`.

There are 4 things going on:

1. The ordering of “state channel complete” and “result complete” is undefined.
2. There is effectively a race between a parent receiving the signal that its child has finished, and the cleanup code in `WorkflowPool` removing the child from the pool.
3. `ShellReactor` hits a special case where it immediately starts a new child as soon as the previous one finishes and *both children are the same type*.
4. We’re using the `Unconfined` dispatcher, so all signals are delivered synchronously.

Before, we explicitly completed the result before returning from the channel builder, so the completion order was [result,channel]. Now, we return from the async builder to complete the result, and then close the state channel as a side effect of emitting result. However, emitting result actually has another side effect – `WorkflowPool` uses the same completion handler mechanism to remove the workflow from the map.

Since we’re using the `Unconfined` dispatcher everywhere, when the first completion handler fires, closing the state channel, it triggers the `nextDelegateResult` call to emit, which causes `ShellReactor` to restart the game, which in turn calls `nextDelegateResult` again to ask for the (expected-to-be new) child workflow’s (first) state.

However, since this is all happening synchronously, the first completion handler hasn’t returned yet, so the second completion handler (which removes the workflow from the map), hasn’t run. So when the re-entrant call to `nextDelegateResult` checks if a child workflow of the given type exists, it gets the completed workflow from the map.

Ultimately I think the root cause here is that there’s a race between delivering the result to the parent and seeing the child as “finished”. I knew that race was there, but didn’t expect it to show up like this (figured it would only bite us if we tried using multiple threads). I believe the fix is to change `WorkflowPool` to have the following contract: a child workflow will be removed from the pool only _after its result is requested_ (via one of the `await*` methods), and it will be removed from the pool _before its result is delivered_.

workflow-core/src/main/java/com/squareup/workflow/CoroutineWorkflow.kt
```diff
@@ -0,0 +1,82 @@
...
+  return object : Workflow<S, E, O>, Deferred<O> by result {
+    override fun openSubscriptionToState(): ReceiveChannel<S> = state.openSubscription()
+
+    override fun sendEvent(event: E) {
+      // The events channel has an unlimited-size buffer, this will only fail
+      // if the event channel was closed.
+      try {
+        events.offer(event)
+      } catch (e: ClosedSendChannelException) {
+        // This may mean the workflow was canceled or finished, or that the workflow closed the
```
rjrjr: I wonder how often we'll find ourselves crashing due to late events being aimed at a finished workflow. Hopefully those will be actual bugs, not something we'll need to become more forgiving about. (Not suggesting any changes, just chatting.)

workflow-core/src/main/java/com/squareup/workflow/Reactor.kt
```diff
@@ -179,19 +182,59 @@ interface Reactor<S : Any, E : Any, out O : Any> : WorkflowPool.Launcher<S, E, O
 }
 
 /**
- * Use this to implement [WorkflowPool.Launcher.launch].
+ * Use this only to create a top-level workflow. If you're implementing [Launcher.launch], use
+ * [doLaunch].
  */
-fun <S : Any, E : Any, O : Any> Reactor<S, E, O>.doLaunch(
+fun <S : Any, E : Any, O : Any> Reactor<S, E, O>.startRootWorkflow(
```
rjrjr: This method is bullshit, will delete later (#44).

workflow-core/src/main/java/com/squareup/workflow/WorkflowPool.kt
```diff
@@ -176,7 +175,7 @@ class WorkflowPool {
     name: String,
     type: Type<I, Nothing, O>
   ): O {
-    register(worker.asReactor(), type)
+    register(worker.asLauncher(), type)
```
rjrjr: Happy name change.

workflow-core/src/main/java/com/squareup/workflow/Reactor.kt
zach-klippenstein: #49

PR #48, opened 2018-12-06
Ensure all workflow coroutines are named, to help with debugging.

zach-klippenstein:


zach-klippenstein:
Merged this into another PR branch instead of master because I was in a hurry. Replaced by #70.

PR #72, opened 2018-12-09
Fix workflow operators

### LINKED ISSUE #69: switchMapState isn't actually a switchMap, it's a concatMap

### LINKED ISSUE #71: Result of mapResult operator doesn't propagate cancellation upstream
If you do something like `source.mapResult { it }.cancel()`, `source` will not be cancelled.

zach-klippenstein:
* 9c879cd Fix #69 by correctly implementing switchMapState operator.
* e304784 Fix #71 by cancelling upstream workflow when mapResult workflow is cancelled.
* 1876788 Move the workflow operators into their own file and write unit tests.

### DIFF
--- a/workflow-core/src/main/java/com/squareup/workflow/WorkflowOperators.kt
+++ b/workflow-core/src/main/java/com/squareup/workflow/WorkflowOperators.kt
@@ -88,9 +88,20 @@ fun <S1 : Any, S2 : Any, E : Any, O : Any> Workflow<S1, E, O>.switchMapState(
 fun <S : Any, E : Any, O1 : Any, O2 : Any> Workflow<S, E, O1>.mapResult(
   transform: suspend (O1) -> O2
 ): Workflow<S, E, O2> {
+  // We can't just make the downstream a child of the upstream workflow to propagate cancellation,
+  // since the downstream's call to `await` would never return (parent waits for all its children
+  // to complete).
   val transformedResult = GlobalScope.async(operatorContext) {
     transform(this@mapResult.await())
   }
+
+  // Propagate cancellation upstream.
+  transformedResult.invokeOnCompletion { cause ->
+    if (cause != null) {
+      this@mapResult.cancel(cause)
+    }
+  }
+
   return object : Workflow<S, E, O2>,
       Deferred<O2> by transformedResult,
       WorkflowInput<E> by this {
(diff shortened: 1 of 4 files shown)


workflow-core/src/main/java/com/squareup/workflow/WorkflowOperators.kt
```diff
@@ -0,0 +1,100 @@
...
+import kotlinx.coroutines.experimental.GlobalScope
+import kotlinx.coroutines.experimental.async
+import kotlinx.coroutines.experimental.channels.Channel.Factory.CONFLATED
+import kotlinx.coroutines.experimental.channels.ReceiveChannel
+import kotlinx.coroutines.experimental.channels.consumeEach
+import kotlinx.coroutines.experimental.channels.produce
+import kotlin.coroutines.experimental.CoroutineContext
+
+/**
+ * [CoroutineContext] used by [Workflow] operators below.
```
rjrjr: Could explain why you chose `Unconfined`. (Follow up for nits is fine w/me.)

workflow-core/src/main/java/com/squareup/workflow/WorkflowOperators.kt
```diff
@@ -73,10 +76,36 @@ fun <S1 : Any, S2 : Any, E : Any, O : Any> Workflow<S1, E, O>.switchMapState(
...
-        transform(rawState).consumeEach { transformedState ->
-          send(transformedState)
+      val upstreamChannel = this@switchMapState.openSubscriptionToState()
+      var transformedChannel: ReceiveChannel<S2>? = null
+      val downstreamChannel = channel
+
+      upstreamChannel.consume {
+        whileSelect {
+          upstreamChannel.onReceiveOrNull { upstreamState ->
+            if (upstreamState == null) {
```
rjrjr: I wonder if both of these would be more readable flipped and converted to use `?.let / ?:`.

zach-klippenstein: That feels like code golf to me. `let` improves readability when the nullable value is being transformed, in this case it's not, it's being used as a sentinel for a completely different signal (channel close). This isn't ideal obviously and there's an open issue to make a more explicit receive call that doesn't use null.

workflow-core/src/main/java/com/squareup/workflow/WorkflowOperators.kt
zach-klippenstein: https://github.com/square/workflow/issues/74

Issue #81, opened 2018-12-11
kdoc and sample code needed for Worker


rjrjr:
AC

- [x] `rx2.Reactor` kdoc covers them as well
- [x] core `Reactor` kdoc covers `Worker`
- [x] Sample code uses `rx2.asWorker` and `rx2.workerResult`

zach-klippenstein:
@rjrjr There is _some_ sample code using it already: https://github.com/square/workflow/blob/730d356c52d491ddbb06e74da98a931db70afdbb/sample-game-common/src/main/java/com/squareup/sample/tictactoe/RunGameReactor.kt#L67
https://github.com/square/workflow/blob/730d356c52d491ddbb06e74da98a931db70afdbb/sample-game-common/src/main/java/com/squareup/sample/tictactoe/RunGameReactor.kt#L115
Should we add a more complex sample? The

rjrjr:
Nah, that was the main thing I didn't want to forget about.

zach-klippenstein:
Rx2 `Reactor` kdoc covers workers as well:
![image](https://user-images.githubusercontent.com/101754/50368919-9c504f00-0542-11e9-817b-0c9ddcf55fcf.png)

Issue #79, opened 2018-12-10
rx2.Reactor needs kdoc


rjrjr:
AC: `rx2.Reactor` is documented well enough that it can be used without reference to the base coroutine `Reactor` interface.

PR #82, opened 2018-12-11
More class kdcoc on both Reactor interfaces.

rjrjr:
Adds missing kdoc for `rx2.Reactor`. Adds more meat to `WorkflowPool` doc for
core `Reactor`. Addresses #79 

Still lacking proper usage docs, samples for `Worker`, but that's tracked
separately. (#81)


workflow-core/src/main/java/com/squareup/workflow/Reactor.kt
```diff
@@ -114,11 +114,37 @@ import kotlin.coroutines.experimental.EmptyCoroutineContext
...
+ *    class MyReactor(
+ *      private val nestedReactor : NestedReactor
+ *    ) {
+ *      override fun launch(
+ *        initialState: MyState,
+ *        workflows: WorkflowPool
+ *      ) : Workflow<MyState, MyEvent, MyResult> {
+ *        workflows.register(nestedReactor)
+ *        return doLaunch(initialState, workflows)
+ *      }
```
zach-klippenstein: Nit: missing closing brace

rjrjr: That was intentional, since the class is incomplete. Confusing?

PR #84, opened 2018-12-11
Greatly simplify the implementation of switchMapState.

zach-klippenstein:


workflow-core/src/main/java/com/squareup/workflow/WorkflowOperators.kt
```diff
@@ -18,24 +18,22 @@ package com.squareup.workflow
...
 import kotlin.coroutines.experimental.CoroutineContext
 
 /**
- * [CoroutineContext] used by [Workflow] operators below.
+ * Defines the [CoroutineContext] used by [Workflow] operators below.
  *
  * See this module's README for an explanation of why Unconfined is used.
  */
-private val operatorContext: CoroutineContext = Unconfined
+private val operatorScope = CoroutineScope(Unconfined)
```
rjrjr: What's the significance of the change from `GlobalScope` to our own `CoroutineScope()`?

zach-klippenstein: Just changes how we use it for the builders below.
```kotlin
GlobalScope.foo(operatorContext) {
```
becomes
```
operatorScope.foo {
```

rjrjr: Right, but we're building a scope from scratch instead of using that magic singleton. `GlobalScope` is the same as `CoroutineScope()`?

rjrjr: Answer appears to be yes, since we were always overriding its context anyway.

zach-klippenstein: `GlobalScope` is just a scope with an empty context.

PR #86, opened 2018-12-12
Add Reactor test helper functions #85

### LINKED ISSUE #85: Add Reactor test helper functions
Add `assertTransition()` and `assertFinish()` helper functions.

qrazhan:
Closes #85

### DIFF
--- a/sample-game-common/build.gradle
+++ b/sample-game-common/build.gradle
@@ -38,4 +38,5 @@ dependencies {
   testImplementation deps.test.hamcrestCore
   testImplementation deps.test.junit
   testImplementation deps.rxjava2.extensions
+  testImplementation project(':workflow-test')
 }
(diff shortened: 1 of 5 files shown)


workflow-rx2/src/main/java/com/squareup/workflow/rx2/TestUtils.kt
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright 2017 Square Inc.
```
zach-klippenstein: Should be 2018.

workflow-rx2/src/main/java/com/squareup/workflow/rx2/TestUtils.kt
```diff
@@ -0,0 +1,50 @@
+/*
```
zach-klippenstein: Let's make a new module called `workflow-test` and put this there, so it doesn't pollute non-test code. I think we will probably end up putting more stuff in there at some point.

@rjrjr WDYT?

rjrjr:
Definitely

sample-game-common/build.gradle
```diff
@@ -38,4 +38,5 @@ dependencies {
   testImplementation deps.test.hamcrestCore
   testImplementation deps.test.junit
   testImplementation deps.rxjava2.extensions
+  testImplementation project(':workflow-test')
```
zach-klippenstein: Project dependencies should go above any third-party ones in each group.

qrazhan:
Gunna just merge as-is, since I don't have a ton of experience with co-routines. It'd be great to pair on the conversion so I can learn more! : )

workflow-rx2/src/main/java/com/squareup/workflow/rx2/TestUtils.kt
rjrjr: Yes yes yes.

PR #89, opened 2018-12-12
Make nextDelegateReaction and workerResult first-class EventChannel citizens.

### LINKED ISSUE #80: Deprecate EventSelectBuilder.onSuccess?
`EventSelectBuilder.onSuccess(Single)` is either a wart that people resort to when they wish their Rx code could do things that come for free from a `kotlinx.coroutines.experimental.selects.select` statement, or else a necessary escape hatch because we can't anticipate everything.

If we're sure it's a wart, we should instead make `EventSelectBuilder` those use cases more directly.

The specific problem case that prevents us from deprecating `onSuccess` today is this one from the `rx2.Reactor` kdoc. It could be handled by making `EventSelectBuilder` aware of `WorkflowPool`.

> If you need to handle other events while the workflow is running, use
> the [EventChannel] and a combination of `onSuccess()` and `onEvent()`.
> Remember to call [WorkflowPool.abandonDelegate] if you leave while
> the nested workflow is still running!

```kotlin
   is Delegating -> events.select {
     onSuccess(workflows.nextDelegateReaction(state)) {
       when (it) {
         is EnterState -> EnterState(state.copy(delegateState = it.state))
         if FinishWith -> when (it.result) {
           is DoSomething -> EnterState(DoingSomething)
           is DoSomethingElse -> EnterState(DoingSomethingElse)
         }
       }
     }
     onEvent<Cancel> {
       workflows.abandonDelegate(state.id)
       EnterState(NeverMind)
     }
   }
```

zach-klippenstein:
Now instead of saying
```kotlin
onSuccess(workflowPool.nextDelegateReaction(delegating)) { reaction ->
  …
}
```

you just do
```kotlin
workflowPool.onNextDelegateReaction(delegating) { reaction ->
  …
}
```

This means we can finally deprecate `onSuccess` (which this change also does).

Closes #80.


zach-klippenstein:
Forgot to update the kdoc…

workflow-rx2/src/main/java/com/squareup/workflow/rx2/EventSelectBuilder.kt
```diff
@@ -74,10 +80,55 @@ class EventSelectBuilder<E : Any, R : Any> internal constructor(
...
+    handler: (Reaction<S, O>) -> R
+  ) {
+    with(builder) {
+      nextDelegateReaction(delegating)
+          .onAwait { just(handler(it)) }
+    }
+  }
+
+  /**
+   * Selects on [worker] being started, if necessary, and then finishing.
```
rjrjr:
```
Starts the indicated [worker] if it wasn't already running, and selects on the result from the worker finishing.
```

zach-klippenstein:
`Reactor` docs are updated.

workflow-rx2/src/main/java/com/squareup/workflow/rx2/EventSelectBuilder.kt
zach-klippenstein: Done.

Issue #90, opened 2018-12-13
WorkflowPools should be scoped to workflows.


zach-klippenstein:
AC:
* If my child starts a `Worker` with the same `Type` as a `Worker` that I've started, it should get its own invocation.
* When a workflow finishes, it's `WorkflowPool` should be `abandonAll`ed.

Note: May play into structured concurrency support (#68), but is a separate issue.

PR #91, opened 2018-12-13
Restores the `abandonWorker()` call in `RunGameReactor`.

rjrjr:
Otherwise it just hangs out in the global space and
the result is picked up the next time we enter this
state. Confusing, bad example.

(That'll be less of a landmine when #90 is fixed.)


rjrjr:
Update fixes trailing blanks spotted by ktlint.

sample-game-common/src/main/java/com/squareup/sample/tictactoe/RunGameReactor.kt
```diff
@@ -126,6 +123,8 @@ class RunGameReactor(
         }
 
         onEvent<PlayAgain> {
+          workflows.abandonWorker(logGameWorker)
```
zach-klippenstein: We should have a comment here saying why this is necessary, and what will happen if we don't.

Issue #97, opened 2018-12-13
WorkflowPool.workflowUpdates(...) : Observable<WorkflowHandle<...>>


rjrjr:
Presuming the `WorkflowHandle` PR lands (#87 ), we should add something like:

```fun WorkflowPool.workflowUpdates(handle: RunningWorkflow<...>) :
    Observable<WorkflowHandle<...>>```

Will be useful at the top level. E.g., simplify `ShellActivity` even further, and move even further to `WorkflowPool` being The Way to run things.

rjrjr:
Oops, dup  of #101 , and that one is a bit better described.

PR #98, opened 2018-12-13
Goodbye Delegating, hello WorkflowHandle

### LINKED ISSUE #87: Delegating interface is too confusing
Perhaps it should be called WorkflowTicket, and perhaps it can be a struct rather than an interface. So, instead of

```kotlin
  internal data class Playing(override val delegateState: Turn) : RunGameState(),
      Delegating<Turn, TakeTurnsEvent, CompletedGame> {
    override val id = TakeTurnsReactor::class.makeWorkflowId()
  }
```

maybe something like

```kotlin
  internal data class Playing(
    val takingTurns: Ticket<Turn, TakeTurnsEvent, CompletedGame> ) : RunGameState()
```

And perhaps `awaitNextDelegateReaction` is changed to include the updated `Ticket` every time.

Also need doc, examples of parallel child workflows (master / detail, picker grid, status bar...). Two point of confusion:

 * How do you hang multiple `Delegating` / `Ticket` off of single parent
 * If workflows only run once and emit one result, how do you keep, say, the picker grid running a long time.

These are known and simple, but not easy to discover.

rjrjr:
Closes #87

`Delegating` was a confusing interface, an artifact of how I backed into
stitching together composite workflow states.

It's replaced by `WorkflowHandle`, a sealed class with two concrete types:

 - `RunningWorkflow`, which includes the workflows current (or bootstrap) state
 - `FinishedWorkflow`, which includes its result

Compositing now looks a lot like it did when we were still hanging live
workflows off of state objects, but everything remains a value type.

We also introduce a convention of reactor companion objects providing
a `run() : RunningWorkflow<...>` methods, for easy bootstrapping.

E.g., this is the state of the `RunGameReactor` while it's
delegating to the `TakeTurnsReactor`:

```
  internal data class Playing(
    val takingTurns: RunningWorkflow<Turn, TakeTurnsEvent, CompletedGame>
  ) : RunGameState() {
    constructor(turn: Turn) : this(TakeTurnsReactor.run(turn))
  }
```

This is how that reactor behaves in that state:

```
    is Playing -> events.select {
      workflows.onWorkflowUpdate(state.takingTurns) {
        when (it) {
          is RunningWorkflow -> EnterState(Playing(it))
          is FinishedWorkflow -> when (it.result.ending) {
            Quitted -> EnterState(MaybeQuitting(it.result))
            else -> EnterState(GameOver(it.result))
          }
        }
      }
    }

```

and here is how it's rendered:

```
      is Playing -> {
        return TakeTurnsRenderer
            .render(state.takingTurns.state, workflows.input(state.takingTurns), workflows)
            .let { MainAndModalScreen(it) }
      }
```

A composite workflow running several others in parallel might look like this:

```
  data class LeftAndRightState(
    val leftWorkflow: RunningWorkflow<LeftState, LeftEvent, LeftResult>,
    val rightWorkflow: RunningWorkflow<RigthState, RightEvent, RightResult>,
  ) : LeftAndRightState()
```

```
    is LeftAndRightState -> events.select {
      workflows.onWorkflowUpdate(state.leftWorfklow) {
        when (it) {
          is RunningWorkflow -> EnterState(state.copy(leftWorkflow = it))
          is FinishedWorkflow -> when (it.result.ending) { ... }
        }
      }

      workflows.onWorkflowUpdate(state.rightWorfklow) {
        when (it) {
          is RunningWorkflow -> EnterState(state.copy(rightWorkflow = it))
          is FinishedWorkflow -> when (it.result.ending) { ... }
        }
      }
    }

```

Also tidies up `ShellReactor` and `ShellState` a bit, to make them more
consistent with the others. This is done with an eye toward #97.


rjrjr:
To do: 

- [ ] Rebase
- [ ] kdoc
- [ ] can `Id` be hidden more? E.g. do we need the methods that take `id` arguments?

rjrjr:
Offline review from @loganj:

@ray bikeshed thoughts/sketch:

```kotlin
lateinit var mainWorkflow: WorkflowHandle<Blah,Blah,Blah>
​
val authWorkflow: WorkflowHandle.NotStarted<Blah, Blah, Blah> = AuthReactor::class.createWorkflow(initialState)
​
mainWorkflow = authWorkflow
​
​
workflows.run(mainWorkflow) {
 when(it) {
  is NotStarted -> workflows.start(it)
  is Running -> EnterState(state.copy(mainWorkflow = it))
  is Finished -> when (it.result) { ... }
 }
}
```
 
points of interest:
 - creating the workflow is just named create, since nothing’s getting started/run
 - not-started and actually-running are distinct. this isn’t _necessary_, but it saves you from trying to figure out that it isn’t necessary
also tempted to remove `createWorkflow` from `AuthReactor::class` and put it on `WorkflowPool`, just because that gives you a clear starting point!

```kotlin
val authWorkflow: WorkflowHandle.NotStarted<Blah, Blah, Blah> = workflows.create(AuthReactor::class, initialState)
```

having handles is pretty awesome though, definitely will kill boilerplate and is a lot easier to understand

rjrjr:
Only thing I strongly disagree with is adding `NotStarted` / `start()`. They literally don’t do anything except offer extra ways to make mistakes and for us to introduce bugs. I’d rather work on the names and docs and teaching.

Thanks for the quick feedback!

rjrjr:
> More from @loganj:

What bug opportunities do they create? Without `start` it's unclear when starting happens. There's literally no obvious entry point. You come to this thing with a mental model and end up working really hard to figure out how the library maps to it.

maybe it’s enough to rename `Running` to `Live` or something. all I know is Shalisa and I spent most of our time with that diff trying to figure out where the workflow got started. We figured out pretty quickly, after looking at a `run()` implementation and seeing that wasn’t it, that it _must_ be through `onWorkflowUpdates`… but then we had to trace through and prove it to ourselves.

rjrjr:
More moving parts is more moving parts. Let’s work on the namings and the docs

rjrjr:
Besides the confusion that Logan points out, my main concern with this so far is whether we'll be able to test composite workflows in isolation from their subs. Before merging I'm going to add some unit tests to the samples see how tricky that is.

zach-klippenstein:
```kotlin
      is Playing -> {
        return TakeTurnsRenderer
            .render(state.takingTurns.state, workflows.input(state.takingTurns), workflows)
            .let { MainAndModalScreen(it) }
      }
```

We could make an extension for that 3rd line to make it simpler to use:
```kotlin
fun Renderer<S, E, R>.render(handle: RunningWorkflow<S, E, R>, workflows: WorkflowPool): … =
    render(handle.state, workflows.input(handle), workflows)
```
which would let us do
```kotlin
      is Playing -> {
        return TakeTurnsRenderer
            .render(state.takingTurns, workflows)
            .let { MainAndModalScreen(it) }
      }
```

loganj:
We might disagree on this, but I feel pretty strongly that our confusion came at least as much from the interaction pattern as the naming..

Commonly if a library provides a handle to a resource, that resource exists, or at the very least there's a very clear way for me to 1) create it, and 2) if it's a process, start it.

When I walk up to the workflow library API, I'm looking for an obvious entry point to create and start a workflow, and there just isn't one. Probably I find and start at the WorkflowPool class (I guess?), but it doesn't have anything called `createWorkflow` or `startWorkflow`, just a way to subscribe to workflow updates which I'm not ready for yet.

If I'm looking at someone else's code and hoping to cargo-cult it, I have a similar problem:

Where does the workflow get created and started? 

I can see where they ask the pool to give them the next frame for the workflow associated with a handle. Okay, where'd the handle come from? Ah, they just instantiate it. It doesn't seem like it's a handle to anything real, but I don't know that for sure --  maybe the workflow got created as a side effect? There's a pool, but they haven't interacted with it in any way at this point. Maybe there's a static thing? Back to that pool callsite... I guess asking for the next frame might start the workflow, but that's a surprising side effect, so maybe it _is_ the handle creation that did it? I'm still looking for some kind of create or start, but not finding it, and it could just as easily be in either place I did find, but it's almost equally weird in either place.  Maybe there's a third thing I haven't found yet...

And so I spend a bunch of time trying to convince myself that I understand the lifecycle -- when is there actually a workflow, when does it start, what's attached to that handle and where? It's hard though because there are a few layers of helpers in the pool class, and I'm really just looking for something called "create" and something called "start".

So these are the two questions my sketch attempts to answer, and that I don't think the API can answer by just renaming things: 1) how do you create a workflow, and 2) how do you start a workflow? 

My sketch answers the first question by making  `WorkflowPool` an obvious factory -- references come from _from_ it, and can be brought _back_ to it, but you don't make them yourself and bring them to it in the first place. Where do you create a workflow? At WorkflowPool. You'll get a handle.

The second question is answered, again, by making WorkflowPool capable of "starting" a workflow explicitly. Where do you start a workflow? At WorkflowPool. You need to give it a handle that you got from WorkflowPool.

It's definitely not perfect --  the observation interaction is still not the easiest thing to wrap your head around -- but at least it answers those two questions. There may be better ways to answer them as well, but I don't think you can do it without changing the interaction pattern.

rjrjr:
The problem with Logan's sketch is it, like, doesn't work. The handle is no longer part of the parent state, which is kind of the whole point of this scheme. And this bit:

```
workflows.run(mainWorkflow) {
 when(it) {
  is NotStarted -> workflows.start(it)
  is Running -> EnterState(state.copy(mainWorkflow = it))
  is Finished -> when (it.result) { ... }
 }
}
```

That `NotStarted` line _must_ end with either `EnterState` or `FinishWith`. So the closest I see us getting to this sketch is more like:

```
workflows.run(mainWorkflow) {
 when(it) {
  is NotStarted -> EnterState(state.copy(mainWorkflow = workflows.start(it)))
  is Running -> EnterState(state.copy(mainWorkflow = it))
  is Finished -> when (it.result) { ... }
 }
}
```

Which doesn't strike me as an improvement.

(Mostly adding this for the record, I know Mr. Johnson has been working on better / working sketches in the meantime.)

rjrjr:
I was able to eliminate direct uses of `id`, so that's something. We're stuck with the explicit `E` types, I think. 

I played with moving the `run()` methods to `WorkflowPool` instead of hanging them off the `Reactor` classes, but it doesn't work so well -- those methods all shield callers from needing knowledge of the reactor's default state, a service we really don't want to lose.

I've renamed the `RunningWorkflow` handle type to `RunWorkflow`, hoping that the more imperative name gets across that it starts out as a request. Also punched up the kdoc on that class to make the point explicitly:

```
/**
 * Handle that instructs [WorkflowPool.workflowUpdate] to start running the described
 * workflow from the given [state], if it isn't running already, and return
 * a new [WorkflowHandle] when the workflow's state changes from the given,
 * or it produces a result.
 *
 * If you're using an Rx2 `EventChannel`, you'll pass the handle to `onWorkflowUpdate`
 * inside an `events.select {}` block.
 */
data class RunWorkflow<S : Any, E : Any, O : Any>(
```

I haven't written the promised sample unit test yet, but I'd like to get this merged and follow up with that. I'll track the requirement as a blocker for the next milestone so that we don't forget about it and paint ourselves into corners such a test would reveal.

rjrjr:
Let's see if we can find a better name than `run()` for those methods…

rjrjr:
On each reactor:

```
  companion object {
    /**
     * Returns a [RunWorkflow] handle that will instruct a [WorkflowPool] to call
     * [launch] and start a workflow.
     */
    fun getStarter(
```

workflow-core/src/main/java/com/squareup/workflow/WorkflowHandle.kt
```diff
@@ -0,0 +1,72 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.workflow
+
+import com.squareup.workflow.WorkflowPool.Id
+import com.squareup.workflow.WorkflowPool.Type
+
+/**
+ * Describes a workflow to be run by a [WorkflowPool]. Includes the the
```
steveinflow: the the

workflow-core/src/main/java/com/squareup/workflow/WorkflowPool.kt
```diff
@@ -89,9 +89,8 @@ class WorkflowPool {
    * Unique identifier for a particular [Workflow] to be run by a [WorkflowPool].
    * See [Type.makeWorkflowId] for details.
    *
-   * Convenience extension functions exist on `KClass<Launcher>` and [Delegating] to create IDs:
+   * A convenience extension functions exists on `KClass<Launcher>` to create IDs:
```
steveinflow: `function`

sample-auth-common/src/main/java/com/squareup/sample/authworkflow/AuthReactor.kt
```diff
@@ -109,4 +111,15 @@ class AuthReactor(
...
+
+  companion object {
+    /**
+     * Returns a [RunWorkflow] handle that will instruct a [WorkflowPool] to call
+     * [launch] and start a workflow.
+     */
+    fun getStarter(
+      state: AuthState = AuthState.startingState()
+    ): RunWorkflow<AuthState, AuthEvent, String> =
+      RunWorkflow(AuthReactor::class.makeWorkflowId(), state)
```
steveinflow: It's generally so nice to not have to know about `ID`s anymore in client code. Could we make a utility method on `RunWorkflow` or `WorkflowHandle` to just accept a Launcher and generate the id internally?

rjrjr: Will give that a shot.

workflow-core/src/main/java/com/squareup/workflow/WorkflowPool.kt
```diff
@@ -302,15 +305,15 @@ inline fun <reified S : Any, reified E : Any, reified O : Any> WorkflowPool.regi
...
- * @see WorkflowPool.awaitNextDelegateReaction
+ * @see WorkflowPool.awaitWorkflowUpdate
  */
-fun <S : Any, O : Any> WorkflowPool.nextDelegateReaction(
-  delegating: Delegating<S, *, O>
-): Deferred<Reaction<S, O>> = GlobalScope.async(Unconfined) {
-  awaitNextDelegateReaction(delegating)
+fun <S : Any, E : Any, O : Any> WorkflowPool.workflowUpdate(
+  handle: RunWorkflow<S, E, O>
+): Deferred<WorkflowHandle<S, E, O>> = GlobalScope.async(Unconfined) {
```
steveinflow: Nice side-effect of returning a `Handle` instead of the `Reaction` here is the parent doesn't have the option of peeking into the child's state anymore; it's a clarifying step.

rjrjr:
Just snuck in the promised sample unit test. @steveinflow, check it out.

sample-android-app/src/test/java/com/squareup/sample/authgameapp/TestLauncher.kt
```diff
@@ -0,0 +1,40 @@
...
+import com.squareup.workflow.rx2.Reactor
+import com.squareup.workflow.rx2.doLaunch
+import io.reactivex.Single
+import io.reactivex.subjects.PublishSubject
+
+/**
+ * Launches pretend [Workflow] instances for tests. Move them through
+ * states by pushing [Reaction]s into [reactions].
+ */
+open class TestLauncher<S : Any, E : Any, O : Any> : Launcher<S, E, O> {
```
steveinflow: This seems generally useful, should it be in the library somewhere?

zach-klippenstein: Definitely. Let's hide the subject behind a `sendState` method, and then you can drop the rx2 dependencies and put this in the test module.

sample-game-common/src/main/java/com/squareup/sample/tictactoe/RunGameReactor.kt
```diff
@@ -86,10 +107,10 @@ class RunGameReactor(
     }
 
     is Playing -> events.select {
-      workflows.onNextDelegateReaction(state) {
+      workflows.onWorkflowUpdate(state.takingTurns) {
         when (it) {
-          is EnterState -> EnterState(state.copy(delegateState = it.state))
-          is FinishWith -> when (it.result.ending) {
+          is RunWorkflow -> EnterState(Playing(it))
```
zach-klippenstein: This reads so nicely.

sample-android-app/src/test/java/com/squareup/sample/authgameapp/TestLauncher.kt
rjrjr: #104

workflow-core/src/main/java/com/squareup/workflow/WorkflowHandle.kt
```diff
@@ -0,0 +1,88 @@
...
+  abstract val id: Id<S, E, O>
+
+  companion object {
+
+    /**
+     * Returns a [RunWorkflow] handle that will instruct a [WorkflowPool] to call
+     * launch a workflow of the given type in the given state.
+     */
+    inline fun <reified S : Any, reified E : Any, reified O : Any> getStarter(
+      launcherType: KClass<out Launcher<S, E, O>>,
```
zach-klippenstein: Sad Kotlin doesn't have the ability to infer a subset of the type params 😞

workflow-core/src/main/java/com/squareup/workflow/WorkflowHandle.kt
```diff
@@ -0,0 +1,88 @@
...
+  }
+}
+
+/**
+ * Handle that instructs [WorkflowPool.workflowUpdate] to start running the described
+ * workflow from the given [state], if it isn't running already, and return
+ * a new [WorkflowHandle] when the workflow's state changes from the given,
+ * or it produces a result.
+ *
+ * If you're using an Rx2 `EventChannel`, you'll pass the handle to `onWorkflowUpdate`
```
zach-klippenstein: Should have a sentence in here for if you're not using Rx2 as well.

workflow-core/src/main/java/com/squareup/workflow/WorkflowPool.kt
```diff
@@ -89,9 +89,8 @@ class WorkflowPool {
    * Unique identifier for a particular [Workflow] to be run by a [WorkflowPool].
    * See [Type.makeWorkflowId] for details.
    *
-   * Convenience extension functions exist on `KClass<Launcher>` and [Delegating] to create IDs:
+   * A convenience extension function exists on `KClass<Launcher>` to create IDs:
    *  - `KClass<Launcher>.makeWorkflowId()`
```
zach-klippenstein: Nit: if this is just the only item in the bullet list now, doesn't need to be a list.

workflow-core/src/main/java/com/squareup/workflow/WorkflowPool.kt
```diff
@@ -176,13 +177,10 @@ class WorkflowPool {
     type: Type<I, Nothing, O>
   ): O {
     register(worker.asLauncher(), type)
-    val delegating = object : Delegating<I, Nothing, O> {
-      override val id: Id<I, Nothing, O> = type.makeWorkflowId(name)
-      override val delegateState: I get() = input
-    }
-    val workflow = requireWorkflow(delegating)
+    val handle = RunWorkflow(type.makeWorkflowId(name), input)
```
zach-klippenstein: So much nicer.

workflow-core/src/test/java/com/squareup/workflow/ReactorIntegrationTest.kt
```diff
@@ -322,7 +322,9 @@ class ComposedReactorIntegrationTest {
     echoJobId: String,
     event: String
   ) {
-    pool.input(StringEchoer::class.makeWorkflowId(echoJobId))
+    val sketchyHandle =
```
zach-klippenstein: Nit: Could use a comment on here explaining why it's sketchy.

workflow-core/src/main/java/com/squareup/workflow/WorkflowHandle.kt
rjrjr: How's this?

```
/**
 * Handle that instructs [WorkflowPool.workflowUpdate] to start running the described
 * workflow from the given [state], if it isn't running already, and return
 * a new [WorkflowHandle] when the workflow's state changes from the given,
 * or it produces a result.
 *
 * In a [Reactor.onReact] implementation, pass the handle to
 * `WorkflowPool.`[workflowUpdate()] inside a `select` block
 * to start the workflow, handle its state updates, and handle the result
 * when it finishes.
 *
 * If you're using an Rx2 `EventChannel`, you'll pass the handle to `onWorkflowUpdate`
 * inside an `events.select {}` block.
 */
```

PR #111, opened 2019-01-04
Deprecate Reactor + doLaunch and replace with a function that takes a block.

### LINKED ISSUE #63: Reactor should go away, just become a single function.
AC: 
 * Replace with CoroutineScope.statefulWorkflow() extension
 * Takes onReact() and doBeforeLaunch() lambdas

### LINKED ISSUE #88: Separate Reactor and Launcher concerns
It's still too hard to work with hot observables / legos. In particular, no good way for launch to instantiate one that lives only as long as a particular workflow instance does. Generated workflow can receive events from the lego, and the lego can be unsubscribed and gc'd when the workflow ends, but the workflow can't push commands into such an on-demand lego. 

Seems easy to solve if reactor interface dies and is replaced with a builder. See pic. 

img_20181212_123213

zach-klippenstein:
* Instead of implementing `Reactor` (which extends `Launcher`), you now
  implement `Launcher` directly.
* In your `launch` implementation, instead of calling `doLaunch`, you now
  call `workflows.discreetStateWorkflow` and pass it either a lambda or a
  method reference to what used to be your `onReact` method (for coroutines,
  suspend references aren't allowed until 1.3 so you must currently wrap
  with another lambda).

This:
```kotlin
class MyReactor : Reactor<Foo, Bar, Baz> {
  override fun launch(…) = doLaunch(…)

  override fun onReact(…): Single<…> = …
}
```
becomes this:
```kotlin
class MyLauncher : Launcher<Foo, Bar, Baz> {
  override fun launch(…) = workflows.discreetStateWorkflow(…, ::onReact)

  private fun onReact(…): Single<…> = …
}
```
or even:
```kotlin
class MyLauncher : Launcher<Foo, Bar, Baz> {
  override fun launch(…) = workflows.discreetStateWorkflow(…) { state, events, workflows ->
    …
  }
}
```

Closes #63, closes #88.


zach-klippenstein:
@rjrjr I'll hold off rebasing and merging this until #114 lands so we can release 0.7.0.

zach-klippenstein:
Hey yall (@steveinflow @loganj @afollestad @mbrubin56), let's discuss naming. `Reactor` has always been infeasible, knew that going in. Since this change is getting rid of that interface, it seems an appropriate time to fix. We need to figure out what to call:
* `Reaction`
* The workflow-building function that takes a state transition function and returns a `Workflow` (currently called `discreteStateWorkflow` as of this comment).
* The typealias for state transition functions (currently called `StateTransitionFunction`).

steveinflow:
Recommend maximizing simplicity:
`Reactor` -> `TransitionFunction`
`Reaction` -> `Transition`
`doLaunch` -> launch (or launchWorkflow since you need to differentiate from Workers?)

The library calls out its state machine roots front and center, so `State` doesn't need to be specified in every function name.

workflow-rx2/src/test/java/com/squareup/workflow/rx2/Rx2WorkflowPoolIntegrationTest.kt
```diff
@@ -45,34 +46,34 @@ class Rx2WorkflowPoolIntegrationTest {
...
+    ): Single<out Reaction<String, String>> = events.select {
+      onEvent<String> {
+        eventsSent += it
+        if (it == STOP) FinishWith(state) else EnterState(it)
+      }
+    }
   }
 
-  val myReactor = MyReactor()
+  val myReactor = MyLauncher()
```
steveinflow: nit: `myLauncher`

workflow-rx2/src/main/java/com/squareup/workflow/rx2/DiscreteStateWorkflow.kt
```diff
@@ -0,0 +1,249 @@
...
+ * ## Naming
+ *
+ * To help with debugging, this function accepts a [name]. This name will
+ * be used in exceptions thrown from the workflow. Additionally, if coroutine debugging is enabled,
+ * the name will show up in stack traces and thread names.
+ */
+fun <S : Any, E : Any, O : Any> WorkflowPool.discreteStateWorkflow(
+  initialState: S,
+  name: String? = null,
+  nextState: StateTransitionFunction<S, E, O>
```
steveinflow: nit: this parameter is not a state, so `nextState` is a misleading name.
Would also reorder these so the optional param is last.

zach-klippenstein: The function parameter must come last so it can take a block, that's a kotlin syntax limitation.

zach-klippenstein:
Chatted offline with the people working on the iOS side of this, and realized it's probably generally time to rethink what we're calling `Workflow` and naming in general. Will probably hold off on trying to land this until we've figured out the broader naming strategy.

rjrjr:
Naming ideas:

 - `WorkflowPool` > `WorkPool`
 - `Workflow` > `InteractiveWorker`
    - `discreteStateWorkflow` > `interactiveWorker`
    - If we do need other strategies, they could just be overloads of that method.
 - I know this will get shot down, but I still wonder if we should rename `state` to `prompt`. It makes more sense given then `InteractiveWorker` name.

That frees up `Workflow` to move over to the `viewbuilder` side of the pond, as discussed offline.

zach-klippenstein:
I like `InteractiveWorker`. The prompt thing would also be weird for any non-UI uses, not just liminal states.

Here's some more ideas:
`Workflow` → `ActorWorker`
`Worker` → `FutureWorker`
`WorkflowPool` → `WorkerPool`
`Workflow.state` → `Workflow.output`
`workflow()` → `actorWorker()`
`StateTransitionFunction` → `StatefulActor`
`discreteStateWorkflow()` → `statefulActor()`
`Reaction(EnterState|FinishWith)` → `Action(EnterState|FinishWith)`

zach-klippenstein:
This PR is obsolete given the API changes for convergence with the Swift API.

PR #114, opened 2019-01-04
Separates WorkflowResult and WorkflowPool.Handle.

### LINKED ISSUE #108: WorkflowHandle.getStarter() is confusing
AC: `WorkflowHandle.getStarter()` replaced by an overload of `RunWorkflow()`

We mostly refer directly to `RunWorkflow`, but its main factory method is `WorkflowHandle.getStarter()`  -- it's hard to find. Let's just put a convenience
constructor on `RunWorkflow`.

### LINKED ISSUE #110: WorkflowResult / WorkflowHandle
The boundaries around the `WorkflowHandle` classes aren't quite right.  `RunWorkflow` is the only real handle, you never pass a `FinishedWorkflow` anywhere. So, perhaps `RunWorkflow` is what should be called `WorkflowHandle`, carried around like this:

```kotlin
sealed class WorkflowUpdate<S, E, O> {
  data class Running<S, E, O>(
    val handle: WorkflowHandle<S, E, O>
  ) : WorkflowUpdate<S, E, O>

  data class Finished<O>(
    val result: O
  ) : WorkflowUpdate<Nothing, Nothing, O>
}
```

```kotlin
    is Playing -> events.select {
      workflows.onWorkflowUpdate(state.takingTurnsWorkflow) {
        when (it) {
          is Running -> EnterState(Playing(it.handle))
          is Finished -> when (it.result.ending) {
            Quitted -> EnterState(MaybeQuitting(it.result))
            else -> EnterState(GameOver(it.result))
          }
        }
      }
    }
```

rjrjr:
Closes #108, closes #110.

WorkflowHandle was still confusing as a sealed class. Really, the only handle
was its `RunWorkflow` sub-type. Everything else was about getting updates from
`WorkflowPool.awaitWorkflowUpdate`.

With this change,

- `RunWorkflow` becomes `WorkflowPool.Handle`, whose sole job is to define a workflow to be run.
- `WorkflowHandle.getStarter()` is replaced by `WorkflowPool.handle()`
- `WorkflowPool.awaitWorkflowUpdate` now returns the `WorkflowUpdate` sealed class, which
  is either `Running(val handle: WorkflowHandle)` or `Finished(val result: R)`

We also get rid of most variants of `makeWorkflowId()`, and make the sole
remaining one internal, and `WorkflowPool.input(Id)` is gone. Handle` becomes
the one and only public way to refer to a [Workflow] instance.


workflow-core/src/main/java/com/squareup/workflow/WorkflowUpdate.kt
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright 2017 Square Inc.
```
zach-klippenstein: Nit: copyright year for new file should be this year.

rjrjr: Fixed

workflow-core/src/main/java/com/squareup/workflow/WorkflowPool.kt
```diff
@@ -299,6 +317,25 @@ class WorkflowPool {
...
+    inline fun <reified S : Any, reified E : Any, reified O : Any> handle(
+      launcherType: KClass<out Launcher<S, E, O>>,
+      state: S,
+      name: String = ""
+    ): Handle<S, E, O> = Handle(launcherType.workflowType.makeWorkflowId(name), state)
+
+    /**
+     * Creates a [Handle] describing a [Workflow] to be managed via [awaitWorkflowUpdate].
+     */
+    inline fun <reified E : Any, reified O : Any> handle(
```
rjrjr: This is for @steveinflow .

PR #118, opened 2019-01-14
Update the AGP to 3.3.0 and Kotlin to 1.3.11

SalvatoreT:
Update AGP, Kotlin, and remove the "experimental" modifier for coroutines.


zach-klippenstein:
Discussed offline: this is not possible to land until we have internally unblocked ourselves from upgrading to kotlin 1.3 as well.

There are also many more changes that will have to be made to this library as part of the upgrade because the coroutines API and behaviors changed significantly between versions 0.26.1 and 1.0. Notably, cancellation API changed, `Unconfined` dispatcher became trampolining, etc. Most of these changes should be pretty mechanical, but we'll need to be careful not to break any of our own contracts.

SalvatoreT:
Discussed offline: Zach is going to take care of this.

It's time to thank this branch and let it go.

PR #121, opened 2019-01-16
Removed "Basically a monad"

pyricau:
Several internal slack messages showed developers are getting scared and confused by "basically a monad". The sentence that follows is super clear, so seems worth sticking to that?


zach-klippenstein:
Yea, that wording was a bad idea. Thanks for fixing!

PR #124, opened 2019-01-17
ViewBuilder is ViewBinder. Makes Scenes as well as Views.

### LINKED ISSUE #57: Demonstrate use  of TransitionManager with ViewBuilder
AC

 * Sample use of Scene() and TransitionManager() 

ViewBuilder name may need to change

rjrjr:
This still doesn't address any kind of declarative / pluggable scheme for
selecting animators or transitions, but it does fix two gaps:

 * Containers can use TransitionManager
 * Views can be created from code as well as layout files

I'm going to look into transition effects configuration separately, but may
well decide that requiring containers to just write code for what they want is
good enough.

Closes #57.


viewbuilder-android/src/main/java/com/squareup/viewbuilder/ViewStack.kt
```diff
@@ -18,69 +18,107 @@ package com.squareup.viewbuilder
...
+   * @param setUpNewView function to run against the incoming view to restore
+   * its previously saved [view hierarchy state][View.saveHierarchyState], if there
+   * is any. This should be called after the view is inflated / instantiated, and before
+   * it is attached to a window.
    */
-  fun replace(
-    currentView: View,
-    newView: View
-  ): ReplaceResult {
+  class UpdateTools(
```
rjrjr: Should this be an interface instead?

rjrjr:
Update fixes some lint checks, polishes up docs and a number of names.

rjrjr:
Update changes `UpdateTools` to an interface.

PR #132, opened 2019-01-22
Introduce WorkflowPoolMonitor and an implementation that produces trace files.

### LINKED ISSUE #129: Debugging feature to write chrome-compatible trace files of workflow activity
AC: Some output of a json file that can be loaded into `chrome://tracing` to inspect workflow timelines.

See [this gdoc](https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/edit?usp=drivesdk) for the trace file format spec.

zach-klippenstein:
Introduces the concept of `WorkflowPoolMonitor`, an interface that can monitor `WorkflowPool` events such as state changes, events, etc.

Also introduces an implementation of that interface, `TracingWorkflowPoolMonitor`, that generates trace files compatible with Chrome's `chrome://tracing` viewer.

Closes #129.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -40,6 +40,7 @@ buildscript {
       'mavenPublishPlugin': '0.6.0',
       'mockito': '2.7.5',
       'mockitoKotlin': '1.5.0',
+      'moshiVersion': '1.8.0',
       'rxjava2Extensions': '0.20.4',
   ]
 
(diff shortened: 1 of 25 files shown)


zach-klippenstein:
This PR is obsolete given the API changes for convergence with the Swift API. I'll do something similar later.

PR #139, opened 2019-02-12
Move Kotlin files into their own directory to prep for Swift import.

### LINKED ISSUE #136: Move Kotlin files into their own folder to prep for Swift
AC: All Kotlin-specific files live in a top-level `kotlin` directory.

zach-klippenstein:
Updated link in root README, and added a root project name so it doesn't
fall back to being "kotlin". Changed the Travis config to use an explicit
matrix include for the android build, so we can also add a Swift build later.

Documentation for configuring Travis matrix builds is:
  - https://stackoverflow.com/questions/27644586/how-to-set-up-travis-ci-with-multiple-languages
  - https://docs.travis-ci.com/user/build-matrix/

Closes #136.


kotlin/settings.gradle
```diff
@@ -13,6 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+rootProject.name = "workflow"
```
rjrjr: What does this do?

zach-klippenstein: (At least) two things:
image
image

Without that name, the project shows up as being called "kotlin" everywhere.

Issue #37, opened 2018-12-03
AlertDialog integration or replacement


rjrjr:
AC

 * Workflow apps can drive AlertDialog or, more likely, something that looks like AlertDialog

PR #142, opened 2019-02-15
AlertContainerScreen, AlertContainer

rjrjr:
MainAndModalScreen was going to try to solve too many problems.

Update to #37.


rjrjr:
Declining this. It's more re-usable than I thought.

Issue #145, opened 2019-03-01
Add ViewPersistence to AbstractModalContainer


rjrjr:
AC:
 * Rotating a `PanelScreen` in Tic Tac Workflow sample will preserve field content.
 * Restoring such a panel in a new process will preserve field content.

PR #146, opened 2019-03-01
Introduces ModalContainer in ViewRegistry, and PanelContainer sample.

rjrjr:
Actually three SHAs, includes a couple of necessary preliminaries:

 * Consolidates the sample modules. There were too many of them. Now just `android` and `common`.

 * Eliminates useless D param from event screens.

 * Introduces ModalContainer in ViewRegistry, and PanelContainer sample.

Update to #37.

Extracts AbstractModalContainer from AlertContainer, and builds ModalContainer from it.
ModalContainer show arbitrary views in Dialogs, whose size is WRAP_CONTENT around
those views.

PanelContainer sample is added to Tic Tac Workflow app, showing a way to use
ModalContainer to build modal subflows similar to those in Square PoS.

This is a bit more complex than is ideal, I think mainly due to the
limitations of `ViewRegistry` and its awkward `String` keys. The need
to add `PanelScreen` as a hook to get things wrapped in `PanelContainer`
is particularly unfortunate. But it's something we can iterate on.

AbstractModalContainer doesn't handle view persistence yet, tracked in #145.


rjrjr:
Meh, too big. Will break this up into separate PRs.

PR #148, opened 2019-03-01
Eliminates EventHandlingScreen

rjrjr:
Started out to remove its `D` param, but just nuked the whole thing instead. It's not very useful, and will be even less so w/v2.


rjrjr:
@zach-klippenstein In prep for v2, would it be more useful to go the whole way and just inline this interface?

zach-klippenstein:
> @zach-klippenstein In prep for v2, would it be more useful to go the whole way and just inline this interface?

You mean `EventHandlingScreen`? Yes might as well.

rjrjr:
Updated to eliminate the `EventHandlingScreen` interface entirely. It's just not useful.

PR #152, opened 2019-03-11
ModalContainer play nice with Android view persistence.

### LINKED ISSUE #145: Add ViewPersistence to AbstractModalContainer

rjrjr:
ModalContainer now implements `onSaveInstanceState()` and `onRestoreInstanceState()`.
Holds onto the most recently restored state, and applies it to the first shown base
view and dialog stacks.

Note also that ModalContainer subclasses and BackStackContainer must have their
`View.id` set in order to participate in view persistence.

Gives the 2fa screen a back button so that we can exercise popping
BackStackContainer again.

Closes #145

### DIFF
--- a/kotlin/samples/tictactoe/android/src/main/res/values/ids.xml
+++ b/kotlin/samples/tictactoe/android/src/main/res/values/ids.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!--
-  ~ Copyright 2017 Square Inc.
+  ~ Copyright 2019 Square Inc.
   ~
   ~ Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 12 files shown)


kotlin/samples/tictactoe/android/src/main/res/values/ids.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+  ~ Copyright 2017 Square Inc.
```
zach-klippenstein: 2017?

rjrjr:
WRT testing, need to get an espresso test going I think.

PR #154, opened 2019-03-13
Add source for documentation site

timdonnelly:


docs/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "gatsby-starter-default",
+  "description": "Gatsby default starter",
+  "version": "1.0.0",
+  "author": "Kyle Mathews <mathews.kyle@gmail.com>",
```
zach-klippenstein: Is this all correct?

docs/gatsby-config.js
```diff
@@ -0,0 +1,44 @@
+module.exports = {
+  pathPrefix: '/pages/~TDONNELLY/workflowdocs/master/browse',
```
davidapgar: This path prefix should probably be updated

docs/src/markdown/test.md
```diff
@@ -0,0 +1,279 @@
+---
```
davidapgar: I think this was a test, should be removed.

docs/package.json
timdonnelly: It is now!

PR #163, opened 2019-03-14
Use verbose output for `pod lint` on Travis

bencochran:


bencochran:
Hmm, `--verbose` is … [quite verbose](https://travis-ci.com/square/workflow/jobs/185014654). Maybe we should pipe through `xcpretty` or something?

davidapgar:
> Hmm, `--verbose` is … [quite verbose](https://travis-ci.com/square/workflow/jobs/185014654). Maybe we should pipe through `xcpretty` or something?

I think I agree that this is *too* verbose, and we could turn this on/off as needed if we're seeing issues.

bencochran:
> I think I agree that this is _too_ verbose, and we could turn this on/off as needed if we're seeing issues.

Yeah. I think I agree. Let’s leave things as-is for now. My ideal would be to have Travis collapse the Xcode build output but show the test output. But I don’t think we have any easy way to accomplish that.

PR #169, opened 2019-03-15
Zachklipp/release 0.8.0

zach-klippenstein:


Workflow.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name         = 'Workflow'
-  s.version      = '0.8.0'
+  s.version      = '0.9.0'
```
zach-klippenstein: @timdonnelly Is this the only version number I need to change?

dnkoutso: yes for iOS and distribution for CocoaPods.

zach-klippenstein: Thanks!

PR #173, opened 2019-03-15
Switch async workflow node tests to using expectations instead of spinning the runloop

davidapgar:
Because the tests were based upon sleeping for the right amount
of time they ended up flaking. The expectations should shorten the time
that we wait, while also giving us longer if the CI system happens
to be a bit slower.


bencochran:
Much better, thank you!

PR #179, opened 2019-03-16
Add Package.swift

timdonnelly:


.travis.yml
```diff
@@ -27,10 +27,23 @@ matrix:
...
       osx_image: xcode10.1
       before_install:
-      - gem update --system
-      - gem install bundler
+        - gem update --system
+        - gem install bundler
       script:
-      - bundle exec pod lib lint Workflow.podspec --allow-warnings
+        - bundle exec pod lib lint Workflow.podspec --allow-warnings
+    - language: swift
```
bencochran: Could we name these jobs to make differentiating between the two Xcode builds easier?

Package.swift
```diff
@@ -0,0 +1,30 @@
...
+        .library(
+            name: "Workflow",
+            targets: ["Workflow"]),
+    ],
+    dependencies: [
+        .package(url: "https://github.com/ReactiveCocoa/ReactiveSwift.git", from: "4.0.0")
+    ],
+    targets: [
+        // Targets are the basic building blocks of a package. A target can define a module or a test suite.
+        // Targets can depend on other targets in this package, and on products in packages which this package depends on.
```
bencochran: nit: remove template comments

.travis.yml
timdonnelly: Done!

Package.swift
timdonnelly: Done!

PR #181, opened 2019-03-16
Update ReactiveSwift to 5.0.0

timdonnelly:
ReactiveSwift 5.0.0 adds Swift 5 support (meaning that we can adopt Xcode 10.2 without a bunch of warnings coming out of our dependencies)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,7 @@ matrix:
         - gem update --system
         - gem install bundler
       script:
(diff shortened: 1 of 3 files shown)


bencochran:
Oh, also has `materializeResults` which might make @davidapgar happy (turns `Signal<T,E>` into `Signal<Result<T,E>, NoError>`)

PR #182, opened 2019-03-17
Added Xcode 10.2 support, add 10.2 to Travis matrix

timdonnelly:


swift/Workflow/Tests/WorkflowNodeTests.swift
```diff
@@ -179,6 +179,8 @@ final class WorkflowNodeTests: XCTestCase {
             struct State {}
 
             typealias Output = Int
+            
+            typealias Rendering = Void
```
timdonnelly: It appears that the Swift 5 compiler (even when running in 4.2 mode) has different behavior when it comes to type inference.

bencochran:
Should we add Swift 5 support to the podspec and Package.swift?

timdonnelly:
> Should we add Swift 5 support to the podspec and Package.swift?

I would, but `.v5` is only available when using SPM 5+ – so you always have to build with the latest version of the tooling.

bencochran:
> I would, but `.v5` is only available when using SPM 5+

Cool cool. I tried to find the answer to that very question documented somewhere and had no luck (same with source code, it wasn’t immediately obvious what version shipped with what definitions).

PR #183, opened 2019-03-17
A nicer way to accept Android SDK license on CI

kirillzh:
https://docs.travis-ci.com/user/languages/android/#dealing-with-licenses


kirillzh:
Turns out that "If Travis CI could not detect Maven or Gradle files, Travis CI Android builder will try to use Ant to build your project. By default it will use `ant debug install test` to run your test suite"...

kirillzh:
No idea why Travis is trying to use SDK 26...

PR #187, opened 2019-03-18
Fix some words

timdonnelly:
Reworded to be more diplomatic

### DIFF
--- a/docs/src/markdown/documentation/faq.md
+++ b/docs/src/markdown/documentation/faq.md
@@ -20,10 +20,8 @@ Of course! Workflow was designed to make complex application architecture predic
 
 ### Why do we need another architecture?
 
-Mobile applications have grown to be incredibly complicated over the last decade. 
-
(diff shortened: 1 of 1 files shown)


docs/src/markdown/documentation/faq.md
```diff
@@ -20,10 +20,8 @@ Of course! Workflow was designed to make complex application architecture predic
 
 ### Why do we need another architecture?
 
-Mobile applications have grown to be incredibly complicated over the last decade. 
-
-Meanwhile, the system-provided architecture patterns that we use to build those applications is rooted in a simpler time when a handful of developers could fit an entire codebase in their head. This led to design choices that are great for building a todo-list app (subclass a view controller and go!) but are dangerous for more complex cases.
```
zach-klippenstein: This seemed like a nice paragraph, any particular reason why you removed it?

timdonnelly: When introducing a new pattern we need to always err on the side of talking up our strengths without attacking existing patterns that (despite their faults) many people use (or even like).

If someone is coming for the perspective of "the system frameworks are good enough", they might feel attacked by the snarky reference about todo-list apps. I'd rather avoid that conversation if we can.

PR #189, opened 2019-03-18
Add WorkflowUI

timdonnelly:


swift/Samples/SampleApp/Podfile.lock
```diff
@@ -8,10 +8,16 @@ PODS:
   - Workflow/Tests (0.9.0):
     - ReactiveSwift (~> 5.0.0)
     - Result
+  - WorkflowUI (0.9.0):
```
zach-klippenstein: Next time we do a version bump, do we need to change all these numbers too?

swift/WorkflowUI/Sources/Container/UIKitRenderableViewController.swift
```diff
@@ -0,0 +1,81 @@
...
+    init(viewRegistry: ViewRegistry) {
+        self.viewRegistry = viewRegistry
+        super.init(nibName: nil, bundle: nil)
+    }
+
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    override func viewDidLoad() {
```
davidapgar: This `viewDidLoad` doesn't need to be over-ridden (since it's a no-op)

swift/WorkflowUI/Sources/Screen/Screen.swift
```diff
@@ -0,0 +1,15 @@
+import ReactiveSwift
```
davidapgar: I don't think this import is needed (no ReactiveSwift visible in this API)

swift/WorkflowUI/Tests/ContainerViewControllerTests.swift
```diff
@@ -0,0 +1,98 @@
...
+        let container = ContainerViewController(workflow: workflow, viewRegistry: registry)
+
+        var outputCalled = false
+        let disposable = container.output.observeValues { value in
+            XCTAssertEqual(3, value)
+            outputCalled = true
+        }
+
+        observer.send(value: 3)
+        RunLoop.current.run(until: Date().addingTimeInterval(0.1))
```
davidapgar: Should update these tests to use XCTExpectation instead of explicitly spinning the runloop (to reduce the likelihood of flakes if CI is a bit slow)

swift/WorkflowUI/Tests/WorkflowUITests.swift
```diff
@@ -0,0 +1,7 @@
+import XCTest
+
+class WorkflowUITests: XCTestCase {
```
davidapgar: We can get rid of the placeholder, as there are real tests in here.

swift/WorkflowUI/Tests/ContainerViewControllerTests.swift
timdonnelly: Certainly should!

timdonnelly: Done

swift/WorkflowUI/Tests/WorkflowUITests.swift
timdonnelly: Done

swift/WorkflowUI/Sources/Container/UIKitRenderableViewController.swift
timdonnelly: Done

swift/WorkflowUI/Sources/Screen/Screen.swift
timdonnelly: Done

swift/Samples/SampleApp/Podfile.lock
timdonnelly: It isn't strictly necessary, but if they are not updated subsequent `bundle exec pod install` calls will leave a dirty `Podfile.lock`

PR #190, opened 2019-03-18
Add Swift to RELEASING

timdonnelly:


timdonnelly:
@zach-klippenstein I started to write separate Swift releasing instructions, then realized that they need to live together as we are sharing tagged releases.

RELEASING.md
```diff
@@ -1,16 +1,48 @@
...
+1. Make sure you're on the `master` branch.
+2. In `kotlin/gradle.properties`, remove the `-SNAPSHOT` prefix from the `VERSION_NAME` property.
+   E.g. `VERSION_NAME=0.1.0`
+   Update `swift/CHANGELOG.md` and `kotlin/CHANGELOG.md` with the changes since the last release.
+3. Create a commit and tag the commit with the version number:
+   ```
+   git commit -am "Releasing v0.1.0."
+   git tag v0.1.0
+   ```
+4. Upload the artifacts:
```
zach-klippenstein: Would be good to explicitly state that no action is required here for Swift. If you hadn't told me that cocoapods directly hit the repo, I would feel like it was missing something.

timdonnelly: We will eventually have instructions for pushing a new version of the podspec to the master repo, but I haven't bothered to set that up yet.

PR #193, opened 2019-03-19
Add xcode templates

davidapgar:
This includes templates for Workflow and ScreenViewController.


swift/Workflow/Tooling/Templates/install-xcode-templates.sh
```diff
@@ -0,0 +1,22 @@
...
+# Configuration
+XCODE_TEMPLATE_DIR=$HOME'/Library/Developer/Xcode/Templates/File Templates/Workflow'
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+# Copy workflow file templates into the local workflow template directory
+xcodeTemplate () {
+  echo "Copying workflow Xcode file templates..."
+
+  if [ -d "$XCODE_TEMPLATE_DIR" ]; then
+    rm -R "$XCODE_TEMPLATE_DIR"
```
bencochran: Should we be at all worried about people with preexisting `~/Library/Developer/Xcode/Templates/File Templates/Workflow` contents?

davidapgar: Hmm. I've generally liked this, as over time when we've changed the templates it cleans up, and the name of the pod of `Workflow` is unique.

If you feel strongly, I can remove this though.

bencochran: I think the functionality is good. But I worry that `Workflow` is too generic to assume we own it. Maybe a more-namespaced directory would work?

davidapgar: Yeah, I think you're right that it's too generic to make assumptions.

Updated and removed the rm -R

swift/Workflow/Tooling/Templates/Workflow (Verbose).xctemplate/___FILEBASENAME___Workflow.swift
```diff
@@ -0,0 +1,94 @@
...
+// MARK: Input
+
+struct ___VARIABLE_productName___Workflow: Workflow {
+
+}
+
+
+// MARK: Output
+
+extension ___VARIABLE_productName___Workflow {
```
timdonnelly: Now that the protocol is shorter I would bring all of this together (not putting everything in a separate extension).

davidapgar: For all the ones of these I've build, I've kept the separation. Especially since as the complexity grows of a Workflow, the separation I find helps readability

timdonnelly:
I just realized that these are in `swift/Workflow/Tooling`. Let's move them to `swift/Tooling`

davidapgar:
> I just realized that these are in `swift/Workflow/Tooling`. Let's move them to `swift/Tooling`

Good call on that. I moved it up a level, and also rolled `Input` and `Output` together, just to keep it a bit smaller on the verbose example

PR #195, opened 2019-03-19
Add convenience extensions for makeSink and awaitResult

timdonnelly:
This allows for things like

```swift
let sink = context.makeSink(of: String.self) { string, state in
    state.customerName = string
    return nil
}
```

and

```swift
context.awaitResult(for: AccountLoadingWorker()) { account, state in
    state = .loaded(account: account)
}
```


timdonnelly:
FYI @davidapgar @bencochran

PR #200, opened 2019-03-20
Improve type safety by removing AnyScreenViewController as a base type, and making AnyScreen a specific screen type backed by a concrete view controller implementation

timdonnelly:
The unsafe nature of our screens and `AnyScreenViewController` has been bugging me for a while – this makes it type safe by *always* using specific screen types.

If a workflow wants to return some unknown screen type, it can use `AnyScreen`, which is actually a screen implementation that wraps an underlying screen. `AnyScreen` is backed by `AnyScreenViewController`, which updates / replaces the underlying view controller as needed. This removes all of the type checking that was leaking into our container screen implementations.

This *does* mean that a Workflow should always use a specific screen type - no more `compose(...) -> Screen`. If the workflow wants to hide the underlying screen type, it should instead return `compose(...) -> AnyScreen`. `ViewRegistry` will no longer return a view controller when given the `Screen` metatype – it will only accept a concrete instance of a type implementing `Screen`.


rjrjr:
At this point what purpose does the `Screen` type serve? I'd argue it's only a source of confusion at this point. You can put that extension method anywhere.

rjrjr:
I don't really see how this is making anything more type safe. You've just moved the problem around, and adding a confusing bit of wrapping: my workflow can still emit an unknown screen model, and I still won't learn about the mistake until runtime.

timdonnelly:
> I don't really see how this is making anything more type safe. You've just moved the problem around, and added a confusing bit of wrapping: my workflow can still emit an unknown screen model, and I still won't learn about the mistake until runtime.

A workflow *can't* emit an unknown screen model. If always has to emit a known screen model, or it cannot be used.

As a convenience, we provide `AnyScreen` as a *specific* screen type that can show any type of wrapped screen.


The consequence of all of this is that we no longer have to do things like:

```
if type(of: screen) == myViewController.screenType {
    myViewController.update(untypedScreen: screen)
} else {
    myViewController = viewRegistry.provideView(for: screen)
}
```

timdonnelly:
> Update: oh, I think I get it -- you need _something_, because Swift lacks a universal base type?

It's just a marker at this point: it implies that conforming types are meant to be used as screen models with a view registry.

timdonnelly:
@rjrjr Another way to think about this:

We **don't want** to be writing code that requires runtime type checks.

In the old world, screen view controllers looked like this:
```
open class AnyScreenViewController {
    var screenType: Any.Type { get }
    func update(untypedScreen: Screen)
}

open class ScreenViewController<ScreenType: Screen>: AnyScreenViewController
```

The `AnyScreenViewController` API was always unsafe, as `update(untypedScreen:)` could (by contract) only be invoked with a screen of the type returned by `screenType`. The compiler was no help here, however.

In the new world, there is no unsafe `AnyScreenViewController` API on screen view controllers – their entire API is type safe around their specific screen type:

```
open class ScreenViewController<ScreenType: Screen>: UIViewController {
    func update(screen: ScreenType)
}
```

That change brings much more compile time type safety to screen view controllers.

***

However: we sometimes **want** the ability to use unknown screen types (e.g. when forming a back stack – they shouldn't all need to be the same screen type).

This is why `AnyScreen` exists. `AnyScreen` must be initialized with a *specific* type of wrapped screen.

`AnyScreen` is displayed by a corresponding view controller that follows the same type-safe contract as all screens:

```
internal final class AnyScreenViewController: ScreenViewController<AnyScreen>
```

*Internal to the implementation of AnyScreenViewController:*
- Upon initialization, `AnyScreenViewController` uses the wrapped screen type from `AnyScreen` to build a view controller for the wrapped screen (and shows it)
- If `AnyScreenViewController` is updated with an `AnyScreen` containing the same wrapped screen type, it updates the existing wrapped view controller.
- If `AnyScreenViewController` is updated with an `AnyScreen` containing a *different* wrapped screen type, it tears down its current wrapped view controller and builds a new one for the new wrapped screen type.

rjrjr:
Super helpful replies, thanks. Part of my confusion is because we don't have anything like UIViewController on the Android side yet, though the need is becoming apparent.

davidapgar:
I definitely think this is a good change overall - any case we can increase type safety is a benefit.

Will need to explore the fallout of this on our already existing code using this prior to merging, but I don't expect there to be anything too major.

timdonnelly:
@bencochran @davidapgar I'd like to merge this before it conflicts with other changes – any thoughts or objections?

bencochran:
Sorry, took me a while to take the space to internalize the effects of this change. I like that this encourages sharing rendering type information (while still allowing erasure when needed).

timdonnelly:
@davidapgar Ok to merge?

swift/WorkflowUI/Sources/Screen/AnyScreen/AnyScreen.swift
```diff
@@ -0,0 +1,18 @@
...
+
+    public init<T: Screen>(_ screen: T) {
+        self.wrappedScreen = screen
+        self.viewControllerBuilder = { viewRegistry in
+            return viewRegistry.provideView(for: screen)
+        }
+    }
+
+    // Internal method to inflate the wrapped screen from a view registry
+    func makeViewController(from viewRegistry: ViewRegistry) -> AnyScreenViewController.WrappedViewController {
```
davidapgar: With both this and the `AnyScreenViewController` type being internal, how would we go about building containers that can show any screen type?

Right now, it looks like the `ContainerViewController` is the only one that can inflate (since it's within the same module).

Previously, since there was an `inflate` method on `Screen`, the other containers (like backstack, modal, etc) could also handle inflating.

Off the cuff, I would propose adding an inflate method like
```
    public func inflate(from viewRegistry: ViewRegistry) -> ScreenViewController<AnyScreen> {
        return viewRegistry.provideView(for: self)
    }
```

That will still return a `ScreenViewController` for the screen, but I'm open to something else. Just need to have a path for making container that can take `AnyScreen`

Alternatively, could expose the `provideView` on the view registry.

timdonnelly: `provideView` was supposed to be public – that was an oversight. I just added a commit to make it public.

PR #204, opened 2019-03-22
Use explicit KType + String parameters instead of Any for idempotence key for subscriptions.

### LINKED ISSUE #194: Should WorkflowContext extension methods not try to be clever with inlines?
I’m thinking of getting rid of the `idempotenceKey: Any`/`key: String` distinction in `WorkflowContext` and just forcing you to always pass a string key. It sucks and I hate it but the current strategy of generating a fancy idempotence key for you means you can’t call those methods from generic workflows that don’t know their types.

cc @rjrjr @timdonnelly

zach-klippenstein:
This makes the "raw" APIs harder to misuse (`KType` is a specific type), which lets us safely expose
those raw methods for consumers to use if they can't use the inline functions that automatically
calculate the type. We also provide a simple helper for creating a `KType` from a `Foo<Bar>`-style
type.

A use case for this is concrete workflows that are themselves generic, and can't pass their type
arguments to functions expecting reified params.

Closes #194.


zach-klippenstein:
This also gives us better toStrings for the types.

kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowContext.kt
```diff
@@ -74,26 +77,25 @@ interface WorkflowContext<StateT : Any, in OutputT : Any> {
...
    * method with an equal key. At that time, the channel will be [cancelled][ReceiveChannel.cancel],
    * with the assumption that cancelling the channel will release any resources allocated by the
    * [channelProvider].
    *
-   * **This method is intended for library authors**, since a unique idempotence key must be specified
-   * explicitly. Most code will want to call one of the variants that accepts a [String] `key`
-   * parameter (i.e. [onReceive]). See [IdempotenceKey] for factory methods to derive keys from types.
-   *
+   * @param type The [KType] that represents both the type of data source (e.g. `ReceiveChannel` or
+   * `Deferred`) and the element type [E].
```
rjrjr:
```
e.g. `KTypes.fromGenericType(ReceiveChannel::class, E::class)
```

kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowContext.kt
```diff
@@ -74,26 +77,25 @@ interface WorkflowContext<StateT : Any, in OutputT : Any> {
...
+   * @param key A string key that is used to distinguish between subscriptions of the same [type].
```
rjrjr: `An optional string key…`

kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowContext.kt
rjrjr: Is there any way that we can enforce that `E` matches this? Assuming not, just thought I'd ask.

kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowContext.kt
```diff
@@ -205,44 +209,61 @@ fun <StateT : Any, OutputT : Any, ChildStateT : Any, ChildRenderingT : Any>
...
+/**
+ * Will wait for [deferred] to complete, then pass its value to [handler]. Once the handler has been
+ * invoked for a given deferred+key, it will not be invoked again until an invocation of
+ * [Workflow.compose] that does _not_ call this method with that deferred+[key].
+ *
+ * @param type The [KType] that represents both the type of data source (e.g. `Deferred`) and the
+ * element type [T].
+ * @param key A string key that is used to distinguish between subscriptions of the same [type].
+ */
+fun <T, StateT : Any, OutputT : Any> WorkflowContext<StateT, OutputT>.onDeferred(
```
rjrjr: You sure we want this to be public? Wondering if the ones that expose the `type` argument should all be `@PublishedApi` instead.

kotlin/workflow-rx2/src/main/java/com/squareup/workflow/rx2/WorkflowContexts.kt
```diff
@@ -33,23 +34,39 @@ import kotlinx.coroutines.experimental.rx2.openSubscription
...
 
-@PublishedApi
-internal fun <T : Any, StateT : Any, OutputT : Any> WorkflowContext<StateT, OutputT>.onSuccess(
+/**
+ * Subscribes to [single] and invokes [handler] with the value it emits.
+ *
+ * This allows us to use this method to, for example, wait for a response from a service call while
+ * also moving between states in response to UI events. (Use [key] if you need to work simultaneously
+ * with several `Single`s of the same type.)
+ * The subscription will be disposed when `compose` returns without passing that single to this
```
rjrjr: s/`that single`/`a single of the same type and with the same key`/

kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowContext.kt
zach-klippenstein: Nope :(

kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowContext.kt
zach-klippenstein: It's needing to be public is the whole reason I started this PR. The inline functions can't be used by generic workflows.

zach-klippenstein:
> This is great. Outside of a couple of doc nits, main question is whether we really want to make the `KType` variants public, or put them behind `@PublishedApi`.

They need to be public, because generic workflows can't use the reified functions. I didn't want to expose them with the `Any` types, because that's so easy to misuse. Restricting it to `KType` specifically keeps all the flexibility (too much?) but slaps a big scary ADVANCED MODE sticker on them.

kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowContext.kt
rjrjr: Right right right.

PR #207, opened 2019-03-22
Rename makeSink to onEvent, make EventHandler type.

### LINKED ISSUE #203: makeSink describes what the function does, but is not super scannable
`makeSink` describes what the function does, but it doesn't necessarily make the function more readable given how it's used:
```kotlin
return MyScreen(
  onClick = context.makeSink { emitOutput(Foo) },
  onBack = context.makeUnitSink { emitOutput(Bar) }
)
```
 
When you're reading the code, you don't care that the context is "making a sink", you only care that you're trying to specify an event handler. It might be better if the name described the intention of the caller, instead of the way the method fulfills that intention:
```kotlin
return MyScreen(
  onClick = context.handle { emitOutput(Foo) },
  onBack = context.handleUnit { emitOutput(Bar) }
)
```

cc @qrazhan

### LINKED ISSUE #206: Renderings are hard to test because event sinks/handler functions aren't equatable.
It is a common pattern for a workflow's rendering type to be a data class. Data classes automatically get `equals` implementations, so when writing unit tests it's convenient to compare all the fields of two data classes by just creating a new instance with the expected data and doing a simple comparison.

The way event handlers work right now breaks this pattern, because the event handling functions are compared by identity, so they will never be equal.

There are some _very_ gross workarounds:
 1. Compare all fields one-by-one.
 2. Use your favorite assertion library's functionality to ignore certain fields when comparing (e.g. AssertJ lets you do this).
 3. Clone the subject data class and set all its event handlers to some known value.

A more long-term solution is to split the data and the event handlers into separate types, more like the legacy `Screen`, which is basically a tuple of screen data + event handler. This solution requires more boilerplate, especially if you still want to keep different handlers separate for different event types.

cc @qrazhan

zach-klippenstein:
This also means we can get rid of the separate `makeUnitSink` function, because we can
just define an override `invoke` for `EventHandler<Unit>`.

Closes #206 and closes #203.

### DIFF
--- a/kotlin/workflow-core/src/main/java/com/squareup/workflow/EventHandler.kt
+++ b/kotlin/workflow-core/src/main/java/com/squareup/workflow/EventHandler.kt
@@ -1,0 +1,38 @@
+package com.squareup.workflow
+
+/**
+ * Wraps a function that handles an event of type [EventT]. Implements the raw function type
+ * `(EventT) -> Unit`, so instances can be invoked and passed around as functions.
+ *
+ * This type differs from a typical function reference in how it defines equality. All instances
+ * of this class are considered equal (and have the same hashcode), which means they can be stored
+ * in data classes and the event handlers will be ignored. This makes it much easier to write
+ * test assertions for [Workflow] renderings that include event handlers, since you can compare
+ * entire rendering types at once and not field-by-field.
+ */
+class EventHandler<in EventT>(private val handler: (EventT) -> Unit) : (EventT) -> Unit {
+
+  override fun invoke(event: EventT) = handler(event)
+
+  /**
+   * Returns `true` iff `other` is an [EventHandler] – all [EventHandler]s are considered equal.
+   */
+  override fun equals(other: Any?): Boolean = when {
+    other === this -> true
+    other is EventHandler<*> -> true
+    else -> false
+  }
+
+  /**
+   * All [EventHandler]s have the same hashcode, since they are all equal.
+   */
+  override fun hashCode(): Int = 0
+
+  override fun toString(): String = "EventHandler@${super.hashCode()}"
+}
+
+/**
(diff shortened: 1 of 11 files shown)


kotlin/workflow-core/src/main/java/com/squareup/workflow/EventHandler.kt
```diff
@@ -0,0 +1,47 @@
+package com.squareup.workflow
+
+/**
+ * TODO kdoc
+ */
+abstract class EventHandler<in EventT> : (EventT) -> Unit {
```
rjrjr: Does this really need to be an abstract class? Seems like it could be final and wrap a function pointer. I ask b/c its so dicey to have equality of mismatching types, and seems unnecessary.

kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowContext.kt
```diff
@@ -28,7 +28,7 @@ import kotlinx.coroutines.experimental.suspendCancellableCoroutine
  *
  * ## Handling Events
  *
- * See [makeSink], [makeUnitSink].
+ * See [onEvent], [makeUnitSink].
```
rjrjr: You killed `makeUnitSink`, right? Still a lot of references to it.

kotlin/workflow-core/src/main/java/com/squareup/workflow/EventHandler.kt
zach-klippenstein: Good point. Will make final.

kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowContext.kt
zach-klippenstein: You mean we have comments? How did those get in there!

PR #208, opened 2019-03-22
Updated changelogs for 0.9.0.

zach-klippenstein:


zach-klippenstein:
@davidapgar @timdonnelly Still would like one of yall's signoff on the Swift changes – I just pulled stuff from the commit history, not sure if it's all relevant.

PR #214, opened 2019-03-26
Remove knowledge of child StateT type from compose.

### LINKED ISSUE #213: WorkflowContext.compose shouldn't have to know child workflow's state type.
This was only required back when we internally keyed off of the tuple of type parameters. Since we use the actual `KClass` of `Workflow` implementations now, we don't need to care anymore.

zach-klippenstein:
Closes #213.

### DIFF
--- a/kotlin/workflow-host/src/main/java/com/squareup/workflow/internal/WorkflowNode.kt
+++ b/kotlin/workflow-host/src/main/java/com/squareup/workflow/internal/WorkflowNode.kt
@@ -15,13 +15,13 @@
  */
 package com.squareup.workflow.internal
 
-import com.squareup.workflow.util.ChannelUpdate.Closed
 import com.squareup.workflow.Snapshot
 import com.squareup.workflow.Workflow
 import com.squareup.workflow.WorkflowAction
 import com.squareup.workflow.internal.Behavior.SubscriptionCase
 import com.squareup.workflow.parse
 import com.squareup.workflow.readByteStringWithLength
+import com.squareup.workflow.util.ChannelUpdate.Closed
 import com.squareup.workflow.writeByteStringWithLength
 import kotlinx.coroutines.experimental.CoroutineName
 import kotlinx.coroutines.experimental.CoroutineScope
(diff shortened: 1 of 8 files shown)


kotlin/workflow-host/src/main/java/com/squareup/workflow/internal/WorkflowNode.kt
```diff
@@ -38,7 +38,7 @@ import kotlin.coroutines.experimental.CoroutineContext
  * [Workflow.initialState].
  */
 internal class WorkflowNode<InputT : Any, StateT : Any, OutputT : Any, RenderingT : Any>(
```
rjrjr: If `WorkflowContext.compose` no longer knows about `StateT`, how is `WorkflowNode` learning about it? Where does it come from?

zach-klippenstein: It basically doesn't, it's effectively `Any`. But it doesn't need to in practice. The generic param is basically just there so the `Workflow` methods all agree.

kotlin/workflow-host/src/main/java/com/squareup/workflow/internal/WorkflowNode.kt
```diff
@@ -91,22 +91,12 @@ internal class WorkflowNode<InputT : Any, StateT : Any, OutputT : Any, Rendering
...
-
-    behavior = context.buildBehavior()
-    // Start new children/subscriptions, and drop old ones.
-    subtreeManager.track(behavior.childCases)
-    subscriptionTracker.track(behavior.subscriptionCases)
-
-    return rendering
-  }
+  ): RenderingT =
+    composeWithStateType(workflow as Workflow<InputT, StateT, OutputT, RenderingT>, input)
```
zach-klippenstein: @rjrjr The cast you're looking for is here.

kotlin/workflow-host/src/main/java/com/squareup/workflow/internal/WorkflowNode.kt
rjrjr: Let me see if I understand: the `StateT` of the root `WorkflowNode` comes from `WorkflowHost.Factory#run$workflow_host`. But of every other `WorkflowNode` in the tree has `StateT` of `Any`? Given that, can you push this PR farther and get rid of `StateT` from `WorkflowNode` and `RealWorkflowContext`?

zach-klippenstein:
Declining in favor of #219.

PR #215, opened 2019-03-27
Suppress valid deprecation and unchecked cast warnings.

zach-klippenstein:
These have been annoying the hell out of me.


kotlin/legacy/legacy-workflow-rx2/src/test/java/com/squareup/workflow/legacy/rx2/CoroutineEventChannelTest.kt
```diff
@@ -307,6 +310,7 @@ class CoroutineEventChannelTest {
 
   @Test fun `onWorkerResult succeeds`() {
     val trigger = CompletableSubject.create()
+    @Suppress("RedundantLambdaArrow")
     val worker = singleWorker { it: String ->
```
rjrjr: nit: I'd just have named the thing.

PR #219, opened 2019-03-27
Factor out a common parent interface for StatelessWorkflow and Workflow (now StatefulWorkflow).

### LINKED ISSUE #213: WorkflowContext.compose shouldn't have to know child workflow's state type.

zach-klippenstein:
tl;dr:
 - `Workflow` -> `StatefulWorkflow`
 - `Workflow` is an interface, `StatefulWorkflow` and `StatelessWorkflow` are abstract classes
   that both implement `Workflow`.

This change makes it much easier for consumers of this library to define public APIs with private
implementations. The public API is an interface that extends `Workflow`. The private
implementation is a class that subclasses either `StatefulWorkflow` or `StatelessWorkflow`,
depending on whether it needs to keep track of any state. Either way, the workflow's state type
is kept as a completely private concern.

`interface Workflow<I, O, R>`
-----------------------------

This is the main currency of this library. `Workflow`s can compose
other `Workflow`s. A `Workflow` is anything that can be expressed as a `StatefulWorkflow`.
It has a single method, `asStatefulWorkflow()`, that is responsible for expressing the
workflow in terms of a `StatefulWorkflow` – a workflow that has an internal state. This
commit renames what used to be `Workflow` to `StatefulWorkflow`.

`abstract class StatefulWorkflow<I, S, O, R>`
---------------------------------------------

This is what you subclass to write a `Workflow` that has internal state that persists
as long as the workflow is active in its parent. It knows about things like initial state
and snapshotting/restoring.

`abstract class StatelessWorkflow<I, O, R>`
-------------------------------------------

This is what you subclass to get a `Workflow` that only has a single `compose` method.
It can compose children workflows, transform inputs/outputs/renderings up/down the tree,
and listen to external events and subscriptions, but it has no state.

Closes #213.

Based on #214 

[Internal discussion](https://square.slack.com/archives/CE470RL5T/p1553704080224400)

### DIFF
--- a/kotlin/workflow-core/src/main/java/com/squareup/workflow/StatefulWorkflow.kt
+++ b/kotlin/workflow-core/src/main/java/com/squareup/workflow/StatefulWorkflow.kt
@@ -1,0 +1,141 @@
+/*
+ * Copyright 2019 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.workflow
+
+/**
+ * A composable, stateful object that can [handle events][WorkflowContext.onEvent],
+ * [delegate to children][WorkflowContext.compose], [subscribe][onReceive] to arbitrary streams from
+ * the outside world, and be [saved][snapshotState] to a serialized form to be
+ * [restored][restoreState] later.
+ *
+ * The basic purpose of a `Workflow` is to take some [input][InputT] and return a
+ * [rendering][RenderingT]. To that end, a workflow may keep track of internal [state][StateT],
+ * recursively ask other workflows to render themselves, subscribe to data streams from the outside
+ * world, and handle events both from its [renderings][WorkflowContext.onEvent] and from workflows
+ * it's delegated to (its "children"). A `Workflow` may also emit [output events][OutputT] up to its
+ * parent `Workflow`.
+ *
+ * Workflows form a tree, where each workflow can have zero or more child workflows. Child workflows
+ * are started as necessary whenever another workflow asks for them, and are cleaned up automatically
+ * when they're no longer needed. [Input][InputT] propagates down the tree, [outputs][OutputT] and
+ * [renderings][RenderingT] propagate up the tree.
+ *
+ * @param InputT Typically a data class that is used to pass configuration information or bits of
+ * state that the workflow can always get from its parent and needn't duplicate in its own state.
+ * May be [Unit] if the workflow does not need any input data.
+ *
+ * @param StateT Typically a data class that contains all of the internal state for this workflow.
+ * The state is seeded via [input][InputT] in [initialState]. It can be [serialized][snapshotState]
+ * and later used to [restore][restoreState] the workflow. **Implementations of the `Workflow`
+ * interface should not generally contain their own state directly.** They may inject objects like
+ * instances of their child workflows, or network clients, but should not contain directly mutable
+ * state. This is the only type parameter that a parent workflow needn't care about for its children,
+ * and may just use star (`*`) instead of specifying it. May be [Unit] if the workflow does not have
+ * any internal state (see [StatelessWorkflow]).
+ *
+ * @param OutputT Typically a sealed class that represents "events" that this workflow can send
+ * to its parent.
+ * May be [Nothing] if the workflow doesn't need to emit anything.
+ *
+ * @param RenderingT The value returned to this workflow's parent during [composition][compose].
+ * Typically represents a "view" of this workflow's input, current state, and children's renderings.
+ * A workflow that represents a UI component may use a view model as its rendering type.
+ *
+ * @see StatelessWorkflow
+ */
+abstract class StatefulWorkflow<
+    in InputT : Any,
+    StateT : Any,
+    out OutputT : Any,
+    out RenderingT : Any
+    > : Workflow<InputT, OutputT, RenderingT> {
+
+  /**
+   * Called from [WorkflowContext.compose] when the state machine is first started, to get the
+   * initial state.
+   */
+  abstract fun initialState(input: InputT): StateT
+
+  /**
+   * Called from [WorkflowContext.compose] instead of [initialState] when the workflow is already
+   * running. This allows the workflow to detect changes in input, and possibly change its state in
+   * response. This method is called eagerly: `old` and `new` might be the same value, so it is up
+   * to implementing code to perform any diffing if desired.
+   *
+   * Default implementation does nothing.
+   */
+  open fun onInputChanged(
+    old: InputT,
+    new: InputT,
+    state: StateT
+  ): StateT = state
+
+  /**
+   * Called at least once† any time one of the following things happens:
+   *  - This workflow's [input] changes (via the parent passing a different one in).
+   *  - This workflow's [state] changes.
+   *  - A descendant (immediate or transitive child) workflow:
+   *    - Changes its internal state.
+   *    - Emits an output.
+   *
+   * **Never call this method directly.** To nest the rendering of a child workflow in your own,
+   * pass the child and any required input to [WorkflowContext.compose].
+   *
+   * This method *should not* have any side effects, and in particular should not do anything that
+   * blocks the current thread. It may be called multiple times for the same state. It must do all its
+   * work by calling methods on [context].
+   *
+   * _† This method is guaranteed to be called *at least* once for every state, but may be called
+   * multiple times. Allowing this method to be invoked multiple times makes the internals simpler._
+   */
+  abstract fun compose(
+    input: InputT,
+    state: StateT,
+    context: WorkflowContext<StateT, OutputT>
+  ): RenderingT
+
+  /**
+   * Called whenever the state changes to generate a new [Snapshot] of the state.
+   *
+   * **Snapshots must be lazy.**
+   *
+   * Serialization must not be done at the time this method is called,
+   * since the state will be snapshotted frequently but the serialized form may only be needed very
+   * rarely.
+   *
+   * If the workflow does not have any state, or should always be started from scratch, return
+   * [Snapshot.EMPTY] from this method.
+   *
+   * @see restoreState
+   */
+  abstract fun snapshotState(state: StateT): Snapshot
+
+  /**
+   * Deserialize a state value from a [Snapshot] previously created with [snapshotState].
+   *
+   * If the workflow should always be started from scratch, this method can just ignore the snapshot
+   * and return the initial state.
+   *
+   * @see snapshotState
+   */
+  abstract fun restoreState(snapshot: Snapshot): StateT
+
+  /**
+   * Satisfies the [Workflow] interface by returning `this`.
+   */
+  final override fun asStatefulWorkflow(): StatefulWorkflow<InputT, StateT, OutputT, RenderingT> =
+    this
+}
--- a/kotlin/workflow-core/src/main/java/com/squareup/workflow/StatelessWorkflow.kt
+++ b/kotlin/workflow-core/src/main/java/com/squareup/workflow/StatelessWorkflow.kt
@@ -16,15 +16,80 @@
 package com.squareup.workflow
 
 /**
- * A [Workflow] that has no state.
+ * A composable object that can [handle events][WorkflowContext.onEvent],
+ * [delegate to children][WorkflowContext.compose], and [subscribe][onReceive] to arbitrary streams
+ * from the outside world.
+ *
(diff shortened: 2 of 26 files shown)


kotlin/workflow-core/src/main/java/com/squareup/workflow/StatelessWorkflow.kt
```diff
@@ -17,14 +17,43 @@ package com.squareup.workflow
 
 /**
  * A [Workflow] that has no state.
+ *
+ * @see StatefulWorkflow
  */
-typealias StatelessWorkflow<InputT, OutputT, RenderingT> =
-    Workflow<InputT, Unit, OutputT, RenderingT>
+interface StatelessWorkflow<InputT : Any, OutputT : Any, RenderingT : Any> :
```
rjrjr: I don't want to increase the scope of this PR, more a discussion of possible follow up work: I still find myself wondering if the code would be simpler (both for us to maintain and for others to consume) if `StatelessWorkflow` and `StatefulWorkflow` were abstract classes.

kotlin/workflow-core/src/main/java/com/squareup/workflow/StatefulWorkflow.kt
```diff
@@ -0,0 +1,132 @@
...
+ * to its parent.
+ * May be [Nothing] if the workflow doesn't need to emit anything.
+ *
+ * @param RenderingT The value returned to this workflow's parent during [composition][compose].
+ * Typically represents a "view" of this workflow's input, current state, and children's renderings.
+ * A workflow that represents a UI component may use a view model as its rendering type.
+ *
+ * @see StatelessWorkflow
+ */
+interface StatefulWorkflow<in InputT : Any, StateT : Any, out OutputT : Any, out RenderingT : Any> :
```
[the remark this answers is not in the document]
rjrjr: Pretty sure I'm kidding about that, but not completely.

kotlin/workflow-core/src/main/java/com/squareup/workflow/StatelessWorkflow.kt
zach-klippenstein:
```kotlin
sealed class Workflow<I, O, R> {
  internal abstract fun asStatefulWorkflow(): StatefulWorkflow<I, *, O, R>
}

abstract class StatefulWorkflow<I, S, O, R> : Workflow<I, O, R>() {
  final override internal fun asStatefulWorkflow() = this

  abstract fun initialState(…
}

abstract class StatelessWorkflow<I, O, R> : Workflow<I, O, R>() {
  final override internal fun asStatefulWorkflow() = object : …
  
  abstract fun compose(I, …
}
```

zach-klippenstein: So using abstract classes (without a sealed parent) gives us the ability to make the `asStatefulWorkflow` override final in the 2 children, which would be nice.

Making them _sealed_ lets us get rid of that function entirely. We could have a separate `WorkflowNode` just for stateless workflows, which would be much simpler than the stateful one and could use less memory.

Those are the only two advantages of using abstract classes over interfaces that I can think of. Did you have anything else in mind @rjrjr?

zach-klippenstein: Trying to think if there are any disadvantages to using abstract classes…

The only thing is that it limits multiple inheritance, but I'm not sure that matters. In fact, it's probably even a nice benefit since other classes shouldn't be implementing workflow types anyway.

zach-klippenstein: Nvm about sealed classes - `Workflow` can't be a class because it would prevent consumers from using the public-interface-private-impl pattern (can't extend both `AuthWorkflow` and `StatefulWorkflow` if they're both classes).

rjrjr: So excited!

kotlin/workflow-core/src/main/java/com/squareup/workflow/StatefulWorkflow.kt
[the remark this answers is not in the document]
zach-klippenstein: 🤔

kotlin/workflow-core/src/main/java/com/squareup/workflow/StatelessWorkflow.kt
zach-klippenstein: This PR now also makes `StatelessWorkflow` and `StatefulWorkflow` abstract classes!

zach-klippenstein:
Fleshed out the docs, added a commit message, this is now ready for real review. When it's ready, I'll squash the first commit out and decline #214.

cc @rjrjr

kotlin/workflow-core/src/main/java/com/squareup/workflow/StatefulWorkflow.kt
```diff
@@ -0,0 +1,136 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.workflow
+
+/**
+ * A composable, stateful object that can [handle events][WorkflowContext.onEvent],
```
rjrjr: The docs of the base classes and the interface seem unnecessarily redundant. I don't think a big deal for this PR, but something I'm planning to make a pass on once I'm done with the samples — heads up.

kotlin/workflow-core/src/main/java/com/squareup/workflow/StatefulWorkflow.kt
```diff
@@ -0,0 +1,136 @@
...
+ */
+abstract class StatefulWorkflow<
+    in InputT : Any,
+    StateT : Any,
+    out OutputT : Any,
+    out RenderingT : Any
+    > : Workflow<InputT, OutputT, RenderingT> {
+
+  /**
+   * Called when the state machine is first started to get the initial state.
```
rjrjr: Called from `WorkflowContext.compose` when a state machine is first started…

kotlin/workflow-core/src/main/java/com/squareup/workflow/StatefulWorkflow.kt
```diff
@@ -0,0 +1,136 @@
...
+   */
+  abstract fun initialState(input: InputT): StateT
+
+  /**
+   * Called whenever [WorkflowContext.compose] is about to get a new [input][InputT] value, to allow
```
rjrjr: Called from `WorkflowContext.compose` instead of `initialState` when a workflow is already running. This 
allows the workflow to detect changes in input, and possibly change its state in response.

kotlin/workflow-core/src/main/java/com/squareup/workflow/StatefulWorkflow.kt
```diff
@@ -0,0 +1,136 @@
...
+    old: InputT,
+    new: InputT,
+    state: StateT
+  ): StateT = state
+
+  /**
+   * Called at least once† any time one of the following things happens:
+   *  - This workflow's [input] changes (via the parent passing a different one in).
+   *  - This workflow's [state] changes.
+   *  - A child workflow's state changes.
```
rjrjr: Instead of _A child workflow's state changes_, how about _A child workflow emits output, or updates its rendering._

kotlin/workflow-core/src/main/java/com/squareup/workflow/StatefulWorkflow.kt
```diff
@@ -0,0 +1,136 @@
...
+   *
+   * **Never call this method directly.** To get the rendering from a child workflow, pass the child
```
rjrjr: To nest the rendering of a child workflow in your own,

kotlin/workflow-core/src/main/java/com/squareup/workflow/StatefulWorkflow.kt
```diff
@@ -0,0 +1,136 @@
...
+
+  /**
+   * Deserialize a state value from a [Snapshot] previously created with [snapshotState].
+   *
+   * If the workflow should always be started from scratch, this method can just ignore the snapshot
+   * and return the initial state.
+   *
+   * @see snapshotState
+   */
+  abstract fun restoreState(snapshot: Snapshot): StateT
```
rjrjr: See #220

kotlin/workflow-core/src/main/java/com/squareup/workflow/StatelessWorkflow.kt
```diff
@@ -16,15 +16,78 @@
...
  */
-typealias StatelessWorkflow<InputT, OutputT, RenderingT> =
-    Workflow<InputT, Unit, OutputT, RenderingT>
+abstract class StatelessWorkflow<InputT : Any, OutputT : Any, RenderingT : Any> :
+    Workflow<InputT, OutputT, RenderingT> {
+
+  /**
+   * Called at least once any time one of the following things happens:
+   *  - This workflow's [input] changes (via the parent passing a different one in).
+   *  - A child workflow updates in some way (e.g. a stateful child's state changes).
```
rjrjr: s/_(e.g.…)/(i.e., it emits output or updates its rendering)/

And add a bullet about workers?

kotlin/workflow-core/src/main/java/com/squareup/workflow/Workflow.kt
```diff
@@ -18,107 +18,55 @@ package com.squareup.workflow
...
-  /**
-   * Deserialize a state value from a [Snapshot] previously created with [snapshotState].
-   *
-   * If the workflow should always be started from scratch, this method can just ignore the snapshot
-   * and return the initial state.
-   *
-   * @see snapshotState
-   */
-  fun restoreState(snapshot: Snapshot): StateT
+  fun asStatefulWorkflow(): StatefulWorkflow<InputT, *, OutputT, RenderingT>
```
rjrjr: Provides a `StatefulWorkflow` view of the receiver. Necessary because `StatefulWorkflow` is the common API required for `WorkflowContext.compose()` to do its work.

kotlin/workflow-core/src/main/java/com/squareup/workflow/Workflows.kt
```diff
@@ -24,26 +24,10 @@ import com.squareup.workflow.WorkflowAction.Companion.emitOutput
...
       transform: (FromRenderingT) -> ToRenderingT
-    ): StatelessWorkflow<InputT, OutputT, ToRenderingT> = StatelessWorkflow { input, context ->
+    ): Workflow<InputT, OutputT, ToRenderingT> = StatelessWorkflow { input, context ->
 // @formatter:on
   context.compose(this@mapRendering, input) { emitOutput(it) }
       .let(transform)
 }
-
-/**
- * Returns a rendering that directly delegates to this [Workflow] but masks its state type.
```
rjrjr: 🔥

kotlin/workflow-core/src/main/java/com/squareup/workflow/StatefulWorkflow.kt
zach-klippenstein: Yea, I'm not a fan either. Probably fine to put most of the docs on `Workflow`, point both children to it, and then only have state details on `StatefulWorkflow`.

kotlin/workflow-core/src/main/java/com/squareup/workflow/StatefulWorkflow.kt
zach-klippenstein: But state changes _do_ trigger composition, and rendering changes don't (because "render" is just another word for this method). Updating to include both state and output at any rate.

kotlin/workflow-core/src/main/java/com/squareup/workflow/StatelessWorkflow.kt
zach-klippenstein: What would you like to see about workers? That term doesn't exist in any of the v2 code yet. I've updated this comment to match the updated wording for `StatefulWorkflow`.

kotlin/workflow-core/src/main/java/com/squareup/workflow/Workflow.kt
zach-klippenstein: And this is why i always leave `TODO kdoc` comments everywhere…

zach-klippenstein:
Thanks for the merge conflict fixes @rjrjr 😁 Gonna merge when CI goes green.

kotlin/workflow-core/src/main/java/com/squareup/workflow/StatefulWorkflow.kt
rjrjr: Ah, right right.

kotlin/workflow-core/src/main/java/com/squareup/workflow/StatelessWorkflow.kt
rjrjr: _ That term doesn't exist in any of the v2 code yet_ Excellent.

zach-klippenstein:
Travis is having issues:
image

I ran `./gradlew clean connectedCheck` locally, everything passed, so I'm just gonna merge.

PR #226, opened 2019-03-28
Moves StatelessWorkflow, Workflows.kt methods to Workflow.

rjrjr:
Hangs methods off extensions of Workflow.companion.
Deletes Workflows.kt.


kotlin/workflow-core/src/main/java/com/squareup/workflow/Workflow.kt
```diff
@@ -73,4 +75,63 @@ interface Workflow<in InputT : Any, out OutputT : Any, out RenderingT : Any> {
...
+
+/**
+ * Uses the given [function][transform] to transform a [Workflow] that
+ * renders [FromRenderingT] to one renders [ToRenderingT],
+ */
+// Intellij refuses to format this parameter list correctly because of the weird line break,
+// and detekt will complain about it.
+// @formatter:off
+fun <InputT : Any, OutputT : Any, FromRenderingT : Any, ToRenderingT : Any>
+    Workflow<InputT, OutputT, FromRenderingT>.mapRendering(
```
zach-klippenstein: Wait where did this live before? I don't see it being deleted from anywhere in this PR.

kotlin/workflow-core/src/main/java/com/squareup/workflow/Workflow.kt
```diff
@@ -73,4 +75,63 @@ interface Workflow<in InputT : Any, out OutputT : Any, out RenderingT : Any> {
...
+
+    /**
+     * A convenience function to implement [stateless] by just passing the
+     * [compose][stateless.compose] function as a lambda.
+     *
+     * Note that while a stateless workflow doesn't have any _internal_ state of its own, it may use
+     * [input][InputT] received from its parent, and it may compose child workflows that do have their own
+     * internal state.
+     */
+    fun <InputT : Any, OutputT : Any, RenderingT : Any> stateless(
```
zach-klippenstein: How would you feel about keeping `Workflow`'s companion object empty and put these functions back in `StatelessWorkflow.kt` as extension methods on it?

rjrjr: I'm really liking `Workflow.stateless`.

kotlin/workflow-core/src/main/java/com/squareup/workflow/Workflow.kt
rjrjr: Only method in Workflows.kt, the deleted file is collapsed in the diff.

kotlin/workflow-core/src/main/java/com/squareup/workflow/Workflow.kt
rjrjr: Even find myself wondering if we can make that abstract class internal.

kotlin/workflow-core/src/main/java/com/squareup/workflow/Workflow.kt
zach-klippenstein: Ah, derp

kotlin/workflow-core/src/main/java/com/squareup/workflow/Workflow.kt
zach-klippenstein: I like that too, I want to keep it I just want to make it an extension function to keep this file more focused.

zach-klippenstein: And we could make it internal, but it might be more convenient to just subclass it if you want to take constructor params from Dagger, e.g.

kotlin/workflow-core/src/main/java/com/squareup/workflow/Workflow.kt
zach-klippenstein: Anyway much prefer this method living in this file, thanks!

zach-klippenstein: Anyway much prefer this method living in this file, thanks!

kotlin/workflow-core/src/main/java/com/squareup/workflow/Workflow.kt
```diff
@@ -73,4 +75,63 @@ interface Workflow<in InputT : Any, out OutputT : Any, out RenderingT : Any> {
...
+    ): StatelessWorkflow<Unit, OutputT, RenderingT> =
+      stateless { _: Unit, context: WorkflowContext<Nothing, OutputT> ->
+        compose(context)
+      }
+
+    /**
+     * A version of [stateless] that does nothing but compose a given [rendering].
+     * Handy for testing.
+     */
+    fun <OutputT : Any, RenderingT : Any> stateless(
```
zach-klippenstein: I’d rather call this variant something other than stateless, like `rendering` or something, since it’s not only stateless it’s also behaviorless.

kotlin/workflow-rx2/src/test/java/com/squareup/workflow/rx2/WorkflowContextsTest.kt
```diff
@@ -41,7 +42,7 @@ class WorkflowContextsTest {
     val single = singleSubject
         .doOnSubscribe { subscriptions++ }
         .doOnDispose { disposals++ }
-    val workflow = StatelessWorkflow<String, Unit> { context ->
+    val workflow = Workflow.stateless<String, Unit> { context ->
```
zach-klippenstein: 💯

kotlin/workflow-core/src/main/java/com/squareup/workflow/StatelessWorkflow.kt
```diff
@@ -110,12 +95,39 @@ fun <InputT : Any, OutputT : Any, RenderingT : Any> StatelessWorkflow(
...
+ */
+// Intellij refuses to format this parameter list correctly because of the weird line break,
+// and detekt will complain about it.
+// @formatter:off
+fun <InputT : Any, OutputT : Any, FromRenderingT : Any, ToRenderingT : Any>
+    Workflow<InputT, OutputT, FromRenderingT>.mapRendering(
+      transform: (FromRenderingT) -> ToRenderingT
+    ): Workflow<InputT, OutputT, ToRenderingT> = Workflow.stateless { input, context ->
+  // @formatter:on
+  context.compose(this@mapRendering, input) { WorkflowAction.emitOutput(it) }
```
zach-klippenstein: Nit: static import `emitOutput`

zach-klippenstein:
:shipit: :shipit: :shipit:

rjrjr:
Whatever, Travis. Passes locally, and the identical branch CI passed ages ago.

PR #232, opened 2019-03-29
Fixed broken kdoc links and some other warnings.

zach-klippenstein:
Also added configuration to generate dokka output for some modules.

### DIFF
--- a/kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowContext.kt
+++ b/kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowContext.kt
@@ -44,9 +44,9 @@ import kotlin.reflect.KType
 interface WorkflowContext<StateT : Any, in OutputT : Any> {
 
   /**
-   * Given a function that takes an [event][EventT] and can mutate the state or emit an output, returns
-   * a function that will perform that workflow update when called with an event.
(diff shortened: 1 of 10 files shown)


kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowContext.kt
```diff
@@ -101,18 +101,21 @@ interface WorkflowContext<StateT : Any, in OutputT : Any> {
...
    * Ensures [child] is running as a child of this workflow, and returns the result of its
-   * [compose][Workflow.compose] method.
+   * `compose` method.
    *
-   * **Never call [Workflow.compose] directly, always do it through this context method.**
+   * **Never call [StatefulWorkflow.compose] or [StatelessWorkflow.compose] directly, always do it
+   * through this context method.**
    *
    * 1. If the child _wasn't_ already running, it will be started either from
    *    [initialState][Workflow.initialState] or its snapshot.
```
rjrjr: nit: no longer really an _or_ situation, since we unified the methods.

zach-klippenstein: Those are still related though, and I want to avoid confusion about people trying to call them directly.

PR #239, opened 2019-03-29
Renames ui modules to workflow-ui-core, workflow-ui-android.

rjrjr:


kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/backstack/BackStackEffect.kt
```diff
@@ -1,15 +1,15 @@
...
-import com.squareup.viewregistry.ViewRegistry
-import com.squareup.viewregistry.backstack.ViewStateStack.Direction
-import com.squareup.viewregistry.buildScene
-import com.squareup.viewregistry.buildView
+import com.squareup.workflow.ui.BackStackScreen
+import com.squareup.workflow.ui.BackStackScreen.Key
+import com.squareup.workflow.ui.ViewBinding
+import com.squareup.workflow.ui.ViewRegistry
+import com.squareup.workflow.ui.backstack.ViewStateStack.Direction
+import com.squareup.workflow.ui.buildScene
```
zach-klippenstein: Is `com.squareup.workflow.ui.android.*` too long a package name? It would be nice if the package names corresponded to the module names.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/backstack/NoEffect.kt
```diff
@@ -44,7 +44,9 @@ class NoEffect(
     setUpNewView: (View) -> Unit,
     direction: Direction
   ) {
-    execute(to, screens, viewRegistry, container, setUpNewView)
+    execute(
+        to, screens, viewRegistry, container, setUpNewView
+    )
```
zach-klippenstein: Nit: unintentional reformat?

kotlin/workflow-ui-android/gradle.properties
```diff
@@ -13,6 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-POM_ARTIFACT_ID=viewregistry-android
-POM_NAME=View Registry Android
+POM_ARTIFACT_ID=workflow-ui-android
+POM_NAME=Workflow UI Android
```
zach-klippenstein: 👍

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/backstack/NoEffect.kt
rjrjr: Thanks, IntelliJ!

zach-klippenstein:
You've inspired me to do #240.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/backstack/BackStackEffect.kt
rjrjr: Not so much that it's too long as that it's kind of an orthogonal concern, kind of redundant. Is that extra package name really useful to the android developer, or would it be cognitive clutter?

I did the same thing in the sample code -- the coordinators in the android module are in the same packages as the related screen objects in the common module.

rjrjr: I'll merge this as-is, and follow up if you feel strongly.

rjrjr:
Updates made unintentional reformat intentional / better, and rebased to keep up w/master

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/backstack/BackStackEffect.kt
zach-klippenstein: Yea, I don't feel strongly. And if you're writing a UI for non-android, you're not going to depend on both android and javafx (e.g.) artifacts anyway.

PR #258, opened 2019-04-04
Only allocate one StatefulWorkflow per StatelessWorkflow instance.

zach-klippenstein:


kotlin/workflow-core/src/main/java/com/squareup/workflow/StatelessWorkflow.kt
```diff
@@ -61,21 +77,7 @@ abstract class StatelessWorkflow<InputT : Any, OutputT : Any, RenderingT : Any>
    * state.
    */
   final override fun asStatefulWorkflow(): StatefulWorkflow<InputT, *, OutputT, RenderingT> =
```
zach-klippenstein: This method is called a few times per instance, but we don't need to allocate a new StatefulWorkflow every time.

PR #262, opened 2019-04-10
Add extension on `Screen` to wrap as `AnyScreen` or passthrough if it already is

davidapgar:
While converting over some containers view controllers, I found that often I would either have to explicitly check if something was already an AnyScreen, or be specific with the generic passed in, ie:

```
struct ContainerScreen {
  var displayedScreen: AnyScreen

  public init<ScreenType: Screen>(displayedScreen: ScreenType) {
    self.displayedScreen = AnyScreen(displayedScreen) // non-ideal, since it might already be an AnyScreen.
  // OR
    if let anyScreen = displayedScreen as? AnyScreen {
      self.displayedScreen = anyScreen
    } else {
      self.displayedScreen = AnyScreen(displayedScreen) // non-ideal, since it might already be an AnyScreen.
    }
  }
}
```

Adding this extension on screen allows it to become:
```
  public init(displayedScreen: Screen) {
    self.displayedScreen = displayedScreen.asAnyScreen() // If it already is, this is a nop, otherwise wrap it.
  }
```

This is also useful for a case where we need to pass along an `AnyScreen` out of a compose method (where two or more children will return different screens) using chaining.

```
compose(...) -> AnyScreen {
  switch state {
    case first:
      // FirstChildWorkflow's Rendering is "FirstScreen"
      return FirstChildWorkflow().rendered(with: context).asAnyScreen()
    case second:
      // SecondChildWorkflow's Rendering is "SecondScreen"
      return SecondChildWorkflow().rendered(with: context).asAnyScreen()
  }
}
```

### DIFF
--- a/swift/WorkflowUI/Sources/Screen/AnyScreen/AnyScreen.swift
+++ b/swift/WorkflowUI/Sources/Screen/AnyScreen/AnyScreen.swift
@@ -16,3 +16,14 @@ public struct AnyScreen: Screen {
         return viewControllerBuilder(viewRegistry)
     }
 }
+
+
+extension Screen {
+    public func asAnyScreen() -> AnyScreen {
+        if let anyScreen = self as? AnyScreen {
+            return anyScreen
+        } else {
+            return AnyScreen(self)
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


timdonnelly:
First question: can you make your `Rendering` types explicit instead of using `AnyScreen`? I know we were using just `Screen` in many cases where we could have been more explicit

If we *do* want to keep this behavior, I would stick with an initializer instead of an extension. I typed up an example here #264 . The initializer should be implemented like that regardless, I didn't think of it in the initial PR.

davidapgar:
Closing in favor of https://github.com/square/workflow/pull/264

PR #264, opened 2019-04-11
Don't allow `AnyScreen` to wrap itself

timdonnelly:


timdonnelly:
Or it's at least an improvement that fixes an oversight – `asAnyScreen()` is a potential future ergonomic enhancement

PR #272, opened 2019-04-11
Updated compose in template to return an explicit Screen

### LINKED ISSUE #270: [swift] Update templates to have compose return a typed screen, instead of `Screen`
The templates currently have compose being defined as:
`func compose(state: Workflow.State, context: WorkflowContext<Workflow>) -> Screen`

This is no longer valid, as we must return an explicitly typed screen as of:
https://github.com/square/workflow/pull/200

Likely will use the product name passed into the template as the prefix on the Screen (which won't compile as it won't exist, but points someone in the right direction)

davidapgar:
A compose method can no longer return `Screen`, but must return a typed screen.
Updates the template to reflect this.

This means that the result of using the template will *not* compile until the screen returned from compose is one that is an actual type. However, I think it's a faster loop to finding this out than to have it be Screen and then be surprised by compiler errors later on.

Resolves #270

### DIFF
--- a/swift/Tooling/Templates/Workflow (Verbose).xctemplate/___FILEBASENAME___Workflow.swift
+++ b/swift/Tooling/Templates/Workflow (Verbose).xctemplate/___FILEBASENAME___Workflow.swift
@@ -79,8 +79,8 @@ extension ___VARIABLE_productName___Workflow {
 
 extension ___VARIABLE_productName___Workflow {
 
-    func compose(state: ___VARIABLE_productName___Workflow.State, context: WorkflowContext<___VARIABLE_productName___Workflow>) -> Screen {
-        #warning("Don't forget your compose implementation!")
-        fatalError()
+    func compose(state: ___VARIABLE_productName___Workflow.State, context: WorkflowContext<___VARIABLE_productName___Workflow>) -> String {
+        #warning("Don't forget your compose implementation and to return the correct rendering type!")
+        return "This is likely not the rendering that you want to return"
     }
 }


swift/Tooling/Templates/Workflow (Verbose).xctemplate/___FILEBASENAME___Workflow.swift
```diff
@@ -79,8 +79,8 @@ extension ___VARIABLE_productName___Workflow {
 
 extension ___VARIABLE_productName___Workflow {
 
-    func compose(state: ___VARIABLE_productName___Workflow.State, context: WorkflowContext<___VARIABLE_productName___Workflow>) -> Screen {
-        #warning("Don't forget your compose implementation!")
+    func compose(state: ___VARIABLE_productName___Workflow.State, context: WorkflowContext<___VARIABLE_productName___Workflow>) -> ___VARIABLE_productName___Screen {
```
bencochran: This makes it our our template can’t compile out of the box. Which to me is a bummer because I often find myself “filling out” the template from top to bottom. Could we use a type that actually exists?

davidapgar: I agree that it's a bummer. On the first use of a template, there won't be any type that exists that we can use (aside from AnyScreen, but absolutely don't want it to be that).

My other thought was to define a type with a comment as a placeholder. I *don't* want to encourage definition of the screen type to be done in the workflow.

Maybe something like:
```
// Delete this type and replace the return type of `compose` with the actual screen this workflow will output
struct PlaceholderScreen {}

func compose(state: ___VARIABLE_productName___Workflow.State, context: WorkflowContext<___VARIABLE_productName___Workflow>) -> PlaceholderScreen {
```

What do you think?

davidapgar: Updated this after talking with @bencochran, there are lots of types that we could return from compose.

Having it default to a String, so that it compiles immediately when the template is created.

PR #274, opened 2019-04-12
Rename WorkflowContext.compose to composeChild.

### LINKED ISSUE #267: Rename WorkflowContext.compose/render method to composeChild/renderChild
It is confusing that both the method you implement (but never call!) and the context method that you call to trigger your children's method are called the same thing. It is difficult to distinguish in the docs, especially since we need to clearly communicate that the former should never be called directly.

To address this, we propose adding a `Child` suffix to the latter method.

 - issue for `compose` vs `render` naming: #265

zach-klippenstein:
Closes #267.


rjrjr:
I take it you're going to follow up and rename both methods to `render` / `renderChild`?

zach-klippenstein:
> I take it you're going to follow up and rename both methods to `render` / `renderChild`?

Yes

PR #275, opened 2019-04-12
Upgrade Gradle to latest version.

### LINKED ISSUE #113: Upgrade to Gradle 5.x
AC: Summary.

Blocked on #115, since AGP 3.2 doesn't work with Gradle 5.
Blocks #112 since the Gradle Kotlin DSL is only stable in gradle 5.

zach-klippenstein:
Closes #113.


zach-klippenstein:
IntelliJ is failing to sync, I think it's because we're on too old an AGP version. Doing that first…

PR #277, opened 2019-04-12
Upgrade kotlin and coroutines dependencies to latest versions.

### LINKED ISSUE #188: Stop using parameterized coroutine cancellation that was deprecated in kotlin 1.3
There are a few places in the code where we forward a caught exception to the cancellation of another context via `cancel(e)`. These parameterized cancellation methods have been deprecated and replaced by parameter_less_ `cancel()`, so we should stop using them. I'm worried about just losing the exceptions if we don't forward them though so some non-trivial refactoring may be necessary.

zach-klippenstein:
Closes #188 because the deprecated `cancel` method that took `Throwable` has been removed.

The issue where workflow subscriptions were disposed multiple times also seems to have been fixed.

The only real thing this "broke" was how legacy workflow's `switchMapState` handles channel cancellation,
now we have to be a bit more explicit. This is why channels shouldn't be used as streams…


kotlin/workflow-runtime/src/main/java/com/squareup/workflow/WorkflowHost.kt
```diff
@@ -149,9 +150,7 @@ internal fun <I : Any, O : Any, R : Any> WorkflowNode<I, *, O, R>.start(
   } catch (e: Throwable) {
     // For some reason the exception gets masked if we don't explicitly pass it to cancel the
     // producer coroutine ourselves here.
-    // TODO https://github.com/square/workflow/issues/188 Stop using parameterized cancel.
-    @Suppress("DEPRECATION")
-    coroutineContext.cancel(e)
+    coroutineContext.cancel(if (e is CancellationException) e else CancellationException(null, e))
```
rjrjr: Sure this is still necessary? Wondering if the situation described in your comment was fixed.

zach-klippenstein: This `cancel` method only takes `CancellationException`s now, so just trying to not create more exceptions than necessary.

kotlin/legacy/legacy-workflow-rx2/src/main/java/com/squareup/workflow/legacy/rx2/EventChannel.kt
```diff
@@ -123,7 +123,7 @@ fun <E : Any> ReceiveChannel<E>.asEventChannel() = object : EventChannel<E> {
             .also { selectionJob.cancel() }
       } catch (cancellation: CancellationException) {
         val cause = cancellation.cause
-        if (cause == null) {
+        if (cause == null || cause is CancellationException) {
```
zach-klippenstein: Normal cancellations now have a self-referential cause, so we can't just check for null.

PR #279, opened 2019-04-12
Update changelogs for 0.13.0

davidapgar:


swift/CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 Change Log
 ==========
 
+## Version 0.13.0
+
+_2019-4-12_
+
+ * Don't allow `AnyScreen` to wrap itself.
```
bencochran: I think it’d be good to link over to the PR or issue associated with each thing in the changelog (because, for example, I wouldn’t fully know what this bullet point meant on its own)

davidapgar: Good idea. As we move to using issues, we should definitely always do this. Added the PRs for this and the previous changelog as well, to keep the history.

PR #281, opened 2019-04-13
Add support for listening to Flow streams.

### LINKED ISSUE #280: WorkflowContext should provide extensions to subscribe to Flows
`Flow` is the brand-new, preview/experimental abstraction for cold streams built-in to the coroutines standard library (Kotlin/kotlinx.coroutines#254, not to be confused with square/flow, the Android backstack library). It is compatible with the [Reactive Streams](http://www.reactive-streams.org/) spec, and similar to libraries like RxJava or Reactor, but written in pure Kotlin using coroutines. Most notably, this means it can be used in common modules in multi-platform projects, uses declaration-site variance, and propagates nullibility across operators. And since it's included in the coroutines library, we can support it directly from the `workflow-core` module without adding any dependencies.

While this library is still extremely unstable, and shouldn't be used in production, we can still build basic support for it as long as we keep the proper experimental annotations to discourage use in production (`@FlowPreview`).

zach-klippenstein:
Closes #280.

I'm not actually sure this is something we should merge. I'm worried it might make people feel more comfortable using the Flow API than they should be, even though our integration is also tagged with experimental annotations. However, having it in the code base lets us track API and behavior changes and ensure our integration feels natural as the API stabilizes.

### DIFF
--- a/kotlin/workflow-core/src/main/java/com/squareup/workflow/VeryExperimentalWorkflow.kt
+++ b/kotlin/workflow-core/src/main/java/com/squareup/workflow/VeryExperimentalWorkflow.kt
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2019 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.workflow
+
+import kotlin.Experimental.Level.ERROR
+import kotlin.annotation.AnnotationRetention.BINARY
+
+/**
+ * Marks Workflow APIs that are extremely likely to change in future versions, rely themselves on
+ * other unstable, experimental APIs, and SHOULD NOT be used in production code. Proceed with
+ * caution, and be ready to have the rug pulled out from under you.
+ */
+@MustBeDocumented
+@Retention(value = BINARY)
+@Experimental(level = ERROR)
+annotation class VeryExperimentalWorkflow
--- a/kotlin/workflow-testing/src/test/java/com/squareup/workflow/FlowWorkersTest.kt
+++ b/kotlin/workflow-testing/src/test/java/com/squareup/workflow/FlowWorkersTest.kt
@@ -1,0 +1,122 @@
+/*
+ * Copyright 2019 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+@file:Suppress("EXPERIMENTAL_API_USAGE")
+
+package com.squareup.workflow
+
+import com.squareup.workflow.testing.test
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.channels.consumeEach
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.flow.flow
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+class FlowWorkersTest {
+
+  private class ExpectedException : RuntimeException()
+
+  private val subject = Channel<String>(capacity = 1)
+  private var source = flow { subject.consumeEach { emit(it) } }
+
+  @UseExperimental(VeryExperimentalWorkflow::class)
+  private val worker by lazy { source.asWorker() }
+
+  @Test fun `flow emits`() {
+    worker.test {
+      subject.send("foo")
+      assertEquals("foo", nextOutput())
+
+      subject.send("bar")
+      assertEquals("bar", nextOutput())
+    }
+  }
+
+  @Test fun `flow finishes`() {
+    worker.test {
+      subject.close()
+      assertFinished()
+    }
+  }
+
+  @Test fun `flow finishes after emitting interleaved`() {
+    worker.test {
+      subject.send("foo")
+      assertEquals("foo", nextOutput())
+
+      subject.close()
+      assertFinished()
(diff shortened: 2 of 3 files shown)


kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowContext.kt
```diff
@@ -175,6 +179,59 @@ inline fun <StateT : Any, OutputT : Any, reified T> WorkflowContext<StateT, Outp
...
+  flow: Flow<T>,
+  key: String = "",
+  noinline handler: (ChannelUpdate<T>) -> WorkflowAction<StateT, OutputT>
+) = onNext(flow, KTypes.fromGenericType(Flow::class, T::class), key, handler)
+
+/**
+ * Subscribes to [flow] and invokes [handler] when it emits the next value.
+ *
+ * This allows us to use this method to, for example, monitor a stream of values emitted by a single
+ * fow over a series of state transitions. (Use [key] if you need to work simultaneously with
```
rjrjr: fow

kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowContext.kt
```diff
@@ -175,6 +179,59 @@ inline fun <StateT : Any, OutputT : Any, reified T> WorkflowContext<StateT, Outp
...
+ * several [Flow]s of the same type.)
+ * The subscription will be disposed when `compose` returns without passing a [Flow] of the same
```
rjrjr: `compose` is already a stale name. I'd change it to a link s.t. it gets updated as renames continue.

kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowContext.kt
```diff
@@ -175,6 +179,59 @@ inline fun <StateT : Any, OutputT : Any, reified T> WorkflowContext<StateT, Outp
...
+ * type and with the same key to this method.
+ *
+ * ## Experimental Status
+ *
+ * The [Flow] API is still marked experimental, and the compiler will warn you if you use it without
```
rjrjr: I think I feel okay merging it so long as this warning stays in tact.

Is there any practical way we can move this method out to be an extension from a new `workflow-flow` module though?

kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowContext.kt
zach-klippenstein: There's two compose methods though, and it would be weird to specifically link to one when they are the same concept with different shapes.

kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowContext.kt
zach-klippenstein: We can totally do that. Doing so makes sense for safety (you need to opt in to even see the support), but long-term I definitely think it should live in this module, since `Flow` will be as, if not more, fundamental a currency for streams than channels. And I wasn't sure it was worth the hassle of creating separate modules just for safety when we already have the annotations.

How strongly do you feel about making a separate module?

rjrjr: I think we should do the module. The tie breaker for me is forcing people to take in a transitive dependency on flow in order to use our core stuff, and the risk of its inevitably incompatible releases screwing up our own release cadence and backward compatibility. 

I'd really like to minimize how much baggage is required to adopt our work.

zach-klippenstein: If there were a transitive dependency on some additional library, I would agree with you. Flow is included in the core, common coroutines module though, just like channels.

rjrjr: Oh, okay then. That was my only valid argument.

zach-klippenstein:
TODO: Rebase/retarget this onto `kotlin-1.3.30` once #296 has merged.

kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowContext.kt
```diff
@@ -177,6 +181,59 @@ inline fun <StateT : Any, OutputT : Any, reified T> WorkflowContext<StateT, Outp
     handler
 )
 
+/**
+ * Subscribes to [flow] and invokes [handler] when it emits the next value.
+ *
+ * This allows us to use this method to, for example, monitor a stream of values emitted by a single
+ * fow over a series of state transitions. (Use [key] if you need to work simultaneously with
```
vRallev: Typo "fow"

kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowContext.kt
```diff
@@ -177,6 +181,59 @@ inline fun <StateT : Any, OutputT : Any, reified T> WorkflowContext<StateT, Outp
...
+ * several [Flow]s of the same type.)
+ * The subscription will be disposed when `compose` returns without passing a [Flow] of the same
+ * type and with the same key to this method.
+ *
+ * ## Experimental Status
+ *
+ * The [Flow] API is still marked experimental, and the compiler will warn you if you use it without
```
vRallev: Warn or error? Is there a way to make it an error without opting in? I know that we can make all warnings errors, but this isn't feasible in POS.

vRallev: Just read the previous conversation with Ray. It also sounds safer to me to have a separate library given that most people ignore warnings. If Flow would be in a further development stage I wouldn't mind, but since it's the first alpha release it might be dangerous.

kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowContext.kt
zach-klippenstein: I thought i fixed that…

zach-klippenstein:
Just merged #346 so I can rebase this now.

@vrallev @rjrjr How about this: afaik we can't make the FlowPreview annotation error, but we can make our own FlowWorkflowPreview annotation that does, and use that on these APIs as well.

vRallev:
SGTM

zach-klippenstein:
Added an erroring annotation, and finished updating all the unit tests. This is ready for another review.

kotlin/workflow-core/src/main/java/com/squareup/workflow/VeryExperimentalWorkflow.kt
```diff
@@ -0,0 +1,14 @@
+package com.squareup.workflow
```
rjrjr: copyright

kotlin/workflow-testing/src/test/java/com/squareup/workflow/FlowWorkersTest.kt
```diff
@@ -0,0 +1,122 @@
...
+  @UseExperimental(VeryExperimentalWorkflow::class)
+  private val worker by lazy { source.asWorker() }
+
+  @Test fun `flow emits`() {
+    worker.test {
+      subject.send("foo")
+      assertEquals("foo", nextOutput())
+
+      subject.send("bar")
+      assertEquals("bar", nextOutput())
```
kirillzh: Woohoo!

PR #282, opened 2019-04-13
Add the ability to update the input for a WorkflowHost.

### LINKED ISSUE #247: Add an input-update channel to WorkflowHost
AC: It is possible to send updated inputs to a WorkflowHost.

While it's possible to work around the lack of this method, doing so is cumbersome.

zach-klippenstein:
Closes #247.


kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/RealWorkflowHost.kt
```diff
@@ -0,0 +1,111 @@
...
+
+        val rendering = workflowNode.compose(workflow, input)
+        val snapshot = workflowNode.snapshot(workflow)
+
+        onUpdate(Update(rendering, snapshot, output))
+
+        // Tick _might_ return an output, but if it returns null, it means the state or a child
+        // probably changed, so we should re-compose/snapshot and emit again.
+        output = select {
+          // While the inputs channel is still open, select on it so we can detect new inputs.
```
zach-klippenstein: "Stop trying to read from the inputs channel after it's closed."

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/RealWorkflowHost.kt
```diff
@@ -0,0 +1,111 @@
...
+          if (!inputsClosed) {
+            @Suppress("EXPERIMENTAL_API_USAGE")
+            inputs.onReceiveOrNull { newInput ->
+              if (newInput == null) {
+                inputsClosed = true
+              } else {
+                input = newInput
+              }
+              // No output.
```
zach-klippenstein: "No output. Returning from the select will go to the top of the loop to do another compose pass."

zach-klippenstein:
Would it be more useful, instead of requiring an initial input immediately and having an update function, to just only accept a stream of inputs? The workflow would then just not start until the first input was emitted.

rjrjr:
> Would it be more useful, instead of requiring an initial input immediately and having an update function, to just only accept a stream of inputs? The workflow would then just not start until the first input was emitted.

Yes! We've been longing for that in production already.

zach-klippenstein:
Totes. I'll update the PR asap. As for API shape, which do you prefer:

zach-klippenstein:
@rjrjr Updated the API. `WorkflowViewModel` now uses the new `flatMapWorkflow` operator in `workflow-rx2-runtime` to convert a stream of inputs into a stream of outputs. Should `inputs` be an `Observable` conveted to a `Flowable` with `BackpressureStrategy.BUFFER` internally instead?

zach-klippenstein:
Also, maybe the stream operator is too much - could just make an extension function on `Factory` that takes an input stream and just returns a `WorkflowHost`. I just like the operator because the API surface is so tiny and it hides all the channel machinery completely.

kotlin/workflow-rx2-runtime/build.gradle
```diff
@@ -0,0 +1,38 @@
...
+
+sourceCompatibility = JavaVersion.VERSION_1_7
+targetCompatibility = JavaVersion.VERSION_1_7
+
+dokka rootProject.ext.defaultDokkaConfig
+
+dependencies {
+  compileOnly deps.annotations.intellij
+
+  api project(':workflow-runtime')
```
zach-klippenstein: Should be implementation

kotlin/workflow-rx2-runtime/src/main/java/com/squareup/workflow/rx2/FlatMapWorkflow.kt
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Use [Flowable.flatMapWorkflow] instead. This function is just defined to be discoverable when
+ * working with [Observable]s instead of [Flowable]s.
+ *
+ * See [Observable.toFlowable] and [Flowable.toObservable] to convert between stream types.
+ */
+@Suppress("DeprecatedCallableAddReplaceWith", "unused")
+@Deprecated("Use Flowable.flatMapWorkflow instead.", level = ERROR)
+fun Observable<*>.flatMapWorkflow(): Nothing =
+  throw UnsupportedOperationException("Use Flowable.flatMapWorkflow instead.")
```
zach-klippenstein: This is dumb, just make it take an additional `BackPressure` argument and do the conversions itself.

kotlin/workflow-rx2-runtime/src/main/java/com/squareup/workflow/rx2/FlatMapWorkflow.kt
```diff
@@ -0,0 +1,83 @@
...
+ * This function operates on [Flowable] instead of [Observable] because the workflow runtime
+ * inherently supports backpressure. [Flowable] supports backpressure, [Observable] does not.
+ * RxJava provides operators to adapt between [Flowable]s and [Observable]s by explicitly specifying
+ * how to use backpressure. By operating on [Flowable], this operator leaves it up to the caller to
+ * specify strategies for handling backpressure, instead of assuming any particular behavior.
+ */
+fun <InputT : Any, OutputT : Any, RenderingT : Any> Flowable<InputT>.flatMapWorkflow(
+  workflow: Workflow<InputT, OutputT, RenderingT>,
+  initialSnapshot: Snapshot? = null
+): Flowable<Update<OutputT, RenderingT>> = GlobalScope.rxFlowable(
```
zach-klippenstein: Actually I think this might need to take a dispatcher, although it can default to Unconfined.

kotlin/workflow-rx2-runtime/src/main/java/com/squareup/workflow/rx2/FlatMapWorkflow.kt
rjrjr: Actually, I'd just delete this one entirely. No one is going to use this layer except for us. Extra guardrails and extra conveniences here will just encourage people to integrate at the wrong layer, and be that much more for us to maintain.

zach-klippenstein: Sure.

kotlin/workflow-rx2-runtime/src/main/java/com/squareup/workflow/rx2/FlatMapWorkflow.kt
```diff
@@ -0,0 +1,72 @@
...
+ * RxJava provides operators to adapt between [Flowable]s and [Observable]s by explicitly specifying
+ * how to use backpressure. By operating on [Flowable], this operator leaves it up to the caller to
+ * specify strategies for handling backpressure, instead of assuming any particular behavior.
+ */
+fun <InputT : Any, OutputT : Any, RenderingT : Any> Flowable<InputT>.flatMapWorkflow(
+  workflow: Workflow<InputT, OutputT, RenderingT>,
+  initialSnapshot: Snapshot? = null,
+  dispatcher: CoroutineDispatcher = Dispatchers.Unconfined
+): Flowable<Update<OutputT, RenderingT>> =
+// We're ok not having a job here because the lifetime of the coroutine will be controlled by the
```
zach-klippenstein: This formatting is the auto-formatter being stupid.

zach-klippenstein:
A bunch of tests broke in interesting ways because things are more asynchronous now. Added `WorkflowTester.awaitFailure` and `.withFailure` to make testing errors more reliable.

PR #283, opened 2019-04-13
WIP: Write a Flow operator that runs a Workflow.

### LINKED ISSUE #263: Refactor WorkflowHost as a stream transformation
I am thinking about ultimately defining `WorkflowHost` as a stream transformation. E.g., using the new preview cold streams library (Kotlin/kotlinx.coroutines#254): 

```kotlin
val inputs: Flow<WfInput>
val outputsAndRenderings: Flow<Pair<WfRendering, WfOutput?>> = inputs.flatMapWorkflow(myWorkflow)
```

It already has a stream of output events, but since we don't have a native kotlin (not to be confused with kotlin native) streams library yet, the host has to have its own lifetime and vend a hot channel. I'd like to turn that into a cold stream that only fires up the workflow universe when subscribed to, and tears it down when that subscription is disposed.

Issue #247 is a proposal for a stream of inputs. With that, `WorkflowHost` will have a transformer shape - stream in, stream out. So we can think of a workflow as a fancy way to write reactive operators declaratively.

_Tagging this for post-v1.0 because Flow is still preview, so to do this properly we'd currently need to give the runtime direct dependencies on third-party Rx libraries which would be messy._

zach-klippenstein:
Closes #263.

### DIFF
--- a/kotlin/workflow-runtime/src/main/java/com/squareup/workflow/FlatMapWorkflow.kt
+++ b/kotlin/workflow-runtime/src/main/java/com/squareup/workflow/FlatMapWorkflow.kt
@@ -1,0 +1,60 @@
+/*
+ * Copyright 2019 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


zach-klippenstein:
TODO: Rebase/retarget this onto `kotlin-1.3.30` once #296 has merged.

zach-klippenstein:
Declining this, going to keep WorkflowHost but eventually use Flow for the API.

PR #286, opened 2019-04-16
Pass the workflow's scope into initialState.

### LINKED ISSUE #288: Pass the workflow's coroutine scope into StatefulWorkflow.initialState.
Occasionally `Workflow`s need to know about their `CoroutineScope`. The scope is passed to the channel producer lambda in `onReceive` to allow new coroutines to be started to do asynchronous work. To properly support `Worker`s (see #287), `onReceive` no longer gets a lambda, and we need to be able to access the `CoroutineScope` from `initialState`.

And it turns out we can kill two birds with one stone here. Stateful workflows occasionally know when they are started – `initialState` is always called. Occasionally, they also need to know when they're stopped (see #231). Our first attempt to do this was just to add an `onTeardown` hook to `WorkflowContext`. This works, but is very limited in what it can do. If `initialState` gets the `CoroutineScope`, we can access the `Job` in the context, or launch whole coroutines, to track the lifetime of the workflow. We can remove the `onTeardown` hook and put that code in a completion handler instead.

zach-klippenstein:
Closes #288.

### DIFF
--- a/kotlin/workflow-core/src/main/java/com/squareup/workflow/StatefulWorkflow.kt
+++ b/kotlin/workflow-core/src/main/java/com/squareup/workflow/StatefulWorkflow.kt
@@ -15,6 +15,8 @@
  */
 package com.squareup.workflow
 
+import kotlinx.coroutines.CoroutineScope
+
 /**
  * A composable, stateful object that can [handle events][WorkflowContext.onEvent],
  * [delegate to children][WorkflowContext.composeChild], [subscribe][onReceive] to arbitrary streams from
@@ -71,10 +73,15 @@ abstract class StatefulWorkflow<
    * If the workflow is being restored from a [Snapshot], [snapshot] will be the last value
    * returned from [snapshotState], and implementations that return something other than
    * [Snapshot.EMPTY] should create their initial state by parsing their snapshot.
+   * @param scope
+   * The [CoroutineScope] in which this workflow lives. The scope will be cancelled when the
+   * workflow is being torn down, so this scope can be used to start coroutines to track the
+   * lifetime of the workflow "session".
    */
   abstract fun initialState(
     input: InputT,
-    snapshot: Snapshot?
+    snapshot: Snapshot?,
+    scope: CoroutineScope
   ): StateT
 
   /**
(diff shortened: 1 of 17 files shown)


kotlin/workflow-core/src/main/java/com/squareup/workflow/StatefulWorkflow.kt
```diff
@@ -71,10 +73,15 @@ abstract class StatefulWorkflow<
...
+   * @param scope
+   * The [CoroutineScope] in which this workflow lives. The scope will be cancelled when the
+   * workflow is being torn down, so this scope can be used to start coroutines to track the
+   * lifetime of the workflow "session".
    */
   abstract fun initialState(
     input: InputT,
-    snapshot: Snapshot?
+    snapshot: Snapshot?,
+    scope: CoroutineScope
```
rjrjr: @davidapgar @bencochran  @timdonnelly Seems like we should be able to come up with a similar object to pass into the analogous method in Swift — basically something that can accept tearDown duties instead of the method on WorkflowContext.

kotlin/workflow-core/src/main/java/com/squareup/workflow/StatefulWorkflow.kt
```diff
@@ -71,10 +73,15 @@ abstract class StatefulWorkflow<
    * If the workflow is being restored from a [Snapshot], [snapshot] will be the last value
    * returned from [snapshotState], and implementations that return something other than
    * [Snapshot.EMPTY] should create their initial state by parsing their snapshot.
+   * @param scope
+   * The [CoroutineScope] in which this workflow lives. The scope will be cancelled when the
```
rjrjr: Can we provide an example of how to use it?

zach-klippenstein: I'm not sure what kind of example would be useful. It's basically only there to support `Worker`s (which users don't need to care about unless they want to look at source directly), or cancellation of some external resource, and I don't even know yet what the best way will be to use this scope to cancel those. I'll file an issue to track improving this doc as we go forward: #290.

kotlin/workflow-core/src/main/java/com/squareup/workflow/StatefulWorkflow.kt
zach-klippenstein: I'm not suggesting this is a good direction, but if Swift wanted to match this as accurately as possible, it would be a teardown hook plus a scheduler.

PR #289, opened 2019-04-16
Implement Workers.

### LINKED ISSUE #287: Simplify subscriptions: Only support hot channels, support cold subs through workers
The kotlin api currently supports subscribing to "channel producers", which can represent hot or cold streams. It also automatically handles the bookkeeping to keep track of them across compose passes and cancel channels when necessary.

This is the exact same work as we do to track child workflows. In Swift land, they don't bother with any of this. Their `WorkflowContext` only lets you track hot streams (`Signal`s). To track cold observables, they have a `Worker` type that subscribes to a cold stream and uses the hot stream machinery to subscribe to it (more or less). We can do the same thing: `onReceive` should accept a `ReceiveChannel`, not a `() -> ReceiveChannel`, and it should not track when we start/stop caring about a channel to cancel it. The contract becomes: "you are responsible for creating the channel, so you are responsible for cancelling it."

cc @bencochran @timdonnelly @rjrjr

zach-klippenstein:
The need for workers is discussed in #287. This commit also replaces `RenderContext.on*` methods with `Workers`.

For example, if you have a retrofit service that doesn't use any futures, you can create a worker like this:
```kotlin
interface TimeService {
  @GET("/time")
  fun getTime(timezone: String): Long
}

fun TimeService.getTimeWorker(timezone: String): Worker<Long> = TimeWorker(timezone, this)

// We can make a Retrofit CallAdapter to create these classes automatically.
private class TimeWorker(
  private val timezone: String,
  private val service: TimeService
): Worker<Long> {
  override suspend fun performWork(emitter: Emitter<T>) {
    // This is a blocking service call, do it on a background thread.
    withContext(Dispatchers.IO) {
      emitter.emit(service.getTime(timezone))
    }
  }
  
  override fun doesSameWorkAs(otherWorker: Worker<*>): Boolean =
    otherWorker is TimeWorker && timezone == otherWorker.timezone
}
```

Alternatively, if the output type is more specific, you can save some typing:
```kotlin
interface TimeService {
  @GET("/time")
  fun getTime(timezone: String): TimeResponse
}

// This worker is now automatically distinguished by its output type.
fun TimeService.getTimeWorker(timezone: String): Worker<TimeResponse> = Worker.from {
  withContext(Dispatchers.IO) {
    getTime(timezone)
  }
}
```

Then you can use it from a workflow like:
```kotlin
class MyWorkflow @Inject constructor(private val timeService: TimeService) {

  override fun render(…) {
    context.onWorkerResult(timeService.getTimeWorker("PST")) { millis -> emitOutput(Date(millis)) }
  }
}
```

Closes #287.


zach-klippenstein:
I've updated this PR to include the updates from #287. I haven't updated any unit tests now, but it's ready for API review.

cc @rjrjr @vRallev @davidapgar @bencochran @timdonnelly

kotlin/workflow-core/src/main/java/com/squareup/workflow/StartStopWorker.kt
```diff
@@ -0,0 +1,46 @@
...
+
+import com.squareup.workflow.Worker.Emitter
+import kotlinx.coroutines.NonCancellable
+import kotlinx.coroutines.suspendCancellableCoroutine
+import kotlinx.coroutines.withContext
+
+/**
+ * [Worker] that performs some action when started and when stopped.
+ */
+abstract class StartStopWorker : Worker<Nothing> {
```
zach-klippenstein: @rjrjr @davidapgar This replaces the `onTeardown` hook we used to have.

kotlin/workflow-core/src/main/java/com/squareup/workflow/Worker.kt
```diff
@@ -0,0 +1,239 @@
...
+
+  /**
+   * Override this method to do the actual work. This is a suspend function and is invoked in the
+   * context of the workflow runtime. When this [Worker], it's parent [Workflow], or any ancestor
+   * [Workflow]s are torn down, the coroutine in which this function is invoked will be cancelled.
+   * If you need to do cleanup when the [Worker] is cancelled, wrap your work in a `try { }` block
+   * and do your cleanup in a `finally { }` block. For more information on how coroutine
+   * cancellation works, see [the coroutine guide](https://kotlinlang.org/docs/reference/coroutines/cancellation-and-timeouts.html).
+   */
+  suspend fun performWork(emitter: Emitter<T>)
```
zach-klippenstein: Should this be `suspend fun Emitter<T>.performWork()` instead?

zach-klippenstein:
One thing that might be surprising about this change is that it removes `onReceive` completely. In Swift, `RenderContext` still has a `subscribe(Signal)` function for subscribing to _hot_ streams. The rough Kotlin equivalent of that is `ReceiveChannel`, and we could provide a method that takes a channel (instead of a channel provider). However, we don't currently have any need for that. All the streams we're working with internally are RxJava, which don't differentiate between hot and cold with types, so we wouldn't use this method in practice and will just end up converting all our observables to workers anyway. If there's a use case for such a method in the future, we can add it then.

timdonnelly:
@zach-klippenstein I have wondered if we could drop the subscribe method as well. It’s very straightforward to use a hot signal via a worker, so at this point it’s really just a convenience thing.

zach-klippenstein:
Yea, I considered adding extensions to `RenderContext` to get back the same API we had before (`onNext`, `onSuccess`, etc), but the `Foo.asWorker()` functions are more composable so I think I'll just stick with those for now and see how it feels.

zach-klippenstein:
I also intentionally didn't give workers an input type, unlike workflows. When workers don't have an input type, a `Worker` instance represents a specific job. When they _do_ have an input type, an instance represents a _type_ of job. It's expected that service dependencies and the like would be injected into the worker instance. If they had an input type, in kotlin, it would naturally map to, eg, request data, so it seems a little weird that they don't. However, for the case of using `Worker` as a return type from a retrofit service, the input would always be `Unit` (since the request data is passed into the function that returns the worker). Workers created from stream/future types would also always have `Unit` input. It just seems unnecessarily verbose to give them an input type in kotlin.

@timdonnelly @bencochran @davidapgar In Swift, is the request data passed in with the rest of the props?

davidapgar:
> I also intentionally didn't give workers an input type, unlike workflows. When workers don't have an input type, a `Worker` instance represents a specific job. When they _do_ have an input type, an instance represents a _type_ of job. It's expected that service dependencies and the like would be injected into the worker instance. If they had an input type, in kotlin, it would naturally map to, eg, request data, so it seems a little weird that they don't. However, for the case of using `Worker` as a return type from a retrofit service, the input would always be `Unit` (since the request data is passed into the function that returns the worker). Workers created from stream/future types would also always have `Unit` input. It just seems unnecessarily verbose to give them an input type in kotlin.
> 
> @timdonnelly @bencochran @davidapgar In Swift, I don't _think_ this issue even comes up because you don't have separate input types. But do you have any opinions on "worker is a future/stream" vs "worker turns an input into an output"?

You're right that this likely isn't an issue in Swift, since we don't have a separate `Input` type (as the Workflow acts as the input, since it's a description of the workflow). I would lean towards a "worker is a future/stream" as the model for this - I would expect that all of our workers are "fully formed", and that it wouldn't provide additional value to have them take an input to turn into an output.

timdonnelly:
Right, I don't think we would gain anything (apart from complexity) by adding an `Input` type.

zach-klippenstein:
Just pushed a bunch of unit tests, testing helpers for testing `Worker`s, and updates to all the existing unit tests except the Rx2 integration ones.

kotlin/workflow-testing/src/test/java/com/squareup/workflow/WorkerTest.kt
```diff
@@ -0,0 +1,243 @@
...
+
+  @Test fun `create returns non-equivalent workers based on key`() {
+    val worker1 = Worker.create<Unit>("key1") {}
+    val worker2 = Worker.create<Unit>("key2") {}
+
+    assertFalse(worker1.doesSameWorkAs(worker2))
+  }
+
+  @Test fun `create emits and finishes`() {
+    val worker = Worker.create<String> {
```
zach-klippenstein: Fun fact: If we add the `@BuilderInference` annotation to this lambda parameter, the compiler can infer `String` from the `emitOutput` calls below and we can remove all these explicit type arguments. The IDE will even encourage us to do so.

It's _extremely_ tempting, but `@BuilderInference` and all its new type inference friends are still marked experimental so I'm resisting for now. That said, there's no explicit API here, and most real-world usages of workers in our codebase will use the Rx extensions, so the blast radius would probably be pretty small…

vRallev: Don't you need to use the experimental type inference for that? The last time I've tested it many things broke, especially the Rx calls.

zach-klippenstein: Yea.

zach-klippenstein: > Don't you need to use the experimental type inference for that?

Yes.

> The last time I've tested it many things broke, especially the Rx calls.

😢

zach-klippenstein:
Tests are all done, this is now completely ready for review!

kotlin/workflow-core/src/main/java/com/squareup/workflow/Worker.kt
```diff
@@ -0,0 +1,310 @@
...
+interface Worker<out T> {
+
+  /**
+   * Passed into [performWork] to allow the [Worker] to emit outputs.
+   */
+  interface Emitter<in T> {
+    /**
+     * Delivers an output value to the [Workflow] that is running this [Worker].
+     */
+    suspend fun emitOutput(output: T)
```
vRallev: Why does this need to be a suspend function?

kotlin/workflow-core/src/main/java/com/squareup/workflow/Worker.kt
vRallev: No, to me this clearer as is.

kotlin/workflow-core/src/main/java/com/squareup/workflow/Worker.kt
zach-klippenstein: Because this usually maps to a `SendChannel.send`, which is a suspending operation. This is how workers handle backpressure.

kotlin/samples/tictactoe/common/src/main/java/com/squareup/sample/authworkflow/AuthWorkflow.kt
```diff
@@ -79,8 +80,9 @@ class RealAuthWorkflow(private val authService: AuthService) : AuthWorkflow,
       }
 
       is Authorizing -> {
-        context.onSuccess(
+        context.onWorkerOutput(
```
rjrjr: This is so much better.

kotlin/workflow-core/src/main/java/com/squareup/workflow/LifecycleWorker.kt
```diff
@@ -0,0 +1,62 @@
...
+      withContext(NonCancellable) {
+        onCancelled()
+      }
+    }
+  }
+
+  /**
+   * Equates [LifecycleWorker]s that have the same concrete class.
+   */
+  override fun doesSameWorkAs(otherWorker: Worker<*>): Boolean =
```
rjrjr: This remains open, right? Which is good IMHO.

kotlin/workflow-core/src/main/java/com/squareup/workflow/RenderContext.kt
```diff
@@ -148,35 +122,6 @@ interface RenderContext<StateT : Any, in OutputT : Any> {
   fun onTeardown(handler: () -> Unit)
```
rjrjr: I thought this method would be deleted or at least deprecated.

kotlin/workflow-core/src/main/java/com/squareup/workflow/Worker.kt
```diff
@@ -0,0 +1,310 @@
...
+     * The worker will emit the value **if and only if the value is not null**, then finish.
+     *
+     * The returned [Worker] will equate to any other workers created with any of the [Worker]
+     * builder functions that have the same output type.
+     */
+    inline fun <reified T : Any> fromNullable(
+      key: String = "",
+        // This could be crossinline, but there's a coroutines bug that will cause the coroutine
+        // to immediately resume on suspension inside block when it is crossinline.
+        // TODO link to KT ticket.
```
rjrjr: Still need this link.

kotlin/workflow-core/src/main/java/com/squareup/workflow/Worker.kt
```diff
@@ -0,0 +1,310 @@
...
+        }
+      }
+    }
+  }
+}
+
+/**
+ * Returns a [Worker] that will await this [Deferred] and then emit it.
+ *
+ * Note that [Deferred] is a "hot" future type – calling a function that _returns_ a [Deferred]
```
rjrjr: Is this caveat a good reason to get rid of this extension entirely? Wondering if it's going to be easier to misuse than  not.

kotlin/workflow-core/src/main/java/com/squareup/workflow/StartStopWorker.kt
rjrjr: Except we still have it?

kotlin/workflow-core/src/main/java/com/squareup/workflow/RenderContext.kt
rjrjr: Ah, later PR kills it.

kotlin/workflow-core/src/main/java/com/squareup/workflow/Worker.kt
rjrjr: Seems similar to the concern you raise in #337

kotlin/workflow-core/src/main/java/com/squareup/workflow/LifecycleWorker.kt
zach-klippenstein: Yes.

kotlin/workflow-core/src/main/java/com/squareup/workflow/Worker.kt
zach-klippenstein: Added in later PR.

PR #293, opened 2019-04-16
[kotlin] Rename compose and composeChild to render and renderChild.

### LINKED ISSUE #265: [kotlin] Rename compose method and concept in docs to render
The `compose` function was originally called `render` in the Swift implementation. After much discussion, we decided to rename to `compose` to avoid suggesting too strong an association with UI (it can return screen data, but can also be any arbitrary value).

However, this is dumb for a few reasons:
 - we're still calling the _type_ `Rendering`
 - "render" is a much more specific verb to represent this specific concept than the super-generic "compose"
 - It turns out Swift never actually got around to renaming everything anyway.

For these reasons, I propose renaming the `State{ful|less}Workflow.compose` method back to `render`.

 - Swift issue: #266

zach-klippenstein:
Closes #265.


zach-klippenstein:
> You missed a couple of test method / test variable names, nothing important.

Already fixed.

PR #295, opened 2019-04-17
[RFC] Add SideEffect to Workflow

bencochran:
There are cases where side effect work does not have a lifetime or meaningful completion. The common example in my workflows has been logging. These are typically closely tied to `WorkflowAction` events (but also occasionally would like to happen when a workflow starts or updates).

Some ways of representing these in today’s `Workflow`s are:

1. Make an action hold some sort of `logger`
```swift
struct MyWorkflow: Workflow {

    var logger: Logger
    // …

    struct Action: WorkflowAction {

        enum Event {
            case tappedButton
            case tappedOtherButton
        }

        var event: Event
        var logger: Logger

        func apply(toState state: inout State) -> Output? {
            switch event {
            case .tappedButton:
                logger.log(.buttonTap)
                state.awesome = .yes
                return nil
            case .tappedOtherButton:
                logger.log(.otherButtonTap)
                state.awesome = .evenMoreSo
                return nil
            }
        }
    }
}
```

2. As a `Worker`
```swift
struct MyWorkflow: Workflow {

    var logger: Logger
    // …
 
    struct State {
        // …
        var pendingLogs: [Logger.Event]
    }

    struct Action: WorkflowAction {
        case tappedButton
        case tappedOtherButton

        func apply(toState state: inout State) -> Output? {
            switch self {
            case .tappedButton:
                state.pendingLogs.append(.tappedButton)
                state.awesome = .yes
                return nil
            case .tappedOtherButton:
                state.pendingLogs.append(.tappedOtherButton)
                state.awesome = .evenMoreSo
                return nil
            }
        }
    }

    struct LogWorker: Worker {

        typealias Output = Void

        var logger: Logger
        var event: Logger.Event

        func run() -> SignalProducer<Void, NoError> {
            return SignalProducer({
                logger.log(event: event)
            })
        }

    }

    struct LogCompleteAction: WorkflowAction {

        var event: Logger.Event

        func apply(toState state: inout State) -> Output? {
            guard let index = state.pendingLogs.firstIndex(of: event) else {
                return nil
            }
            state.pendingLogs.remove(at: index)
            return nil
        }

    }

    func compose(state: State, context: WorkflowContext<MyWorkflow>) -> Rendering {
        for event in state.pendingLogs {
            context.awaitResult(
               for: LogWorker(
                    logger: logger,
                    event: event),
               outputMap: { _ in
                    return LogCompleteAction(event: event)
                })
        }

        // …

        return MyScreen()
    }
}
```

3. Hand a `logger` to the screen and let UI do the logging.

---

Of these, 1 is my current preferred current approach. But, it makes otherwise-pure `Actions` have hidden side effects. And that’s not something I’d like to encourage. (It’s also a little cumbersome because code that used to read `sink.send(.signInTapped)` now reads `sink.send(Action(logger: logger, event: .signInTapped`). 2 to me is the most “correct” place for effects to live, but is _quite_ cumbersome and verbose. 3 is “easy”, but encourages passing pointers into the outside world into Screens and prevents us from verifying these effects in my workflow tests.

This PR represents a proposal to formalize `SideEffect` as part of the Workflow API. Translating example 1 would produce:

```swift

struct MyWorkflow: Workflow {

    var logger: Logger
    // …

    enum SideEffect {
        case log(Logger.Event)
    }

    func perform(sideEffect: SideEffect) {
        switch sideEffect {
        case .log(let event):
            logger.log(event)
        }
    }

    struct Action: WorkflowAction {

        case tappedButton
        case tappedOtherButton

        func apply(toState state: inout State, context: inout SideEffectContext<MyWorkflow>) -> Output? {
            switch event {
            case .tappedButton:
                context.perform(sideEffect: .log(.buttonTap))
                state.awesome = .yes
                return nil
            case .tappedOtherButton:
                context.perform(sideEffect: .log(.otherButtonTap))
                state.awesome = .evenMoreSo
                return nil
            }
        }
    }

}


```

### DIFF
--- a/swift/Tooling/Templates/Workflow (Verbose).xctemplate/___FILEBASENAME___Workflow.swift
+++ b/swift/Tooling/Templates/Workflow (Verbose).xctemplate/___FILEBASENAME___Workflow.swift
@@ -24,11 +24,11 @@ extension ___VARIABLE_productName___Workflow {
 
     }
 
-    func makeInitialState() -> ___VARIABLE_productName___Workflow.State {
+    func makeInitialState(context: inout SideEffectContext<___VARIABLE_productName___Workflow>) -> ___VARIABLE_productName___Workflow.State {
         return State()
     }
 
-    func workflowDidChange(from previousWorkflow: ___VARIABLE_productName___Workflow, state: inout State) {
+    func workflowDidChange(from previousWorkflow: ___VARIABLE_productName___Workflow, state: inout State, context: inout SideEffectContext<___VARIABLE_productName___Workflow>) {
 
     }
 }
@@ -42,7 +42,7 @@ extension ___VARIABLE_productName___Workflow {
 
         typealias WorkflowType = ___VARIABLE_productName___Workflow
 
-        func apply(toState state: inout ___VARIABLE_productName___Workflow.State) -> ___VARIABLE_productName___Workflow.Output? {
+        func apply(toState state: inout ___VARIABLE_productName___Workflow.State, context: inout SideEffectContext<___VARIABLE_productName___Workflow>) -> ___VARIABLE_productName___Workflow.Output? {
 
             switch self {
                 // Update state and produce an optional output based on which action was received.
--- a/swift/Workflow/Sources/Workflow.swift
+++ b/swift/Workflow/Sources/Workflow.swift
@@ -48,16 +48,37 @@ public protocol Workflow: AnyWorkflowConvertible {
     /// is commonly a view / screen model.
     associatedtype Rendering
 
+    /// `SideEffects` defines work performed in `makeInitialState`,
+    /// `workflowDidChange` or from a `WorkflowAction`.
+    ///
+    /// - Note:
+    /// Side effects provide no mechanism for information coming back in to a
+    /// Workflow. For work that should affect this workflow’s state, see
+    /// `Worker`s.
+    associatedtype SideEffect = Never
+
     /// This method is invoked once when a workflow node comes into existence.
     ///
+    /// - Parameter context: The side effect context collects any requested side
+    ///                      effects to be performed when this workflow node
+    ///                      comes into existence.
     /// - Returns: The initial state for the workflow.
-    func makeInitialState() -> State
+    func makeInitialState(context: inout SideEffectContext<Self>) -> State
 
     /// Called when a new workflow is passed down from the parent to an existing workflow node.
     ///
     /// - Parameter previousWorkflow: The workflow before the update.
     /// - Parameter state: The current state.
-    func workflowDidChange(from previousWorkflow: Self, state: inout State)
+    /// - Parameter context: The side effect context collects any requested side
+    ///                      effects to be performed as a result of this
+    ///                      workflow change.
+    func workflowDidChange(from previousWorkflow: Self, state: inout State, context: inout SideEffectContext<Self>)
+
+    /// Called when this workflow requests a side effect. This is the moment to
+    /// do the work described by the given `SideEffect`.
+    ///
+    /// - Parameter sideEffect: The side effect to be performed.
+    func perform(sideEffect: SideEffect)
 
     /// Called to "compose" the current state into `Rendering`. A workflow's `Rendering` type is commonly a view or
     /// screen model.
@@ -68,6 +89,15 @@ public protocol Workflow: AnyWorkflowConvertible {
     ///                      then used to invoke `context.render(_ workflow:)`, which returns the child's `Rendering`
     ///                      type after creating or updating the nested workflow.
     func compose(state: State, context: WorkflowContext<Self>) -> Rendering
+
+}
+
+extension Workflow where SideEffect == Never {
+
+    /// When `SideEffect` is `Never`, there is no valid implementation of this
+    /// method (it can never be called).
+    public func perform(sideEffect: Never) {
+    }
 
 }
 
--- a/swift/Workflow/Sources/WorkflowNode.swift
+++ b/swift/Workflow/Sources/WorkflowNode.swift
@@ -19,7 +19,14 @@ final class WorkflowNode<WorkflowType: Workflow> {
         
         /// Get the initial state
         self.workflow = workflow
-        self.state = workflow.makeInitialState()
+
+        var context = SideEffectContext<WorkflowType>()
+        self.state = workflow.makeInitialState(context: &context)
+
+        /// Perform side effects
+        for sideEffect in context.sideEffects {
+            workflow.perform(sideEffect: sideEffect)
+        }
 
         subtreeManager.onUpdate = { [weak self] output in
             self?.handle(subtreeOutput: output)
@@ -34,8 +41,15 @@ final class WorkflowNode<WorkflowType: Workflow> {
 
         switch subtreeOutput {
         case .update(let event, let source):
+            var context = SideEffectContext<WorkflowType>()
+
             /// Apply the update to the current state
-            let outputEvent = event.apply(toState: &state)
+            let outputEvent = event.apply(toState: &state, context: &context)
+
+            /// Perform side effects
+            for sideEffect in context.sideEffects {
+                workflow.perform(sideEffect: sideEffect)
+            }
 
             /// Finally, we tell the outside world that our state has changed (including an output event if it exists).
             output = Output(
@@ -66,8 +80,17 @@ final class WorkflowNode<WorkflowType: Workflow> {
 
     /// Updates the workflow.
     func update(workflow: WorkflowType) {
-        workflow.workflowDidChange(from: self.workflow, state: &state)
+        let previousWorkflow = self.workflow
         self.workflow = workflow
+
+        var context = SideEffectContext<WorkflowType>()
+
+        self.workflow.workflowDidChange(from: previousWorkflow, state: &state, context: &context)
+
+        /// Perform side effects
+        for sideEffect in context.sideEffects {
+            self.workflow.perform(sideEffect: sideEffect)
+        }
     }
 
     func makeDebugSnapshot() -> WorkflowHierarchyDebugSnapshot {
--- a/swift/Workflow/Sources/SideEffectContext.swift
+++ b/swift/Workflow/Sources/SideEffectContext.swift
@@ -1,0 +1,23 @@
+/// `SideEffectContext` collects side effects from a Workflow starting,
+/// changing, or a `WorkflowAction` being applied.
+///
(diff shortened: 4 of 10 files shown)


swift/Workflow/Sources/Workflow.swift
```diff
@@ -71,6 +92,15 @@ public protocol Workflow: AnyWorkflowConvertible {
 
 }
 
+extension Workflow where SideEffect == Never {
+
+    /// When `SideEffect` is `Never`, there is no valid implementation of this
+    /// method (it can never be called).
+    public func perform(sideEffect: Never) {
```
zach-klippenstein: I'm so jealous of this.

davidapgar:
I'm pretty strongly in support of this, as it provides a way to keep the apply methods as "pure" functions (a transformation of a state from a previous to new given an action), with an explicit modeling of what side effects *should* be performed.

The mocking/testing use case is a big plus for us. However, it's more of reducing the boilerplate that we need to do when want to have side effects happen as a result of an Action. The current way of doing this (like what was described in (1)) have felt more cumbersome than beneficial.

zach-klippenstein:
> it's more of reducing the boilerplate that we need to do when want to have side effects happen as a result of an Action.

Yea, that seems like a bigger win.

rjrjr:
For a Kotlin flavor of this that doesn't drown us in boilerplate, could we use an orthogonal interface, something like `HasSideEffects<EffectT>`?

```
interface Operator<OperationT> {
  fun perform(operation: OperationT)
}

<OperationT> fun WorkflowContext.performOperation(
  operator: Operator<OperationT>,
  operation: OperationT
)

class MyWorkflow : StatefulWorkflow<I, S, O, R>, Operator<MyOps> {
  fun render(
      input: I,
      state: S,
      context: WorkflowContext<...>
  ) : R = when(state) {
    is OppyState -> {
      context.perform(this, MyOps.BenWasHere)
    }
   // ...
  }
}

**Comment by zach-klippenstein** [comment]
> context.perform(this, MyOps.BenWasHere)

@bencochran's solution specifically avoids giving `render` the side effect context directly. Only `initialState`, `onInputChanged`, and `WorkflowAction` get it. He explained why to me so I'll just let him explain himself instead of trying to paraphrase.

**Comment by zach-klippenstein** [comment]
Maybe we could just do something as simple as:
```kotlin
// Notice the lack of any generic type parameters.
interface SideEffect {
  fun perform()
}

fun enterState(
  state: StateT,
  emittingOutput: OutputT,
  withSideEffects: List<SideEffect> = emptyList()
)

class LogSideEffect(private val logger: Logger) : SideEffect {
  …
}

class MyWorkflow @Inject constructor(private val logSideEffect) {
  override fun render(…) {
    context.onEvent { enterState(Foo, withSideEffects = listOf(logSideEffect)) }
  }
}
```

zach-klippenstein:
Nvm, injectable side effects would not work for the enum `WorkflowAction` use case, since we would have to be injecting enum values.

bencochran:
> @bencochran's solution specifically avoids giving `render` the side effect context directly. Only `initialState`, `onInputChanged`, and `WorkflowAction` get it. He explained why to me so I'll just let him explain himself instead of trying to paraphrase.

For me it came down where I expect idempotence and where not. `render` is more of a transformation than an action IMO. It should describe the state of workers, subscriptions and the screen. The `initialState`, `didChange` and `Action.apply` are more “moments in time” and therefore make sense to me as a place for (the description of) one-time work.

Put slightly differently, without a lot of bookkeeping (like in the `Worker` example), I don’t see a way to say “based on this state, these should be the side effects that have been performed”

rjrjr:
It seems to me that the Operator sketch is exactly as declarative as workers already are. The render method is defining the operations that are expected to be performed for us when we enter this state, not actually doing any work. s/perform/operation/.

Am I missing something?

```kotlin
  fun render(
      input: I,
      state: S,
      context: WorkflowContext<...>
  ) : R = when(state) {
    is OppyState -> {
      context.operation(this, MyOps.BenWasHere)
    }
   // ...
  }
}
```

Bear in mind that besides the type param sadness of your proposal, it also relies heavily on the idiom of implementing WorkflowAction directly. We haven't done that _at all_ in the Kotlin code, and it's not clear to me that doing so can be made palatable with our current syntax. (We still have explicit event enums, but they're per screen, which has been pretty nice.)

zach-klippenstein:
@rjrjr In your example, `render` no longer declarative and idempotent. `Worker`s keep the render method idempotent because we diff the active workers.

rjrjr:
> @rjrjr In your example, `render` no longer declarative and idempotent. `Worker`s keep the render method idempotent because we diff the active workers.

It is though, or it's meant to be. I assumed similar semantics for perform as for any other work we ask the context to manage.

zach-klippenstein:
> It is though, or it's meant to be.

It can't be. "Side effect" as defined by this proposal doesn't have a result – they're fire-and-forget. Thus we have no way of tracking when they're done, and the API intentionally doesn't have a way to acknowledge that it's done.

rjrjr:
> Chatted offline about this. The main use case this proposal solves is when `WorkflowAction`s are defined externally, as in @bencochran's examples in the description. No Kotlin code is doing this yet, so we haven't had the need for this. I think this is a good solution to that specific problem, and am fine with doing it in Swift now, and following up in Kotlin if/when we start needing actions to be written that way.

I'm concerned about how our idioms are drifting.

 * Why does Swift code define actions externally, and why doesn't Kotlin code?
 * Is it even practical to write Kotlin actions externally?
 * If Swift adopts the event-type-per-screen pattern that we've been using in Kotlin, are external actions still tolerable?

zach-klippenstein:
> Why does Swift code define actions externally, and why doesn't Kotlin code?

Primarily, I think, because of history (converted Reactors) and habits.

> Is it even practical to write Kotlin actions externally?

Should be. 

Ultimately, we would converge on these patterns too. But I don't think it pays to go and do a bunch of changes just for the sake of changes. My hope is that, as teams begin to architect features for both platforms together, they would choose whether to use this sort of pattern or not consistently on both platforms.

My current stance is that I don't have the bandwidth to spend a whole bunch of time implementing this in Kotlin right now, especially when there is no immediate need for it. When that time comes, we can build it out. But I don't want to block Swift from doing it just because we haven't gotten there yet.

swift/Tooling/Templates/Workflow (Verbose).xctemplate/___FILEBASENAME___Workflow.swift
```diff
@@ -42,7 +42,7 @@ extension ___VARIABLE_productName___Workflow {
 
         typealias WorkflowType = ___VARIABLE_productName___Workflow
 
-        func apply(toState state: inout ___VARIABLE_productName___Workflow.State) -> ___VARIABLE_productName___Workflow.Output? {
+        func apply(toState state: inout ___VARIABLE_productName___Workflow.State, context: inout SideEffectContext<___VARIABLE_productName___Workflow>) -> ___VARIABLE_productName___Workflow.Output? {
```
davidapgar: In addition to updating the methods affected by this, we should likely add the definition of `SideEffect` and `perform` to the template (since it's the "verbose" one).

swift/Workflow/Sources/WorkflowNode.swift
```diff
@@ -19,7 +19,14 @@ final class WorkflowNode<WorkflowType: Workflow> {
         
         /// Get the initial state
         self.workflow = workflow
-        self.state = workflow.makeInitialState()
+
+        var context = SideEffectContext<WorkflowType>()
+        self.state = workflow.makeInitialState(context: &context)
+
+        /// Perform side effects
+        for sideEffect in context.sideEffects {
```
davidapgar: Consider pulling this into a helper, since we do it in a few places.

timdonnelly:
This seems like the right place to put side effects – I only question the `SideEffectContext` type.

Workflow actions can be described as `(State) -> (State, [SideEffect], Output?)`.

With the addition of `SideEffectContext` we are using three different models for the three return values:
- State is directly passed in as an `inout`
- `SideEffectContext` is passed in, and implementations are expected to call API on the context to perform side effects
- `Output?` is returned

In a previous iteration we had an `UpdateContext` type that would (in this world) hold both side effects and outputs:

```swift
func apply(toState state: inout WorkflowType.State, context: inout SideEffectContext<WorkflowType>) -> WorkflowType.Output?
```

would become:

```swift
// assign to context.output to emit an output event
func update(state: inout WorkflowType.State, context: inout UpdateContext<WorkflowType>)
```

or...

```swift
// context.state is now a thing
func update(context: inout UpdateContext<WorkflowType>)
```

I don't actually have a strong opinion on the right direction here, but my gut reaction is that the current proposal is a bit cumbersome.

zach-klippenstein:
Since Kotlin has finally caught up to Swift wrt recommending explicitly `WorkflowAction` implementations, this applies just as much to Kotlin now so I'd like to come up with a design that feels good in both languages before merging.

zach-klippenstein:
This is probably a terrible idea because it reverts back to the impure "use actions for side effects" state of affairs that we're in now, but…

If we're primarily concerned with solving the injection problem, what if we just pass the workflow/props in to `WorkflowAction.apply`? 

Unfortunately this would still add an uncomfortable amount of generics boilerplate in kotlin, ugh.

bencochran:
Ergonomically, passing the workflow/props into the action would be a step in the right direction. But I still cling to the purity of modeling and describing side-effects rather than performing them. Being able in a test to assert that side effects resulted from an action rather than having to mock out some external object is a huge testability improvement.

I also would worry that passing the workflow/props in tempting to have some of your state live in an external system, using an Action to pull that state off. Having be modeled as this PR suggests forces it to be a write-only path out of the Workflow world.

--------

I’ve taken a new stab at this PR, moving toward the style of:

```swift
// assign to context.output to emit an output event
func apply(toState state: inout WorkflowType.State, context: inout ActionContext<WorkflowType>)
```

which I think strikes the balance where `state` is still the primary “return value” but side-effects and output both flow through the same system.

Will update this PR with my current thinking soon.

zach-klippenstein:
> But I still cling to the purity of modeling and describing side-effects rather than performing them.

From the workflow's perspective, it's still describing side effects (by delegating to `WorkflowAction`s). But I get your point. I just can't see a way to add yet another generic type to the Kotlin types with a straight face – they're already so verbose.

zach-klippenstein:
I think the outcome of a few offline discussions is that we've decided not to pursue this. @bencochran If you agree, can we close this PR?

Issue #269, opened 2019-04-11
[swift] Fail in an obvious way if the WorkflowContext is used outside of `compose`


davidapgar:
Run into this a few times with people new to using workflows.

As is stands right now, if (for instance) one were to call `context.makeSink` in the closure for a screen callback, the sink won't be valid and thus the action will never be sent back. It ends up being surprising behavior, and can be a big hiccup wiring up UI back to a workflow.

eg:
```swift
    func compose(state:State, context: WorkflowContext<MyWorkflow>) -> MyScreen {

        return MyScreen(
            email: state.email,
            onEmailChange: { email in

                // This doesn't actually work, since the sink returned won't be wired up
                let sink = context.makeSink(of: Action.self)

                sink.send(.emailChanged(email))
            })
    }
```

I lean at least initially to it being a fatalError, since it makes it obvious very quickly that it won't work (instead of updating a field/tapping a button and there being no effect)

PR #301, opened 2019-04-18
[Swift] Rename `compose` to `render` and update docs.

### LINKED ISSUE #266: [swift] Rename compose method and concept in docs to render
The `compose` function was originally called `render` in the Swift implementation. After much discussion, we decided to rename to `compose` to avoid suggesting too strong an association with UI (it can return screen data, but can also be any arbitrary value).

However, this is dumb for a few reasons:
 - we're still calling the _type_ `Rendering`
 - "render" is a much more specific verb to represent this specific concept than the super-generic "compose"
 - It turns out Swift never actually got around to renaming everything anyway.

For these reasons, I propose renaming the `State{ful|less}Workflow.compose` method back to `render`.

 - Kotlin issue: #265

davidapgar:
Closes #266


timdonnelly:
💯

swift/Workflow/Sources/Workflow.swift
```diff
@@ -59,15 +59,15 @@ public protocol Workflow: AnyWorkflowConvertible {
...
+    /// Called to "render" the current state into `Rendering`. A workflow's `Rendering` type is commonly a view or
     /// screen model.
     ///
     /// - Parameter state: The current state.
     /// - Parameter context: The workflow context is the composition point for the workflow tree. To use a nested
     ///                      workflow, instantiate it based on the current state. The newly instantiated workflow is
     ///                      then used to invoke `context.render(_ workflow:)`, which returns the child's `Rendering`
     ///                      type after creating or updating the nested workflow.
-    func compose(state: State, context: WorkflowContext<Self>) -> Rendering
+    func render(state: State, context: WorkflowContext<Self>) -> Rendering
```
davidapgar: @zach-klippenstein @timdonnelly @bencochran With the name of `render` here now, does the name of `WorkflowContext` still make sense?

I am starting to think that `RenderContext` would be a better type name - as it evokes that this that is context is *only* valid during the call to render (which has been a usage issue, of capturing and trying to use the context after the render pass is complete).

Not necessary to do it in this change, but wanted to discuss if it makes sense. This came to mind while talking to @bencochran regarding the SideEffects PR, as we now have different contexts, none of which are the "workflow"'s context - more just a context scoped to a single method.

zach-klippenstein: I would be super happy to rename to `RenderContext`.

AFAIK we haven't had as much of a problem with people trying to capture it, but if we start to see that one way to discourage it in Kotlin is to make it a receiver on `render`. It's a somewhat common pattern for `FooContext` types to be receivers on `foo` functions in kotlin. We haven't done that yet because we prefer making it clear what you're calling methods on, but the stdlib authors obviously feel differently about this.

davidapgar: It's only happened a few times, but mostly when someone is trying this out for the first time. The current behavior is "surprising" as a sink created after `render` is complete doesn't go anywhere, which leads to the behavior where events from a Screen don't get back to the workflow.

I have https://github.com/square/workflow/issues/269 as an issue to make it assert if an invalid context is used, to at least make it obvious what's happening. But the naming (I would hope) is an extra signal about the scope.

zach-klippenstein: We throw now as well in that case (as of #273). Filed #303 and #304 to track the context rename.

davidapgar: Updated this PR to also handle the name update to `RenderContext` as well as the docs.

PR #305, opened 2019-04-18
[#269] Invalidate WorkflowContext after `render` is called

### LINKED ISSUE #269: [swift] Fail in an obvious way if the WorkflowContext is used outside of `compose`

timdonnelly:
This immediately explodes (via `fatalError`) if a `WorkflowContext` is used after `render` returns.

This fixes #269

### DIFF
--- a/swift/Workflow/Sources/SubtreeManager.swift
+++ b/swift/Workflow/Sources/SubtreeManager.swift
@@ -41,6 +41,8 @@ extension WorkflowNode {
 
             /// Pass the context into the closure to allow a render to take place
             let rendering = actions(wrapped)
+            
+            wrapped.invalidate()
 
             /// After the render is complete, assign children using *only the children that were used during the render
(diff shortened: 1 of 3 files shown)


bencochran:
Thanks for this!

PR #306, opened 2019-04-18
Fix for WorkflowHost.Factory not using baseContext.

zach-klippenstein:
Commit cf3ae7bfa5f1d3778d057d5d200ee4126ba4cf58 accidentally stopped including
`baseContext` in the context for the workflow loop coroutine. This caused everything
not passing a dispatcher into `run` directly to default to the `Default` dispatcher,
which made a lot of things suddenly happen a lot lazier, and forced us to add a bunch
of additional calls in unit tests.

This PR adds `baseContext` back in, and also adds a default `CoroutineName` element
to aid in debugging.

### DIFF
--- a/kotlin/workflow-rx2/src/test/java/com/squareup/workflow/rx2/SubscriptionsTest.kt
+++ b/kotlin/workflow-rx2/src/test/java/com/squareup/workflow/rx2/SubscriptionsTest.kt
@@ -169,8 +169,6 @@ class SubscriptionsTest {
 
   @Test fun `observable reports emissions`() {
     workflow.testFromStart(true) { host ->
-      // Get the next rendering to unblock the pipeline so the subscription can actually occur.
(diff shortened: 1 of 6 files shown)


kotlin/workflow-rx2/src/test/java/com/squareup/workflow/rx2/SubscriptionsTest.kt
```diff
@@ -169,8 +169,6 @@ class SubscriptionsTest {
 
   @Test fun `observable reports emissions`() {
     workflow.testFromStart(true) { host ->
-      // Get the next rendering to unblock the pipeline so the subscription can actually occur.
-      host.awaitNextRendering()
```
zach-klippenstein: These extra calls were all added to deal with the additional laziness. Using the correct dispatcher again means we can revert them.

PR #300, opened 2019-04-17
Merge CHANGELOG files into one, update RELEASING instructions to include making GitHub release.

zach-klippenstein:
When you create a release on GitHub, you can provide a rich markdown description of the release that will show up directly on the GitHub releases page. I have done this for a single release, and it looks much better. I've updated the RELEASING instructions to include this step. This also requires merging the kotlin and swift changelogs (since the GitHub interface only gives you one location), so we might as well merge our CHANGELOG files as well.

PR #307, opened 2019-04-18
Updated changelog and version numbers for v0.14.1.

zach-klippenstein:


kotlin/CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 Change Log
 ==========
 
+## Version 0.14.1
```
davidapgar: These should be in the root CHANGELOG as https://github.com/square/workflow/pull/300 has landed.

davidapgar: Oh, this is a hotfix which is why it goes into the previous. Disregard.

zach-klippenstein: This is branched off the commit that finishes the 0.14.0 release, I didn't want to bring anything else in.

PR #314, opened 2019-04-23
Fix SubtreeManager not saving new case handlers.

### LINKED ISSUE #261: Stale lambdas on repeated WorkflowContext.compose() calls?
Debugging with @mbrubin56-square, and I think we've found a problem with repeated calls to WorkflowContext.compose() on a child, maybe related to it emitting multiple output and receiving multiple input.

The flow is something like:

 * when(state) is ParentStateA, we start the child with a lambda that captures ParentStateA'
 * The child emits output, so that lambda fires and takes us to ParentStateB
 * when(state) is ParentStateB, a compose call keeps the child running and does other work
 * That work is done we enterState(ParentStateA''), with new information
 * when(state) is ParentStateA, the original compose call is made, with a new lambda capturing ParentState''
 * The child emits more output, the lambda fires, but it's the lambda from the start of this recipe, that capture ParentStateA'. 
 * We crash because the expected new information is null.

He's working around this by giving the parent workflow a `latestState` field that gets updated from the ParentStateA case, and having the lambda capture that instead. Which is gross.

zach-klippenstein:
`SubtreeManager` is responsible for managing a `WorkflowNode`'s child `Workflow`s.
It uses `LifetimeTracker` to track children across render passes, start new ones,
and cancel old ones. It stores the `WorkflowOutputCase` as the `LifetimeTracker`'s
`Factory` type – "factories" have keys for deduping, and are used to create new
"disposables" when new ones are tracked. The `WorkflowOutputCase` also stores the
`handler` lambda that defines how to map a child workflow's output to a parent
`WorkflowAction`. This handler is re-created on every render pass (because it
closes over the workflow's state), and so the `LifetimeTracker` should always be
storing the most recent _instance_ of the `WorkflowOutputCase`. However, previously
it would only store the first factory for a given key. This meant that if a child
was rendered over multiple render passes, and the state changed between those passes,
when the child emitted an output after the state change it the handler lambda would
be stale and thus have a reference to the wrong state.

This change fixes the issue by making `LifetimeTracker` always replace the `FactoryT`
for a given `KeyT` on every call to `ensure` or `track`.

Fixes #261.

### DIFF
--- a/kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/LifetimeTracker.kt
+++ b/kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/LifetimeTracker.kt
@@ -43,11 +43,11 @@ internal class LifetimeTracker<FactoryT, KeyT : Any, DisposableT>(
   private val factoriesByKey = mutableMapOf<KeyT, FactoryT>()
   private val disposablesByKey = mutableMapOf<KeyT, DisposableT>()
 
-  val lifetimes: Map<FactoryT, DisposableT>
+  val lifetimes: List<Pair<FactoryT, DisposableT>>
     get() = factoriesByKey.entries.map { (key, provider) ->
       val lifetime = disposablesByKey.getValue(key)
       provider to lifetime
-    }.toMap()
+    }
 
   /**
    * Additively track a possibly-new provider.
@@ -89,12 +89,15 @@ internal class LifetimeTracker<FactoryT, KeyT : Any, DisposableT>(
     key: KeyT,
     factory: FactoryT
   ): DisposableT {
-    return if (key !in factoriesByKey) {
-      factoriesByKey[key] = factory
+    val existingFactory = factoriesByKey.put(key, factory)
+    return if (existingFactory == null) {
+      // This is the first time we're seeing this key/factory, start it up and start tracking.
+      check(key !in disposablesByKey)
       start(factory).also {
         disposablesByKey[key] = it
       }
     } else {
+      // We're already tracking this factory.
       disposablesByKey.getValue(key)
     }
   }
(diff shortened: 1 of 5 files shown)


kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/LifetimeTracker.kt
```diff
@@ -43,11 +43,11 @@ internal class LifetimeTracker<FactoryT, KeyT : Any, DisposableT>(
   private val factoriesByKey = mutableMapOf<KeyT, FactoryT>()
   private val disposablesByKey = mutableMapOf<KeyT, DisposableT>()
 
-  val lifetimes: Map<FactoryT, DisposableT>
+  val lifetimes: List<Pair<FactoryT, DisposableT>>
```
zach-klippenstein: Doesn't make sense to think of `FactoryT` as a key, and this isn't actually used as a map anywhere other than unit tests, it's intended to be, and used as, a list of pairs, so let's call a spade a spade.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/LifetimeTracker.kt
```diff
@@ -89,12 +89,15 @@ internal class LifetimeTracker<FactoryT, KeyT : Any, DisposableT>(
     key: KeyT,
     factory: FactoryT
   ): DisposableT {
-    return if (key !in factoriesByKey) {
-      factoriesByKey[key] = factory
+    val existingFactory = factoriesByKey.put(key, factory)
```
zach-klippenstein: Semantic change is just pulling this statement up from the first `if` case below, so it's _always_ updated. This is the actual fix.

PR #315, opened 2019-04-23
Give WorkflowTester a sendInput method to update the root workflow-un…

zach-klippenstein:
…der-test's input.


kotlin/workflow-testing/src/test/java/com/squareup/workflow/ChannelSubscriptionsIntegrationTest.kt
```diff
@@ -46,7 +46,7 @@ class ChannelSubscriptionsIntegrationTest {
    */
   private class SubscriberWorkflow(
     private val channel: Channel<String>
-  ) : StatefulWorkflow<Boolean, Boolean, ChannelUpdate<String>, (Boolean) -> Unit>() {
+  ) : StatelessWorkflow<Boolean, ChannelUpdate<String>, Unit>() {
```
zach-klippenstein: Makes this test workflow so much simpler!

vRallev: Nice

PR #317, opened 2019-04-24
Forgot to add ticket links in #289.

zach-klippenstein:


kotlin/workflow-core/src/main/java/com/squareup/workflow/Worker.kt
```diff
@@ -182,7 +182,7 @@ interface Worker<out T> {
       key: String = "",
         // This could be crossinline, but there's a coroutines bug that will cause the coroutine
         // to immediately resume on suspension inside block when it is crossinline.
-        // TODO link to KT ticket.
+        // See https://youtrack.jetbrains.com/issue/KT-31197.
```
kirillzh: TODO?

PR #318, opened 2019-04-24
Workflow tester improvements

### LINKED ISSUE #292: withNextRendering et al aren't necessary
withNextRendering is just a convenience to call awaitNextRendering and then call let. We can do that ourselves, let's inline it.

zach-klippenstein:
Three commits:
 1. Remove WorkflowTester.with* methods. Closes #292.
 2. Make WorkflowTester the receiver of the lambdas to Workflow.test* methods.
     I initially made it an explicit argument because I thought it increased readability,
     but after writing a bunch of tests, and occasionally using different names for that
     argument, since the entire point of the lambda passed to `Workflow.test*` methods is
     to operator on the `WorkflowTester` I think it's actually more readable as a receiver.
 3. Merge WorkflowTesting.kt into WorkflowTester.kt.


kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/WorkflowTester.kt
```diff
@@ -231,3 +183,110 @@ class WorkflowTester<InputT : Any, OutputT : Any, RenderingT : Any> @TestOnly in
...
+  } catch (e: Throwable) {
+    error = e
+    throw e
+  } finally {
+    if (error != null) {
+      // TODO https://github.com/square/workflow/issues/188 Stop using parameterized cancel.
+      @Suppress("DEPRECATION")
+      val cancelled = context.cancel(error)
+      if (!cancelled) {
+        val cancellationCause = context[Job]!!.getCancellationException()
```
kirillzh: `requireNotNull()` would give better error in case of NPE.

kirillzh:
Neat, neat!

kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/WorkflowTester.kt
zach-klippenstein: It's not possible to have a null here, since we explicitly give the context a `Job` on line 262. If it does have a null, someone edited this file and broke something, and I'm not worried about the error message in that case.

PR #321, opened 2019-04-25
Support out-projected types for Rx Worker extensions.

zach-klippenstein:


kotlin/workflow-rx2/src/main/java/com/squareup/workflow/rx2/RxWorkers.kt
```diff
@@ -24,20 +24,20 @@ import kotlinx.coroutines.reactive.openSubscription
 import kotlinx.coroutines.rx2.await
 import kotlinx.coroutines.rx2.openSubscription
 
-inline fun <reified T : Any> Observable<T>.asWorker(key: String = ""): Worker<T> =
+inline fun <reified T : Any> Observable<out T>.asWorker(key: String = ""): Worker<T> =
```
vRallev: Is this really necessary? I remember there are problems with some operators. Would this compile?
```kotlin
Observable.just("2").startWith("1").asWorker(..)
```

zach-klippenstein: Yep

kotlin/workflow-rx2/src/test/java/com/squareup/workflow/rx2/RxWorkersTest.kt
```diff
@@ -259,7 +264,7 @@ class RxWorkersTest {
 
   @Test fun `single emits`() {
     val subject = SingleSubject.create<String>()
-    val worker = subject.asWorker()
+    val worker = (subject as Single<out String>).asWorker()
```
kirillzh: Nice!

PR #323, opened 2019-04-26
Fix the nullability of the returned Workers from Rx extensions.

zach-klippenstein:


kotlin/workflow-rx2/src/main/java/com/squareup/workflow/rx2/RxWorkers.kt
```diff
@@ -20,24 +20,71 @@ import io.reactivex.Flowable
...
+ * The [Single] will be subscribed to when the [Worker] is started, and disposed when it is
+ * cancelled.
+ *
+ * RxJava doesn't allow nulls, but it can't express that in its types. The receiver type parameter
+ * is nullable so that the resulting [Worker] is non-nullable instead of having
+ * platform nullability.
+ */
+inline fun <reified T : Any> Single<out T?>.asWorker(key: String = ""): Worker<T> =
+// This !! works because RxJava types don't actually allow nulls, it's just that they can't
+  // express that in their types because Java.
```
kirillzh: `because of`

kotlin/workflow-rx2/src/main/java/com/squareup/workflow/rx2/RxWorkers.kt
```diff
@@ -20,24 +20,71 @@ import io.reactivex.Flowable
...
+  Worker.from(key) { await()!! }
```
kirillzh: Would personally prefer `requireNotNull()` :)

zach-klippenstein: I definitely wouldn't want `requireNotNull` here. `requireNotNull` throws an `IllegalArgumentException`. `Single` should never emit a null. If it did for some reason, it would not be an `IllegalArgumentException`, something would have gone extremely wrong somewhere and that would be the wrong exception to throw.

kirillzh: Got it, this makes sense.

PR #324, opened 2019-04-26
Fix onWorkerOutput to noop on finish instead of throw.

zach-klippenstein:
This was the previous behavior of all the single/optional-output
`RenderContext` extensions, it's much more convenient this way.


zach-klippenstein:
This should be the last thing blocking the 0.15 release, all our internal tests pass now.

kotlin/workflow-core/src/main/java/com/squareup/workflow/RenderContext.kt
```diff
@@ -148,7 +149,7 @@ fun <InputT : Any, StateT : Any, OutputT : Any, ChildRenderingT : Any>
       child: Workflow<InputT, Nothing, ChildRenderingT>,
       input: InputT,
       key: String = ""
-    ): ChildRenderingT = renderChild(child, input, key) { WorkflowAction.noop() }
+    ): ChildRenderingT = renderChild(child, input, key) { noop() }
```
kirillzh: `noop` sounds so cute

PR #333, opened 2019-05-01
Fixed cardinality of worker output in docs.

zach-klippenstein:


Kiskae:
The replacement might be slightly confusing since the current implementation does not guarantee any events. Calling `emitOutput` isn't strictly required and the `Finished` event will only be fired if the worker terminates. So explicitly mentioning it will emit "one or more" might be incorrect.

zach-klippenstein:
Good point. #336

PR #334, opened 2019-05-02
Remove non-nullable restriction from InputT, StateT, and RenderingT.

zach-klippenstein:
Making these types non-nullable doesn't add any inherent value or safety. There may be workflows
that want to define nullable types, so we should let them. Workflows that define their types to be non-nullable still get all the null safety they expect.

### DIFF
--- a/kotlin/workflow-core/src/main/java/com/squareup/workflow/RenderContext.kt
+++ b/kotlin/workflow-core/src/main/java/com/squareup/workflow/RenderContext.kt
@@ -38,7 +38,7 @@ import com.squareup.workflow.WorkflowAction.Companion.noop
  *
  * See [renderChild].
  */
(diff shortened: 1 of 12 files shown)


rjrjr:
Was there a specific issue that drove this change? And if nullability is okay now, why isn't it okay for `OutputT`?

zach-klippenstein:
> Was there a specific issue that drove this change?

I was working on the legacy integration, and it would have been convenient to use `null` to represent the lack of a value. In general, I would expect a whole host of workflows to have an "optional" state or rendering instead of making a custom union type just for that.

> if nullability is okay now, why isn't it okay for OutputT?

`OutputT` is the only type in the set that represents an "event" instead of a declarative value. There might be a use case for null, but in general "the lack of a value" for outputs is represented by just not emitting any.

PR #346, opened 2019-05-14
Upgrade coroutines dependency to latest version.

### LINKED ISSUE #188: Stop using parameterized coroutine cancellation that was deprecated in kotlin 1.3

zach-klippenstein:
Closes #188 because the deprecated `cancel` method that took `Throwable` has been removed.

The only real thing this "broke" was how legacy workflow's `switchMapState` handles channel cancellation,
now we have to be a bit more explicit. This is why channels shouldn't be used as streams…

### DIFF
--- a/kotlin/workflow-runtime/src/main/java/com/squareup/workflow/WorkflowHost.kt
+++ b/kotlin/workflow-runtime/src/main/java/com/squareup/workflow/WorkflowHost.kt
@@ -27,7 +27,6 @@ import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.InternalCoroutinesApi
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.NonCancellable.isActive
-import kotlinx.coroutines.cancel
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.channels.ReceiveChannel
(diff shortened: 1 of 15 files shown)


kotlin/workflow-runtime/src/main/java/com/squareup/workflow/WorkflowHost.kt
```diff
@@ -207,9 +208,7 @@ suspend fun <InputT, StateT, OutputT : Any, RenderingT> runWorkflowTree(
   } catch (e: Throwable) {
     // For some reason the exception gets masked if we don't explicitly pass it to cancel the
```
rjrjr: Is this still true after the upgrade?

zach-klippenstein: Seems like no – the tests all pass if I remove this catch. Sweet!

PR #349, opened 2019-05-16
Introduce alternative testing infrastructure to test single render passes more directly.

zach-klippenstein:
Given a `StatefulWorkflow` or `StatelessWorkflow`, call `.testRender()` on it.
    Pass the `InputT`, and for `StatefulWorkflow`s also pass the `StateT`.
    This returns a `TestRenderResult` which provides access to the actual rendering
    as well as methods to assert that children or workers were rendered and invoke
    their output handlers.

### DIFF
--- a/kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/Behavior.kt
+++ b/kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/Behavior.kt
@@ -28,7 +28,7 @@ import kotlinx.coroutines.Deferred
  *
  * @see RealRenderContext
  */
-internal data class Behavior<StateT, out OutputT : Any>(
+data class Behavior<StateT, out OutputT : Any> internal constructor(
   val childCases: List<WorkflowOutputCase<*, *, StateT, OutputT>>,
   val workerCases: List<WorkerCase<*, StateT, OutputT>>,
   val nextActionFromEvent: Deferred<WorkflowAction<StateT, OutputT>>
@@ -40,7 +40,7 @@ internal data class Behavior<StateT, out OutputT : Any>(
       ChildOutputT : Any,
       ParentStateT,
       out ParentOutputT : Any
-      >(
+      > internal constructor(
         val workflow: Workflow<*, ChildOutputT, *>,
         val id: WorkflowId<ChildInputT, ChildOutputT, *>,
         val input: ChildInputT,
@@ -52,7 +52,7 @@ internal data class Behavior<StateT, out OutputT : Any>(
       }
       // @formatter:on
 
-  data class WorkerCase<T, StateT, out OutputT : Any>(
+  data class WorkerCase<T, StateT, out OutputT : Any> internal constructor(
     val worker: Worker<T>,
     val key: String,
     val handler: (OutputOrFinished<T>) -> WorkflowAction<StateT, OutputT>
--- a/kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/RealRenderContext.kt
+++ b/kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/RealRenderContext.kt
@@ -27,8 +27,10 @@ import kotlinx.coroutines.CompletableDeferred
 
 /**
  * An implementation of [RenderContext] that builds a [Behavior] via [buildBehavior].
+ *
+ * Not for general application use.
  */
-internal class RealRenderContext<StateT, OutputT : Any>(
+class RealRenderContext<StateT, OutputT : Any>(
   private val renderer: Renderer<StateT, OutputT>
 ) : RenderContext<StateT, OutputT> {
 
--- a/kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/WorkflowId.kt
+++ b/kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/WorkflowId.kt
@@ -30,17 +30,21 @@ internal typealias AnyId = WorkflowId<*, *, *>
  * Value type that can be used to distinguish between different workflows of different types or
  * the same type (in that case using a [name]).
  */
-internal data class WorkflowId<in InputT, out OutputT : Any, out RenderingT>
+data class WorkflowId<in InputT, out OutputT : Any, out RenderingT>
 @PublishedApi
 internal constructor(
   internal val type: KClass<out Workflow<InputT, OutputT, RenderingT>>,
   internal val name: String = ""
-)
+) {
+  constructor(
+    workflow: Workflow<InputT, OutputT, RenderingT>,
+    name: String = ""
+  ) : this(workflow::class, name)
+}
 
 @Suppress("unused")
 internal fun <W : Workflow<I, O, R>, I, O : Any, R>
-    W.id(key: String = ""): WorkflowId<I, O, R> =
-  WorkflowId(this::class, key)
+    W.id(key: String = ""): WorkflowId<I, O, R> = WorkflowId(this, key)
 
 internal fun WorkflowId<*, *, *>.toByteString(): ByteString = Buffer()
     .also { sink ->
--- a/kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/MockChildWorkflow.kt
+++ b/kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/MockChildWorkflow.kt
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2019 Square Inc.
+ *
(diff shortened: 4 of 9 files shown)


kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/Behavior.kt
```diff
@@ -28,7 +28,7 @@ import kotlinx.coroutines.Deferred
  *
  * @see RealRenderContext
  */
-internal data class Behavior<StateT, out OutputT : Any>(
+data class Behavior<StateT, out OutputT : Any> internal constructor(
```
zach-klippenstein: Sad to make this public, but don't want to duplicate the code in the test module.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/RealRenderContext.kt
```diff
@@ -27,8 +27,10 @@ import kotlinx.coroutines.CompletableDeferred
 
 /**
  * An implementation of [RenderContext] that builds a [Behavior] via [buildBehavior].
+ *
+ * Not for general application use.
  */
-internal class RealRenderContext<StateT, OutputT : Any>(
+class RealRenderContext<StateT, OutputT : Any>(
```
zach-klippenstein: Very sad to make this public. See above comment.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/WorkflowId.kt
```diff
@@ -30,17 +30,21 @@ internal typealias AnyId = WorkflowId<*, *, *>
  * Value type that can be used to distinguish between different workflows of different types or
  * the same type (in that case using a [name]).
  */
-internal data class WorkflowId<in InputT, out OutputT : Any, out RenderingT>
+data class WorkflowId<in InputT, out OutputT : Any, out RenderingT>
```
zach-klippenstein: 😢

kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/MockWorker.kt
```diff
@@ -0,0 +1,20 @@
+package com.squareup.workflow.testing
```
rjrjr: copyright

kotlin/workflow-testing/src/test/java/com/squareup/workflow/testing/RenderTesterTest.kt
```diff
@@ -0,0 +1,144 @@
...
+    assertEquals("input=foo, state=bar", result.rendering)
+  }
+
+  @Test fun `assert no composition`() {
+    val workflow = Workflow.stateless<Nothing, Unit> { Unit } as StatelessWorkflow
+
+    val result = workflow.testRender()
+
+    result.assertNoWorkflowsRendered()
+    result.assertNoWorkersRendered()
```
rjrjr: Nit: do we really render a worker? Don't we run them?

kotlin/workflow-testing/src/test/java/com/squareup/workflow/testing/RenderTesterTest.kt
```diff
@@ -0,0 +1,144 @@
+package com.squareup.workflow.testing
```
rjrjr: copyright

kotlin/workflow-testing/src/test/java/com/squareup/workflow/testing/RenderTesterTest.kt
zach-klippenstein: Spent a whole 5 minutes googling the correct tenses of "to run", I think I got it right now.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/RealRenderContext.kt
vRallev: I brought this up in the Kotlin Slack. I should file a ticket: https://kotlinlang.slack.com/archives/C0B8PUJGZ/p1555962527006200

kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/MockChildWorkflow.kt
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * Note this workflow can not actually emit any output itself. Use
+ * [TestRenderResult.executeWorkflowActionFromOutput] to evaluate output handlers.
+ *
+ * @param renderer Function that is invoked in each render pass to calculate the rendering.
+ *
+ * @see StatefulWorkflow.testRender
+ * @see com.squareup.workflow.StatelessWorkflow.testRender
+ */
+class MockChildWorkflow<I, R>(private val renderer: (I) -> R) : Workflow<I, Nothing, R> {
```
vRallev: `FakeChildWorkflow`?

kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/MockWorker.kt
```diff
@@ -0,0 +1,39 @@
...
+ * A mock implementation of [Worker] for use in tests with `testRender` and [TestRenderResult].
+ *
+ * Not this [Worker] can not actually emit any output itself. Use
+ * [TestRenderResult.executeWorkerActionFromOutput] or
+ * [TestRenderResult.executeWorkerActionFromFinish] to evaluate output handlers.
+ *
+ * @see com.squareup.workflow.StatefulWorkflow.testRender
+ * @see com.squareup.workflow.StatelessWorkflow.testRender
+ */
+class MockWorker<T>(val name: String) : Worker<T> {
```
vRallev: `FakeWorker`?

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/RealRenderContext.kt
zach-klippenstein: When you do, lemme know I will vote.

kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/MockChildWorkflow.kt
zach-klippenstein: It actually records the input though so you can check it later, doesn't that make it technically a "mock"?

rjrjr: Yeah, Zach's right. Changed my vote. 

https://blog.pragmatists.com/test-doubles-fakes-mocks-and-stubs-1a7491dfa3da

kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/MockWorker.kt
```diff
@@ -0,0 +1,39 @@
...
+ */
+package com.squareup.workflow.testing
+
+import com.squareup.workflow.Worker
+import com.squareup.workflow.Worker.Emitter
+
+/**
+ * A mock implementation of [Worker] for use in tests with `testRender` and [TestRenderResult].
+ *
+ * Not this [Worker] can not actually emit any output itself. Use
```
davidapgar: Typo on `Not` -> `Note` (I believe this should read "Note this [Worker] can not actually emit any output...")

kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/MockChildWorkflow.kt
vRallev: Oh, yes. This blog is handy 👍

zach-klippenstein: Although we lose symmetry with the worker then, which actually is a fake…

I want to add more instrumentation on them anyway, so they'll both be mocks.

zach-klippenstein: @vRallev @rjrjr  I just pushed another commit that changes the API a little. Now you call `assert*` and `execute*Output*` directly on the workers/workflows themselves, from within a block. Please take another look.

kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/MockWorker.kt
zach-klippenstein: Fixed.

rjrjr:
Still LG with the tidier syntax (extension on the fakes),

PR #350, opened 2019-05-16
WIP: Block stale events.

zach-klippenstein:


zach-klippenstein:
Swift is doing something similar: https://github.com/square/workflow/pull/362#discussion_r286611769

zach-klippenstein:
This is the wrong solution. The correct one is #541: allowing sinks of the same type to be re-used across render passes.

PR #355, opened 2019-05-20
Ray/runner cleanup

rjrjr:
Some preliminary tidying of `WorkflowActivityRunner` to prep for fragment work. In particular, one less parameter type.


kotlin/samples/tictactoe/android/src/main/java/com/squareup/sample/mainactivity/MainActivity.kt
```diff
@@ -31,7 +31,7 @@ import com.squareup.workflow.ui.setContentWorkflow
 @Suppress("EXPERIMENTAL_API_USAGE")
 class MainActivity : AppCompatActivity() {
   private lateinit var component: MainComponent
-  private lateinit var workflowViewModel: WorkflowActivityRunner<*, *>
+  private lateinit var workflowRunner: WorkflowActivityRunner<*>
```
zach-klippenstein: Nice catch

PR #357, opened 2019-05-20
Make the RenderContext parameter to Workflow.stateless the receiver instead.

zach-klippenstein:
Looking for feedback on whether this is actually nicer. This feels nice to me:
 - "context" types are often receivers in these sort of helper builder higher-order functions.
 - Don't need a separate override to make the input argument optional, since there's only one argument now.


zach-klippenstein:
I didn't change `StatelessWorkflow` itself either. Staying more conservative on actual methods that belong to classes and just passing everything in explicitly. The whole point of the lambda-y helpers is to be more concise though, so a little more syntactic smoke & mirrors seems appropriate. 

Taking inspiration from the Flow API, which passes the `FlowEmitter` in as a normal argument to `FlowCollector.collect`, but makes it a receiver in the sugary helpers.

PR #358, opened 2019-05-21
Ray/workflow fragment 344

rjrjr:
Introduces WorkflowFragment, and couple of new minimal sample apps.


kotlin/samples/hello-workflow/build.gradle
```diff
@@ -0,0 +1,43 @@
...
+ */
+apply plugin: 'com.android.application'
+apply plugin: 'kotlin-android'
+
+android rootProject.ext.defaultAndroidConfig
+
+android {
+  defaultConfig {
+    applicationId "com.squareup.sample.helloworkflow"
+    minSdkVersion 26
```
vRallev: Is there a specific reason for the high min SDK?

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowFragment.kt
```diff
@@ -0,0 +1,99 @@
...
+import android.support.v4.app.Fragment
+import android.support.v4.app.FragmentActivity
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import com.squareup.workflow.Workflow
+import io.reactivex.Flowable
+
+@ExperimentalWorkflowUi
+abstract class WorkflowFragment<InputT, OutputT : Any> : Fragment() {
```
vRallev: I'm not a fan of base classes, but I also don't see how to do better here.

kotlin/samples/hello-workflow/build.gradle
rjrjr: Screen Shot 2019-05-21 at 8 55 29 AM

rjrjr: Just a bit of fun in the sample code.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowFragment.kt
rjrjr: Same. I used a support class for the activity use case, but for fragments this seemed like the most natural fit. And at least the bulk of it is in the view model.

rjrjr:
Update fixes a couple of lint issues (unused imports mostly), and adds missing compiler flag to quiet warning about use of `@Experimental`.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowRunner.kt
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+package com.squareup.workflow.ui
+
+import io.reactivex.Observable
+
+@ExperimentalWorkflowUi
+interface WorkflowRunner<out OutputT> {
+  /**
+   * A stream of the [output][OutputT] values emitted by the running  [Workflow].
```
zach-klippenstein: Nit: double space before `[Workflow]`

kotlin/samples/hello-workflow-fragment/src/main/java/com/squareup/sample/helloworkflowfragment/HelloWorkflow.kt
```diff
@@ -0,0 +1,67 @@
...
+    val message: String,
+    val onClick: (Unit) -> Unit
+  )
+
+  override fun initialState(
+    input: Unit,
+    snapshot: Snapshot?
+  ): State {
+    return snapshot?.bytes?.parse { source ->
+      if (source.readInt() == 1) Hello else Goodbye
```
zach-klippenstein: Not for this PR, but once we're on 1.3.40 we can use `Parcelize` for these.

rjrjr:
As I write the docs, the need to work with Flowable for the input is already getting obnoxious. Going to add some conveniences for Observable like @zach-klippenstein suggested.

kotlin/workflow-ui-android/build.gradle
```diff
@@ -27,6 +27,10 @@ android.packagingOptions {
   exclude 'META-INF/atomicfu.kotlin_module'
 }
 
+tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
+  kotlinOptions.freeCompilerArgs += ["-Xuse-experimental=kotlin.Experimental"]
+}
+
```
zach-klippenstein: Why `tasks.withType(…)` instead of `compileKotlin`?

```groovy
compileKotlin {
    kotlinOptions {
        freeCompilerArgs += "-Xuse-experimental=org.mylibrary.ExperimentalMarker"
    }
}
```
_from https://kotlinlang.org/docs/reference/experimental.html#module-wide-use_

zach-klippenstein: Also shouldn't we only need to allowlist our own `ExperimentalWorkflowUi` annotation?

rjrjr: Didn't see that recipe in my five seconds of Googling, will fix. 

WRT why, compiler warnings say that you have to opt into the use of the @Experimental annotation itself. 

And I actually don't want to whitelist _our_ annotation, b/c the warnings it causes are useful to ensure we've put the annotation in enough places.

rjrjr: https://stackoverflow.com/questions/44141076/compilekotlin-block-in-build-gradle-file-throws-error-could-not-find-method-com

zach-klippenstein: 🤦‍♂️🤦‍♂️🤦‍♂️🤦‍♂️🤦‍♂️🤦‍♂️🤦‍♂️🤦‍♂️

rjrjr:
Update tweaks kdoc on `WorkflowRunner` interface.

kotlin/samples/hello-workflow-fragment/src/main/java/com/squareup/sample/helloworkflowfragment/HelloWorkflow.kt
rjrjr: #359

PR #361, opened 2019-05-21
Warnings are now errors.

rjrjr:
Suppresses delegation warnings like crazy, but only in legacy code. Good job
us!

### DIFF
--- a/kotlin/build.gradle
+++ b/kotlin/build.gradle
@@ -156,4 +156,14 @@ subprojects {
   afterEvaluate {
     project.tasks.findByName('check')?.dependsOn 'detekt'
   }
+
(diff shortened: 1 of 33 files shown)


kotlin/build.gradle
```diff
@@ -156,4 +156,11 @@ subprojects {
   afterEvaluate {
     project.tasks.findByName('check')?.dependsOn 'detekt'
   }
+
+  tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
+    kotlinOptions {
+      kotlinOptions.allWarningsAsErrors = true
+      freeCompilerArgs += "-Xuse-experimental=kotlin.Experimental"
```
rjrjr: Should add a comment explaining what this means.

PR #362, opened 2019-05-22
Handle `Actions` from a sink synchronously.

### LINKED ISSUE #348: Swift workflow loop updates asynchronously, can cause lost events
tl;dr: Handling an action and having a render pass occur yields the main loop, which can cause UI events to be missed, as the previous sink will be invalid, and the updated sink won't be set on the UI side.

Longer form example that shows the behavior:
```swift
// The state of the text field is managed in the workflow. Updates to it when edited are sent back via the `.didChange` closure.
struct ExampleScreen: Screen {
  var text: String
  var didChange: (String) -> Void
}

final class ExampleViewController /*...*/ {
  let textField: UITextField

  /* snipped */

    override func viewDidLoad() {
        super.viewDidLoad()

        // Any time the text changes, send the event back to the workflow
        textField.addTarget(
            self,
            action: #selector(textDidChange(sender:)),
            for: .editingChanged)

        // rest of viewDidLoad
    }

    @objc private func textDidChange(sender: UITextField) {
        guard let text = sender.text else { return }
        screen.onTextChanged(text)
    }

    override func screenDidChange(from previousScreen: PlaceholderScreen) {
        // Update the text field state based on the state from the workflow.
        textField.text = screen.text
    }
}
```

Relevant parts of workflow implementation:
```swift
// MARK: Actions

extension RootWorkflow {

    enum Action: WorkflowAction {

        typealias WorkflowType = RootWorkflow

        case updated(String)

        func apply(toState state: inout RootWorkflow.State) -> RootWorkflow.Output? {

            switch self {
            case .updated(let text):
                state.text = text
            }
            return nil
        }

    }
}


// MARK: Rendering

extension RootWorkflow {

    public typealias Rendering = ExampleScreen

    public func render(state: State, context: RenderContext<RootWorkflow>) -> Rendering {

        return PlaceholderScreen(
            text: state.text,
            onTextChanged: { text in
                sink.send(.updated(text))
            })
    }
}
```
Explanation of what's happening:
The text field's `textDidChange` action occurs, and the current text value is sent back to the workflow with a `sink.send(.updated(text))` call.

The `apply` method is called, and the workflow's state is updated with the updated text. The next render pass occurs, and the value of `state.text` is apply back to the UITextField.

Because the text input is happening rapidly, UIKit will send the next event to through the text field as soon as the main thread is available. This is easiest to reproduce synthetically with a KIF test, but using a hardware keyboard *can* produce this behavior (just less consistently).

Since the Signal backing the sink is no longer valid after a first action has been received (and it doesn't, nor shouldn't, batch the events), when the second time `sink.send(.updated(text))` is called, the action is dropped.

The text field's state now includes two keypresses (ie: "ab"), but the workflow has only gotten an updated text for the first (ie: "a"). When the render pass completes, the text field (being applied from the screen) will have `textField.text` be "a", and we've lost an entered character.

The most likely path forward for this is to ensure that it is synchronous for the next render pass to occur directly after handling an action (ie: use the UIScheduler in the `SubtreeManager`). This does *not* prevent this in the case where the UI update based on a new screen is happening asynchronously (deferred applying of the updated closure) - so there still could be cases where the UI implementation can cause dropped events.

davidapgar:
This guarantees that an action being applied from a sink
will immediately cause a render pass, providing a new sink to
the UI side, and prevents lost events.

Additionally, fixes that the SignalProducer from a worker
was not being cancelled on a subsequent render pass if it
was not declared.

Fixes #348

### DIFF
--- a/swift/Workflow/Sources/SubtreeManager.swift
+++ b/swift/Workflow/Sources/SubtreeManager.swift
@@ -1,3 +1,4 @@
+import Dispatch
 import ReactiveSwift
 import Result
 
@@ -9,28 +10,27 @@ extension WorkflowNode {
 
         internal var onUpdate: ((Output) -> Void)? = nil
 
-        /// Signals from the outside world (e.g. UI)
-        private let (childEvent, childEventObserver) = Signal<Signal<Output, NoError>, NoError>.pipe()
+        /// Sinks from the outside world (i.e. UI)
+        private var eventPipes: [EventPipe] = []
 
         /// The current array of children
         private (set) internal var childWorkflows: [ChildKey:AnyChildWorkflow] = [:]
 
+        /// The current array of workers
         private (set) internal var childWorkers: [AnyChildWorker] = []
 
-        private let (lifetime, token) = Lifetime.make()
+        /// Subscriptions from the outside world.
+        private var subscriptions: Subscriptions = Subscriptions(eventSources: [], eventPipe: EventPipe())
 
-        init() {
-            childEvent
-                .flatMap(.latest, { $0 })
-                .take(during: lifetime)
-                .observe(on: QueueScheduler.workflowExecution)
-                .observeValues { [weak self] event in
-                    self?.handle(output: event)
-                }
-        }
+        init() {}
 
         /// Performs an update pass using the given closure.
         func render<Rendering>(_ actions: (RenderContext<WorkflowType>) -> Rendering) -> Rendering {
+
+            /// Invalidate the previous action handlers.
+            for eventPipe in eventPipes {
+                eventPipe.invalidate()
+            }
 
             /// Create a workflow context containing the existing children
             let context = Context(
@@ -49,30 +49,44 @@ extension WorkflowNode {
             /// as a result of this call to `render`.
             self.childWorkflows = context.usedChildWorkflows
             self.childWorkers = context.usedChildWorkers
+            /// Merge all of the signals together from the subscriptions.
+            self.subscriptions = Subscriptions(eventSources: context.eventSources, eventPipe: EventPipe())
 
-            /// Finally, merge all of the events together from the new set of children and start listening to them
-            for child in self.childWorkflows {
-                child.value.onUpdate = { [weak self] output in
-                    self?.handle(output: output)
-                }
-            }
+            /// Capture all the pipes to be enabled after render completes.
+            self.eventPipes = context.eventPipes
+            self.eventPipes.append(self.subscriptions.eventPipe)
 
-            for worker in self.childWorkers {
-                worker.onUpdate = { [weak self] output in
-                    self?.handle(output: output)
-                }
-            }
-
-            let eventSources = Signal
-                .merge(context.eventSources)
-                .map({ update in
-                    return Output.update(update, source: .external)
-                })
-
-            childEventObserver.send(value: eventSources)
+            /// Set all event pipes to `pending`.
+            self.eventPipes.forEach { $0.setPending() }
 
             /// Return the rendered result
             return rendering
+        }
+
+        /// Enable the eventPipes for the previous rendering. The eventPipes are not valid until this has
+        /// be called. If is an error to call this twice without generating a new rendering.
+        func enableEvents() {
+            /// Check for queued events. If there are any, apply the first and yield to the next render loop.
+            let queuedEvents = self.eventPipes.compactMap { pipe in
+                pipe.pendingOutput()
+            }
+            if queuedEvents.count > 0 {
+                self.handle(output: queuedEvents[0])
+                return
+            }
+
+            /// Enable all action pipes.
+            for eventPipe in self.eventPipes {
+                eventPipe.enable { [weak self] output in
+                    self?.handle(output: output)
+                }
+            }
+
+            /// Enable all child workflows.
+            for child in self.childWorkflows {
+                child.value.enableEvents()
+            }
+
         }
 
         func makeDebugSnapshot() -> [WorkflowHierarchyDebugSnapshot.Child] {
@@ -109,6 +123,8 @@ extension WorkflowNode.SubtreeManager {
 
     /// The workflow context implementation used by the subtree manager.
     fileprivate final class Context: RenderContextType {
+
+        private (set) internal var eventPipes: [EventPipe]
         
         private let originalChildWorkflows: [ChildKey:AnyChildWorkflow]
         private (set) internal var usedChildWorkflows: [ChildKey:AnyChildWorkflow]
@@ -119,6 +135,8 @@ extension WorkflowNode.SubtreeManager {
         private (set) internal var eventSources: [Signal<AnyWorkflowAction<WorkflowType>, NoError>] = []
 
         internal init(originalChildWorkflows: [ChildKey:AnyChildWorkflow], originalChildWorkers: [AnyChildWorker]) {
+            self.eventPipes = []
+
             self.originalChildWorkflows = originalChildWorkflows
             self.usedChildWorkflows = [:]
 
@@ -138,6 +156,8 @@ extension WorkflowNode.SubtreeManager {
             }
 
             let child: ChildWorkflow<Child>
+            let eventPipe = EventPipe()
+            self.eventPipes.append(eventPipe)
 
             /// See if we can
             if let existing = originalChildWorkflows[childKey] {
@@ -151,20 +171,34 @@ extension WorkflowNode.SubtreeManager {
                 /// Update the existing child
                 existing.update(
                     workflow: workflow,
-                    outputMap: { AnyWorkflowAction(outputMap($0)) })
+                    outputMap: { AnyWorkflowAction(outputMap($0)) },
+                    eventPipe: eventPipe)
                 child = existing
             } else {
                 /// We could not find an existing child matching the given child key, so we will generate a new child.
                 /// This spins up a new workflow node, etc to host the newly created child.
                 child = ChildWorkflow<Child>(
                     workflow: workflow,
-                    outputMap: { AnyWorkflowAction(outputMap($0)) })
+                    outputMap: { AnyWorkflowAction(outputMap($0)) },
+                    eventPipe: eventPipe)
             }
 
             /// Store the resolved child in `used`. This allows us to a) hold on to any used children after this render
             /// pass, and b) ensure that we never allow the use of a given workflow type with identical keys.
             usedChildWorkflows[childKey] = child
             return child.render()
+        }
+
+        func makeSink<Action>(of actionType: Action.Type) -> Sink<Action> where Action : WorkflowAction, WorkflowType == Action.WorkflowType {
+
+            let eventPipe = EventPipe()
+
+            let sink = Sink<Action> { action in
+                let event = Output.update(AnyWorkflowAction(action), source: .external)
+                eventPipe.handle(event: event)
+            }
+            eventPipes.append(eventPipe)
+            return sink
         }
 
         func subscribe<Action>(signal: Signal<Action, NoError>) where Action : WorkflowAction, WorkflowType == Action.WorkflowType {
@@ -174,20 +208,91 @@ extension WorkflowNode.SubtreeManager {
         func awaitResult<W, Action>(for worker: W, outputMap: @escaping (W.Output) -> Action) where W : Worker, Action : WorkflowAction, WorkflowType == Action.WorkflowType {
 
             let outputMap = { AnyWorkflowAction(outputMap($0)) }
+            let eventPipe = EventPipe()
+            self.eventPipes.append(eventPipe)
 
             if let existingWorker = originalChildWorkers
                 .compactMap({ $0 as? ChildWorker<W> })
                 .first(where: { $0.worker.isEquivalent(to: worker) }) {
-                existingWorker.update(outputMap: outputMap)
+                existingWorker.update(outputMap: outputMap, eventPipe: eventPipe)
                 usedChildWorkers.append(existingWorker)
             } else {
-                let newChildWorker = ChildWorker(worker: worker, outputMap: outputMap)
+                let newChildWorker = ChildWorker(worker: worker, outputMap: outputMap, eventPipe: eventPipe)
                 usedChildWorkers.append(newChildWorker)
             }
         }
         
     }
 
+}
+
+
+extension WorkflowNode.SubtreeManager {
+    fileprivate final class EventPipe {
+
+        var validationState: ValidationState
+        enum ValidationState {
+            case preparing
+            case pending
+            case queued(Output)
+            case valid(handler: (Output) -> Void)
+            case invalid
+        }
+
+        init() {
+            self.validationState = .preparing
+        }
+
+        func handle(event: Output) {
+            if #available(iOS 10.0, OSX 10.12, *) {
+                dispatchPrecondition(condition: .onQueue(DispatchQueue.workflowExecution))
+            }
+
+            switch validationState {
+
+            case .preparing:
+                fatalError("Sink sent an action inside `render`. Sinks are not valid until `render` has completed.")
+
+            case .pending:
+                validationState = .queued(event)
+
+            case .queued:
+                fatalError("Action sent to pipe while already in the `queueing` state.")
+
+            case .valid(handler: let handler):
+                handler(event)
+
+            case .invalid:
+                fatalError("Sink sent an action after it was invalidated. Sinks can only be used for a single valid `Rendering`.")
+            }
+        }
+
+        func setPending() {
+            guard case .preparing = validationState else {
+                fatalError("Attempted to `setPending` an EventPipe that was not in the preparing state.")
+            }
+            validationState = .pending
+        }
+
+        func pendingOutput() -> Output? {
+            if case .queued(let output) = validationState {
+                return output
+            } else {
+                return nil
+            }
+        }
+
+        func enable(with handler: @escaping (Output) -> Void) {
+            guard case .pending = validationState else {
+                fatalError("EventPipe can only be enabled from the `pending` state")
+            }
+            validationState = .valid(handler: handler)
+        }
+
+        func invalidate() {
+            validationState = .invalid
+        }
+    }
 }
 
 extension WorkflowNode.SubtreeManager {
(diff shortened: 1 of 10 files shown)


swift/Workflow/Sources/SubtreeManager.swift
```diff
@@ -230,15 +290,16 @@ extension WorkflowNode.SubtreeManager {
...
 
         init(worker: W, outputMap: @escaping (W.Output) -> AnyWorkflowAction<WorkflowType>) {
             self.worker = worker
             self.signalProducer = worker.run()
             self.outputMap = outputMap
             super.init()
 
-            disposable = signalProducer
+            signalProducer
+                .take(during: lifetime)
```
davidapgar: This change has a worker get cancelled if it's not still declared on a subsequent render pass. See the test in `SubtreeManagerTests` for it.

(Alternatively, this could have been done by holding onto the `disposable`, and calling `disposable?.dispose()` in `deinit()`. I'm open to either approach if someone has strong feelings)

swift/Workflow/Sources/SubtreeManager.swift
```diff
@@ -9,29 +9,28 @@ extension WorkflowNode {
...
 
+        /// The current array of workers
         private (set) internal var childWorkers: [AnyChildWorker] = []
 
-        private let (lifetime, token) = Lifetime.make()
+        /// Subscriptions from the outside world.
+        private var subscriptionContext: SubscriptionContext = SubscriptionContext(eventSources: [])
 
-        init() {
-            childEvent
```
davidapgar: This logic is moved into the `SubscriptionContext`. After the sink changes, it was the only thing still using this (since `Workers` and `Workflows` call the handler function). Mostly refactored to keep it consistent.

swift/WorkflowTesting/Sources/WorkflowRenderTester.swift
```diff
@@ -151,6 +151,15 @@ fileprivate final class RenderTestContext<T: Workflow>: RenderContextType {
         return workflow.render(state: testContext.state, context: context)
     }
 
+    func makeSink<Action>(of actionType: Action.Type) -> Sink<Action> where Action : WorkflowAction, T == Action.WorkflowType {
+        let (signal, observer) = Signal<AnyWorkflowAction<WorkflowType>, NoError>.pipe()
```
davidapgar: For the render tester, it's the same implementation as before `makeSink` was a real function. However, all of this is synchronous in the tester, so it will behave appropriately.

I want to revisit the tester after this change anyway, since this should allow us expect/assert specific concrete Action types (which couldn't be done before, since they were wrapped in the type erasure `AnyWorkflowAction`). But will likely be a followup.

swift/Workflow/Sources/SubtreeManager.swift
```diff
@@ -190,6 +210,46 @@ extension WorkflowNode.SubtreeManager {
...
+            case valid(handler: (Output) -> Void)
+            case invalid
+        }
+
+        init() {
+            self.validationState = .preparing
+        }
+
+        func handle(action: AnyWorkflowAction<WorkflowType>) {
+            switch validationState {
```
davidapgar: This covers the *majority* of pre-use/after-use cases. There is still a gap of `sink.send` being call immediately after `render` completes (ie: in `screenDidChange`). It *does* fail, since the `Property` backing the rendering asserts that it's been re-entrantly updated, but it's a less than ideal failure message.

I'm open to thoughts on this. From the best of my exploration, it would have to be in `WorkflowHost` *after* updating the `rendering` property telling the `WorkflowNode` "you're ready", which seems like a clumsy API.

swift/Workflow/Sources/SubtreeManager.swift
```diff
@@ -190,6 +210,46 @@ extension WorkflowNode.SubtreeManager {
 
 }
 
+
+extension WorkflowNode.SubtreeManager {
+    fileprivate final class SinkContext {
```
timdonnelly: We use the `context` term in a bunch of places, I hope we can find a more descriptive word here. No brilliant suggestions from me at the moment, however.

swift/Workflow/Sources/SubtreeManager.swift
```diff
@@ -49,8 +48,16 @@ extension WorkflowNode {
             /// as a result of this call to `render`.
             self.childWorkflows = context.usedChildWorkflows
             self.childWorkers = context.usedChildWorkers
+            self.sinks = context.sinks
 
-            /// Finally, merge all of the events together from the new set of children and start listening to them
+            /// Enable and start listening to events from sinks, child workflows, and workers.
+            for sink in self.sinks {
```
timdonnelly: As far as I'm aware, `Sink`s are explicitly for use in UI. As such, their events should always come in on the main thread.

We should assert using

```
dispatchPrecondition(condition: .onQueue(.main))
```

either here or somewhere in the sink plumbing.

swift/Workflow/Sources/SubtreeManager.swift
timdonnelly: I actually don't think that's a bad idea at all.

When a tree update happens, renderings flow (down|up) to the root. So from the root we end up calling `render()`:

https://github.com/square/workflow/blob/02a87fc6c4507f7c007f5de0d95fa125c796e32e/swift/Workflow/Sources/WorkflowHost.swift#L56

We can immediately call `didRender()` after the render pass, which can propagate through the entire tree. Each node can then activate its event sinks as needed.

swift/Workflow/Sources/SubtreeManager.swift
davidapgar: `Sink`s *currently* are only used for UI events. I agree with adding a precondition on these, but want to leave open option of running the entirety of a workflow on a background thread (ie: not bound to a container/UI, just from a `WorkflowHost`).

I'll go with adding it for now, and then if/when we want background support, make the appropriate changes.

swift/Workflow/Sources/SubtreeManager.swift
davidapgar: +1. It's a bad name for this (since we already have a bunch of other things with this name). I'll come up with another one.

swift/Workflow/Sources/SubtreeManager.swift
```diff
@@ -9,29 +9,28 @@ extension WorkflowNode {
...
-                }
-        }
+        init() {}
 
         /// Performs an update pass using the given closure.
         func render<Rendering>(_ actions: (RenderContext<WorkflowType>) -> Rendering) -> Rendering {
 
+            /// Invalidate the previous sinks.
+            for sink in sinks {
+                sink.invalidate()
```
zach-klippenstein: What does invalidating a sink do?

swift/Workflow/Sources/SubtreeManager.swift
zach-klippenstein: That sounds very similar to the way it works in Kotlin. Each iteration of the main loop consists of three(-ish) phases:
1. **Render pass** (non-blocking) – constructs the context for the current iteration, subscribes channels to every new signal/event sink, tears down obsolete ones.
    1.5. **Snapshot** (non-blocking, separate pass than rendering but happens immediately after) – constructs snapshot of the current state tree.
2. **Publish** (possibly blocking if backpressure) – send rendering, snapshot, and optionally the last output to the consumer of the workflow runtime.
3. **"Tick"** (usually blocking, bad name) – selects on all channels and waits for the next "thing" to happen. Only allows one atomic "thing" to happen at once, runs all `WorkflowAction`s all the way up the tree, and returns the root workflow's optional `OutputT` for that iteration (which is then emitted in the subsequent iteration's "Publish" pass).

swift/Workflow/Sources/SubtreeManager.swift
davidapgar: Just sets its internal state to "invalid". `sink` in this scope is the (poorly named currently) `SinkContext` that is actually wired into the subtree manager.

If it receives an event when it's in the `invalid` state, it will assert (fail fast) instead of silently dropping it.

zach-klippenstein: Ah, ok, so this is basically the equivalent of this line in a PR i've got open:

https://github.com/square/workflow/blob/798afacb46481a8998deff2c3b47784609fd0de4/kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/WorkflowNode.kt#L125

swift/Workflow/Sources/SubtreeManager.swift
davidapgar: That's good insight, and gives me confidence this is an appropriate extra step to add.

The behavior will/should be (the differentiation is the most of this will by synchronous):
**Render pass**
1) **Prepare** (what you listed as render pass): Construct the context and call `render` on the workflow.
  - Invalidate old sinks, remove obsolete subscriptions, child workflows, workers (that are no longer declared).
  - Gather all the sinks, subscriptions, etc from render, in a "pending" state.
2) **Publish** (still synchronous) - Send the rendering and optional output.
3) **Enable** (synchronous as well) - Enable all of the sinks, subscriptions, workers, etc. The subscriptions and workers are explicitly subscribed on a queue that guarantees that they will not emit synchronously, so for them having a separate "enable" isn't technically necessary, but fits the model nicely. This will walk the tree of workflows again to enable the root and all the children/grandchildren in a synchronous pass.

swift/Workflow/Sources/SubtreeManager.swift
davidapgar: Yup! I assume that `cancel` in that context will make it assert if it's still used?

swift/Workflow/Sources/SubtreeManager.swift
rjrjr: Rather than hard coding main, can you make the queue configurable, with .main as a default value?

swift/Workflow/Sources/SubtreeManager.swift
rjrjr: hyper meta nit: even though these are "internal" concerns, it'd be cool if we used the same naming for these passes in both code bases. That would let us better describe the contract in shared docs.

How does this sound: Dave comes up with good names in this PR, and Kotlin catches up once that mini-bikeshed settles.

zach-klippenstein: The names don't exist in code. "render" is the same in both already I think, and that's really the only part they share.

swift/Workflow/Sources/SubtreeManager.swift
zach-klippenstein: In Rx, we tend to use the latter pattern if we need to re-subscribe later. We add disposables to a `CompositeDisposable` and clear it if it's a one-shot thing.

swift/Workflow/Sources/SubtreeManager.swift
zach-klippenstein: Yep.

swift/Workflow/Sources/SubtreeManager.swift
davidapgar: I'm debating where the best place to document this is. I feel like at least initially I'm going to write up this behavior as doc comments in the code (so that it's accurate). Overall, I have been kicking around the idea of writing (ie: have started and stopped a few times) a "implementation" or "lifecycle" or "internals" doc that codifies all of these details.

Also debated adding an "RFC" section to the docs, specifically talking about issues like this (so they're more discoverable than just github issues), but don't have a strong preference yet.

rjrjr: I actually think this would be good method documentation for render(). render is the most central method in the library, and this describes the precise contract of when the method is called, what it builds, and what is allowed to happen to the thing that it builds.

swift/Workflow/Sources/SubtreeManager.swift
davidapgar: That's likely the long term path for this (making the queue configurable). Since we are only using this on the main queue currently, I'm going to leave it as is for the moment (and do that in a followup).

https://github.com/square/workflow/issues/364

swift/Workflow/Sources/SubtreeManager.swift
zach-klippenstein: > I have been kicking around the idea of writing (ie: have started and stopped a few times) a "implementation" or "lifecycle" or "internals" doc that codifies all of these details.

I had a section in the docs rewrite outline for this, but Ray suggested using it for something else: https://docs.google.com/document/d/1hlJZh72ZQ-Abd9XZRm4MjsULWQE2pCDTHy-WrFA_XiE/edit#heading=h.4l7zcgaeyx6f

davidapgar: Updated, adding an `enableRendering` method that must be called before any actions are valid to be handled.

swift/Workflow/Sources/SubtreeManager.swift
davidapgar: Ended up with `ActionPipe` as it's the pipe on the sink that actions flow down.

I'm still not 100% happy with it, but it's at least not a collision with "context".

`ActionDrain`? `ActionHandler`?

timdonnelly: 😂 I almost suggested pipe.

swift/Workflow/Sources/SubtreeManager.swift
davidapgar: Added verification we are on the `DispatchQueue.workflowExectution` for all actions that come in (covering UI events, as well as asserting that's what we subscribe on for workers and subscriptions)

timdonnelly: We actually have the workflow queue defined in an extension somewhere, you'd want to assert that you're on that queue versus main (the current definition just resolves to main).

swift/Workflow/Sources/SubtreeManager.swift
```diff
@@ -49,30 +48,32 @@ extension WorkflowNode {
...
 
-            for worker in self.childWorkers {
-                worker.onUpdate = { [weak self] output in
+            /// Return the rendered result
+            return rendering
+        }
+
+        /// Enable the actionPipes for the previous rendering. The actionPipes are not valid until this has
+        /// be called. If is an error to call this twice without generating a new rendering.
+        func enableRendering() {
```
davidapgar: This is now the second pass on a render loop. All event sources go through the `ActionPipe`s, so it's a singular path that is synchronous to validate we are on the correct dispatch queue, as well as having been enabled.

swift/Workflow/Sources/SubtreeManager.swift
davidapgar: Yup, that's the one I used.

swift/Workflow/Sources/SubtreeManager.swift
timdonnelly: supernit: not a fan of `enableRendering` – it doesn't actually *do* anything to the rendering. We're really enabling any captured event sinks.

You could either name it based on what it does: `enableEvents()`, etc.

or name it based on the phase in the render cycle

`didEmitRendering()`

davidapgar: Updated to `enableEvents` - you are right, that's a much better name.

swift/Workflow/Sources/SubtreeManager.swift
davidapgar: I think this feels appropriate for ReactiveSwift, and the lifetimes (if it was RxSwift, there would be a "disposeBag" (neat name) that would also get rid of these as the disposable when the object gets deallocated).  Since we are never unsubscribing from these except with the object goes away, I'm not going to add the extra code to get rid of it in `deinit`

swift/Workflow/Sources/WorkflowNode.swift
```diff
@@ -2,7 +2,7 @@ import ReactiveSwift
 import Result
 
 /// Manages a running workflow.
-final class WorkflowNode<WorkflowType: Workflow> {
+final internal class WorkflowNode<WorkflowType: Workflow> {
```
timdonnelly: `internal` is implicit, generally prefer leaving it out (or, for consistency, add it everywhere... but that doesn't sound great)

davidapgar: Took it back off. It was mostly that I was walking through all these types, it was a helpful signal to me of what is and isn't visible.

I agree with this overall (leaving it off, since it is implicit. Adding it everywhere does seem like it too much noise)

swift/Workflow/Sources/WorkflowHost.swift
```diff
@@ -59,6 +60,8 @@ public final class WorkflowHost<WorkflowType: Workflow> {
             outputEventObserver.send(value: outputEvent)
         }
 
+        rootNode.enableEvents()
```
timdonnelly: It seems like this will get a bit complicated in the case where a rendering causes an event. This call to `enableEvents` will causes something down in the tree to dequeue its pending event, which will bubble up and recursively call `self.handle(output:)`.

That seems like it will (at the very least) result in out of order debug info (the recursive invocation will call `debugger?.didUpdate` before the outer one)

timdonnelly: I don't have all of this architecture in my head, but it might be nice if `enableEvents` could optionally return a `WorkflowNode<WorkflowType>.Output` value, then you could predictable handle the update inline here.

davidapgar: Yes - I agree that we need to handle the debug info appropriately (so it's in the "correct" order of what happened).

I *think* reordering in `WorkflowHost` to emit the debug event before calling `enableEvents` will have the correct sequence. Basically, `enableEvents` needs to be the last thing in `handle` called, so that it is properly "tail recursive" to guarantee the right order. I'll add a test case around this, to ensure whatever is done behave correctly.

davidapgar: Updated `WorkflowHost` and added a test for this. By having `enableEvents` be the last thing called in `handle` it's now ensured that the debug info comes through in the correct order.

RoyalPineapple:
LGTM

davidapgar:
@timdonnelly All updated. This now covers all the cases (including events happening between `render` being called, and `enableEvents`), as well as ensuring the debug info is in order.

Validated against all of our internal code using this - please take a look, I'd like to get this landed so we can update.

PR #365, opened 2019-05-23
Updates CHANGELOG with PR links

lucamtudor:


zach-klippenstein:
These are linked correctly on the [releases](https://github.com/square/workflow/releases) page, which we are now updating for all releases as well.

PR #366, opened 2019-05-23
Add a Worker.createSideEffect helper for Nothing-type workers.

zach-klippenstein:


kotlin/workflow-core/src/main/java/com/squareup/workflow/Worker.kt
```diff
@@ -178,12 +178,33 @@ interface Worker<out T> {
      *   emitOutput(3)
      * }
      * ```
+     *
+     * Note: If your worker just need to perform side effects and doesn't need to emit anything,
```
rjrjr: needs

kotlin/workflow-core/src/main/java/com/squareup/workflow/Worker.kt
```diff
@@ -178,12 +178,33 @@ interface Worker<out T> {
...
+     * the same key.
+     *
+     * E.g.:
+     * ```
+     * fun logOnEntered(message: String) = Worker.createSideEffect("logOnEntered") {
+     *   println("Entered state: $message")
+     * }
+     * ```
+     */
+    fun createSideEffect(
```
rjrjr: Should add a test.

zach-klippenstein: For what? There's no logic here it's just a convenience wrapper.

rjrjr: So that we'll know if we manage to break the wrapper, and know that the block is actually called.

PR #368, opened 2019-05-24
Add a Worker.timer function to create simple delay workers.

zach-klippenstein:


kotlin/workflow-core/src/main/java/com/squareup/workflow/Worker.kt
```diff
@@ -381,3 +393,16 @@ internal class TypedWorker<T>(
 
   override fun toString(): String = "TypedWorker(key=\"$key\", type=$type)"
 }
+
+private class TimerWorker(
+  private val delayMs: Long,
```
bencochran: On iOS, the convention is to use floating-point number of seconds, which our implementation should follow. Probably fine to have differences on this, but wanted to note it in case others feel differently.

rjrjr: Agree that this is a reasonable difference. A `Long` of ms is definitely idiomatic for java / kotlin.

kotlin/workflow-testing/src/test/java/com/squareup/workflow/WorkerTest.kt
```diff
@@ -286,4 +287,37 @@ class WorkerTest {
...
+    val worker2 = Worker.timer(1, "key2")
+
+    assertFalse(worker1.doesSameWorkAs(worker2))
+  }
+
+  @Test fun `timer emits and finishes after delay`() {
+    val worker = Worker.timer(1000)
+    val testDispatcher = TestCoroutineDispatcher()
+
+    worker.test(context = testDispatcher) {
```
kirillzh: Neat!

kotlin/workflow-testing/src/test/java/com/squareup/workflow/WorkerTest.kt
```diff
@@ -286,4 +287,37 @@ class WorkerTest {
...
+      assertNoOutput()
+      assertNotFinished()
+
+      testDispatcher.advanceTimeBy(999)
```
kirillzh: Neat!

PR #369, opened 2019-05-24
Make RenderTester not require specific Mock* workflows/workers.

zach-klippenstein:
`RenderTester` currently requires all child workflows and workers that the workflow-under-test delegates to to be instances of `MockChildWorkflow` or `MockWorker`, respectively. This makes tests super verbose because if the actual type of a child workflow is a subinterface of `Workflow`, you need to actually implement that interface and can only use the mocks to do implementation-by-delegation, forcing you to repeat all the generics. It's extremely verbose. By removing this restriction, we can use mocking libraries to generate basic implementations of interfaces that should still work with the tester because all the assertion methods actually talk to the test context.

@rjrjr I want to land this in 0.16.1 as well, because I want to be able to use `RenderTester`.

### DIFF
--- a/kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RenderTester.kt
+++ b/kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RenderTester.kt
@@ -216,7 +216,7 @@ class TestRenderResult<StateT, OutputT : Any, RenderingT> internal constructor(
         ?: throw AssertionError("Expected workflow to be rendered: $workflow (key=\"$key\")")
   }
 
-  internal fun <T> findWorkerCase(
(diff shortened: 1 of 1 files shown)


rjrjr:
Do you want to loosen this up?

```
  override fun doesSameWorkAs(otherWorker: Worker<*>): Boolean =
    otherWorker is MockWorker && name == otherWorker.name
```

zach-klippenstein:
> Do you want to loosen this up?
> 
> ```
>   override fun doesSameWorkAs(otherWorker: Worker<*>): Boolean =
>     otherWorker is MockWorker && name == otherWorker.name
> ```

For what reason? Mocks will still need to be explicit about equivalence, if they're not using this class.

PR #373, opened 2019-05-26
Added a sample terminal app and mini-library.

zach-klippenstein:


zach-klippenstein:
TODO:

- [ ] Kdoc

kotlin/samples/hello-terminal/app/src/main/java/com/squareup/sample/helloterminal/Main.kt
```diff
@@ -0,0 +1,30 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.sample.helloterminal
+
+import com.squareup.sample.helloterminal.terminalworkflow.TerminalWorkflowRunner
+
+data class State(
```
zach-klippenstein: Move this to HelloWorkflow

kotlin/samples/hello-terminal/app/src/main/java/com/squareup/sample/helloterminal/HelloWorkflow.kt
```diff
@@ -0,0 +1,78 @@
...
+import com.squareup.sample.helloterminal.terminalworkflow.TerminalRendering
+import com.squareup.workflow.RenderContext
+import com.squareup.workflow.Snapshot
+import com.squareup.workflow.StatefulWorkflow
+import com.squareup.workflow.WorkflowAction.Companion.emitOutput
+import com.squareup.workflow.WorkflowAction.Companion.enterState
+import com.squareup.workflow.WorkflowAction.Companion.noop
+import com.squareup.workflow.renderChild
+
+class HelloWorkflow : StatefulWorkflow<TerminalConfig, State, ExitCode, TerminalRendering>() {
```
rjrjr: Any name but this. See `typealias` comment on `TerminalWorkflow`.

kotlin/samples/hello-terminal/terminalworkflow/src/main/java/com/squareup/sample/helloterminal/terminalworkflow/KeyStroke.kt
```diff
@@ -0,0 +1,65 @@
...
+import com.squareup.sample.helloterminal.terminalworkflow.KeyStroke.KeyType.Unknown
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.channels.ReceiveChannel
+import kotlinx.coroutines.channels.produce
+import kotlinx.coroutines.isActive
+import com.googlecode.lanterna.input.KeyStroke as LanternaKeystroke
+
+/**
+ * TODO kdoc
```
rjrjr: TODO

kotlin/samples/hello-terminal/terminalworkflow/src/main/java/com/squareup/sample/helloterminal/terminalworkflow/TerminalConfig.kt
```diff
@@ -0,0 +1,51 @@
...
+import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.channels.ReceiveChannel
+import kotlinx.coroutines.channels.produce
+import kotlinx.coroutines.suspendCancellableCoroutine
+
+/**
+ * TODO kdoc
+ */
+data class TerminalConfig(
+  val size: Size
```
rjrjr: YAGNI. Until we have more properties than just size, the nesting is just obfuscating. I'd suggest

```
data class TerminalSize(
  val rows: Int,
  val columns: Int
)
```

kotlin/samples/hello-terminal/terminalworkflow/src/main/java/com/squareup/sample/helloterminal/terminalworkflow/TerminalWorkflow.kt
```diff
@@ -0,0 +1,25 @@
...
+package com.squareup.sample.helloterminal.terminalworkflow
+
+import com.squareup.workflow.Workflow
+
+typealias ExitCode = Int
+
+/**
+ * TODO write documentation.
+ */
+typealias TerminalWorkflow = Workflow<TerminalConfig, ExitCode, TerminalRendering>
```
rjrjr: Is the use of `typealias` really a good example? We've been using interfaces internally, makes it much easier to see what's going on and so far I've seen no penalties. 

So perhaps your `HelloWorkflow` could instead be `TerminalWorkflowImp : TerminalWorkflow, StatefulWorkflow<...>`?

kotlin/samples/hello-terminal/terminalworkflow/src/main/java/com/squareup/sample/helloterminal/terminalworkflow/TerminalWorkflowRunner.kt
```diff
@@ -0,0 +1,156 @@
...
+ */
+class TerminalWorkflowRunner(
+  private val hostFactory: WorkflowHost.Factory = WorkflowHost.Factory(EmptyCoroutineContext),
+  private val ioDispatcher: CoroutineDispatcher = Dispatchers.IO
+) {
+
+  private val screen = DefaultTerminalFactory().createScreen()
+
+  @UseExperimental(ExperimentalCoroutinesApi::class, ObsoleteCoroutinesApi::class)
+  @Suppress("BlockingMethodInNonBlockingContext")
```
rjrjr: Document why the suppress is okay.

kotlin/samples/hello-terminal/app/README.md
```diff
@@ -0,0 +1,7 @@
+# hello-terminal / app
```
rjrjr: More depth explaining what's going on here, in particular where lanterna ends and the sample code begins.

rjrjr: Rather than two README files in `app` and the library, probably easier to find a single one up in `hello-terminal`.

kotlin/samples/hello-terminal/terminalworkflow/src/main/java/com/squareup/sample/helloterminal/terminalworkflow/TerminalWorkflowRunner.kt
```diff
@@ -0,0 +1,156 @@
...
+import kotlinx.coroutines.channels.SendChannel
+import kotlinx.coroutines.coroutineScope
+import kotlinx.coroutines.selects.selectUnbiased
+import kotlinx.coroutines.withContext
+import kotlin.coroutines.EmptyCoroutineContext
+
+/**
+ * TODO write documentation.
+ */
+class TerminalWorkflowRunner(
```
rjrjr: Is it worth trying to tease out a `CliWorkflowRunner` or maybe `LanternaWorkflowRunner` that's separate from `TerminalWorkflow` itself? Still as sample code, obv. way premature to try to pull it into a library.

kotlin/samples/hello-terminal/terminalworkflow/src/main/java/com/squareup/sample/helloterminal/terminalworkflow/TerminalWorkflowRunner.kt
```diff
@@ -0,0 +1,156 @@
...
+  // Need to send an initial input for the workflow to start running.
+  configs.offer(config)
+
+  while (true) {
+    val update = selectUnbiased<Update<ExitCode, TerminalRendering>> {
+      keyStrokes.onReceive { key ->
+        println("2: $key")
+        lastRendering?.onKeyStroke?.invoke(key.toKeyStroke())
+
+        // Sending that event invalidated the lastRendering, so we don't want to re-iterate
```
rjrjr: Good place to over-explain -- _Workflow renderings are able to accept one event / input change only, so sending that last event..._

kotlin/samples/hello-terminal/app/src/main/java/com/squareup/sample/helloterminal/Main.kt
rjrjr: ??

kotlin/samples/hello-terminal/terminalworkflow/src/main/java/com/squareup/sample/helloterminal/terminalworkflow/TerminalWorkflow.kt
zach-klippenstein: The only reason we're using interfaces internally is because Dagger can't handle kotlin.

zach-klippenstein:
I found a workflow bug. Occasionally keystrokes get lost. They're being seen by all the event handlers, so I'm guessing they are getting lost in the actual workflow runtime somewhere.
Maybe this is the "event handler is stale but hasn't been closed" issue that I have
a branch open to throw on. Or maybe the event is getting delivered to the event channel,
but then something else is beating it out so the channel is never actually received from.

zach-klippenstein:
So i can’t believe this worked before, which it did only by accident. I am wrapping my main workflow in another one and trying to forward events. Before, since the event handler ran before being sent to the channel, the forward worked because it triggered the leaf event handler to complete first. If i move the event handler code to only run when the event actually wins the selection, the event forward will always fail because all event handlers from that rendering are now invalid.

kotlin/samples/hello-terminal/terminalworkflow/src/main/java/com/squareup/sample/helloterminal/terminalworkflow/TerminalWorkflow.kt
rjrjr: IIRC that's why we started the conversation that lead to the introduction of the trimmed down Workflow interface. Now that we have it using typealias isn't solving any problems, and I'd argue makess the code harder to follow.

kotlin/samples/hello-terminal/terminalworkflow/src/main/java/com/squareup/sample/helloterminal/terminalworkflow/TerminalWorkflowRunner.kt
zach-klippenstein: That feels over-engineered/YAGNI, especially for a sample. Most of the code here is very specific to Lanterna's API.

kotlin/samples/hello-terminal/app/README.md
zach-klippenstein: Fleshed out the docs a little, lemme know what you think.

kotlin/samples/hello-terminal/terminalworkflow/src/main/java/com/squareup/sample/helloterminal/terminalworkflow/TerminalWorkflow.kt
zach-klippenstein: I disagree with most of your points, but I don't care enough to continue fighting on this.

kotlin/samples/hello-terminal/app/src/main/java/com/squareup/sample/helloterminal/HelloWorkflow.kt
zach-klippenstein: But we already have

https://github.com/square/workflow/blob/79ac0fb418c18c8894a54d8faf3376efa5d0596d/kotlin/samples/hello-workflow/src/main/java/com/squareup/sample/helloworkflow/HelloWorkflow.kt#L13

kotlin/samples/hello-terminal/README.md
```diff
@@ -0,0 +1,27 @@
...
+The API entry points are `TerminalWorkflow` and `TerminalWorkflowRunner`. Create one of each of
+those, and pass your workflow to `TerminalWorkflowRunner.run()`.
+
+Your `TerminalWorkflow` will get the current size of the terminal as `InputT`, render a
+`TerminalRendering` that includes the text to display, foreground and background colors, and an
+event handler to receive keystroke events. It can also exit the process by emitting an integer
+exit code.
+
+This module delegates to the third-party library [Lanterna](https://github.com/mabe02/lanterna)
+to do the actual hard work of talking to the system terminal. Lanterna also supports running a fake
```
rjrjr: Is the Swing trick something we can do w/this code? If so, can we document the gradle command? If not, maybe snip this sentence as a distraction.

kotlin/samples/hello-terminal/hello-terminal-app/src/main/java/com/squareup/sample/helloterminal/HelloWorkflow.kt
```diff
@@ -0,0 +1,81 @@
...
+import com.squareup.sample.helloterminal.terminalworkflow.TerminalWorkflow
+import com.squareup.workflow.RenderContext
+import com.squareup.workflow.Snapshot
+import com.squareup.workflow.StatefulWorkflow
+import com.squareup.workflow.WorkflowAction.Companion.emitOutput
+import com.squareup.workflow.WorkflowAction.Companion.enterState
+import com.squareup.workflow.WorkflowAction.Companion.noop
+import com.squareup.workflow.renderChild
+
+class HelloWorkflow : TerminalWorkflow,
```
rjrjr: Still don't think this class should be called `HelloWorkflow`. I doubled-up that name in the other two samples b/c it's copy/paste identical, and really is a Hello World type of thing. I'd rather that's what gets found when people search for `HelloWorkflow`. 

`HelloTerminalWorkflow`? `SimpleTerminalWorkflow`? `TerminalWorkflowImpl`? `RealTerminalWorkflow`?

kotlin/samples/hello-terminal/app/src/main/java/com/squareup/sample/helloterminal/HelloWorkflow.kt
rjrjr: That's kind of my point. Tried to get that across…somewhere. God, I hate this disappearing and re-appearing reply UI.

kotlin/samples/hello-terminal/README.md
zach-klippenstein: It happens automatically when you run, if the library doesn't detect a terminal connected to the process. I'll delete the sentence.

kotlin/samples/hello-terminal/hello-terminal-app/src/main/java/com/squareup/sample/helloterminal/HelloWorkflow.kt
zach-klippenstein: `HelloTerminalWorkflow` it is.

zach-klippenstein:
For the record, filed that dropped event bug as #374.

PR #375, opened 2019-05-29
Allow specifying the CoroutineDispatcher used by WorkflowRunnerViewModel

Blisse:
This allows customization of the ViewModel's dispatcher, since Dispatchers.Main isn't available on Buck builds.

### DIFF
--- a/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowRunnerViewModel.kt
+++ b/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowRunnerViewModel.kt
@@ -25,6 +25,7 @@ import com.squareup.workflow.rx2.flatMapWorkflow
 import io.reactivex.Flowable
 import io.reactivex.Observable
 import io.reactivex.disposables.Disposable
+import kotlinx.coroutines.CoroutineDispatcher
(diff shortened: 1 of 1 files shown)


kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowRunnerViewModel.kt
```diff
@@ -38,14 +39,15 @@ internal class WorkflowRunnerViewModel<OutputT : Any>(
     private val workflow: Workflow<InputT, OutputT, Any>,
     private val viewRegistry: ViewRegistry,
     private val inputs: Flowable<InputT>,
-    savedInstanceState: Bundle?
+    savedInstanceState: Bundle?,
+    private val dispatcher: CoroutineDispatcher = Dispatchers.Main.immediate
```
zach-klippenstein: This parameter should be a `CoroutineContext` so it can play nice with scoped concurrency, coroutine names for troubleshooting, etc.

zach-klippenstein: Oh, flatMapWorkflow is dumb. Need to delete…

PR #378, opened 2019-05-30
Remove (incorrect) manual implementation of ensureActive.

zach-klippenstein:


kotlin/workflow-runtime/src/main/java/com/squareup/workflow/WorkflowHost.kt
```diff
@@ -25,11 +25,10 @@ import kotlinx.coroutines.CoroutineName
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.InternalCoroutinesApi
-import kotlinx.coroutines.Job
-import kotlinx.coroutines.NonCancellable.isActive
```
zach-klippenstein: This is the wrong import, so this code was never working correctly 😬

PR #383, opened 2019-05-31
Hello `ViewRunner`, goodbye `Coordinator`, `BackStackEffect`.

### LINKED ISSUE #235: Consider removing Rx from ViewRegistry public API.
Swift's view system puts an imperative update method on view controllers. Once we have inline coordinators (#234), consider doing the same.

AC: Coordinator implementers are not forced to think in terms of streams, nor to manage subscriptions.

rjrjr:
Closes #235.

So much simpler now.

`ViewRegistry.getBinding(type:String): ViewBinding` is replaced with
`<T> ViewRegistry.buildView(rendering: RenderingT): View`. And because we no longer create views
before we have a rendering for them to show, the initial view tree is created
synchronously -- less chance for flicker, and simpler code, especially around view
persistence.

That `ViewRegistry.buildView` method promises to return a view that can respond to the
new `View.showRendering` extension. Set up that and `View.canShowRendering` by
calling `View.bindShowRendering`. (Although usually `LayoutRunner.binding` will do that
for you, see below.)

A `ViewRegistry` is still basically a collection of `ViewBinding` instances. But
`ViewBinding` is no longer in the `Scene` business, and `BackstackEffect` is
gone. Nav transitions aren't interesting enough to require that kind of
configuration, and it's just not that hard for a `View.showRendering` method to
do it's own fancy-schmancy morphing without any help from us.

For  working with layout files, `LayoutRunner` replaces `Coordinator` as a public
concern. No more reliance on `Observable<Screen>`, and so no subscriptions to manage.
Instead, just implement a single `showRendering` method.

Bonus: Alert dialogs can be updated in place now.


kotlin/samples/hello-workflow-fragment/src/main/java/com/squareup/sample/helloworkflowfragment/HelloFragmentViewRunner.kt
```diff
@@ -18,34 +18,30 @@ package com.squareup.sample.helloworkflowfragment
...
+    view: View,
+    registry: ViewRegistry
+  ) {
+    messageView = view.findViewById(R.id.hello_message)
   }
 
-  override fun detach(view: View) {
-    subs.clear()
+  @SuppressLint("SetTextI18n")
+  override fun update(newValue: HelloWorkflow.Rendering) {
```
Blisse: Something I really like with the single attach(view) is that technically you can be much "purer" by not having to specify the lateinit vars as class members, and just assign the findViewByIds to local variables.

zach-klippenstein: Weren't you considering that trick Helios did where `update` shares the lexical scope of `bind`?

kotlin/samples/hello-workflow/src/main/java/com/squareup/sample/helloworkflow/HelloWorkflow.kt
```diff
@@ -29,24 +29,17 @@ object HelloWorkflow : StatefulWorkflow<Unit, State, Unit, Rendering>() {
...
     context: RenderContext<State, Unit>
-  ): Rendering {
-    return Rendering(
-        message = state.name,
-        onClick = context.onEvent { Companion.enterState(state.theOtherState()) }
-    )
-  }
+  ): Rendering = Rendering(
+      message = state.name,
+      onClick = context.onEvent { Companion.enterState(state.theOtherState()) }
```
zach-klippenstein: Nit: static import `enterState`

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/backstack/BackStackContainer.kt
```diff
@@ -51,45 +60,45 @@ class BackStackContainer(
...
-                  setUpNewView = updateTools::setUpNewView,
-                  direction = updateTools.direction
-              )
+        ?.let { oldView ->
+          updateTools.saveOldView(oldView)
+
+          val newScene = Scene(this, newView)
+
+          val (outEdge, inEdge) = when (PUSH) {
+            updateTools.direction -> Pair(Gravity.START, Gravity.END)
```
zach-klippenstein: This looks unidiomatic to me, usually the variable is in the when and the cases are constant - is something going on here I'm not seeing?

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/backstack/BackStackContainer.kt
```diff
@@ -51,45 +60,45 @@ class BackStackContainer(
...
+            else -> Pair(Gravity.END, Gravity.START)
```
zach-klippenstein: This might actually be a rare case where the infix `to` function could actually be readable.

kotlin/samples/hello-workflow-fragment/src/main/java/com/squareup/sample/helloworkflowfragment/HelloFragmentViewRunner.kt
rjrjr: @zach-klippenstein I'm not sure what your'e referring to, I'll ask him.

@Blisse Yeah. :/ While you can theoretically be purer, it still requires you to play games if you want to avoid doing a lot of redundant findById calls on every update, and last I looked those are really terrible. I liked how this approach nudges you into one set up / lots of updates.

What would you think of ViewRunner being an abstract class rather than an interface, with a constructor that requires a View? That would be a lot closer to what happens on iOS, actually, where they have a subclass of `ViewController` that you must extend. Something like:

```
abstract class ViewRunner<ViewT : View, ValueT>(protected val view: ViewT) {
  abstract fun display(value: ValueT)
}
```

rjrjr: Taking this further, the `display` method could be from an interface, and the abstract class used only by `LayoutBinding`. That way we avoid putting awkward restrictions on code construction systems like Helios is working on.

```
interface ViewRunner<ValueT> {
  fun display(value: ValueT)
}
```

```
abstract class LayoutRunner<ViewT: View, ValueT>(
  protected val view: ViewT
) : ViewRunner<ValueT> {
}
```

(Also should stop forcing `ValueT` to extend `Any`.)

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ModalViewContainer.kt
```diff
@@ -40,40 +39,59 @@ internal class ModalViewContainer
...
+            // If we don't set or clear the background drawable, the window cannot go full bleed.
+            window!!.setBackgroundDrawable(null)
+          }
+        }
+        .run {
+          DialogRef(initialValue, this, view)
+        }
+  }
+
+  override fun updateDialog(dialogRef: DialogRef<Any>): DialogRef<Any> {
```
rjrjr: Just realized that I'm not actually using that return type. YAGNI, and I can imagine subclasses getting into a lot of confusion or trouble by making edits that I'm not set up to honor.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/backstack/BackStackContainer.kt
rjrjr: Whoa,  that ain't right.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/backstack/BackStackContainer.kt
rjrjr:
```
          val (outEdge, inEdge) = when (updateTools.direction) {
            PUSH -> Gravity.START to Gravity.END
            POP -> Gravity.END to Gravity.START
          }
```

😗 👌

kotlin/samples/hello-workflow-fragment/src/main/java/com/squareup/sample/helloworkflowfragment/HelloFragmentViewRunner.kt
rjrjr: Oh, has to extend Any if I want to use the type info the way I am. Fine with me.

zach-klippenstein: I mean this sort of thing:
```kotlin
bind { view ->
  val myView = view.findViewById(R.id.my_view)
  update { data ->
    myView.text = data.text
  }
}
```

rjrjr: This is shaping up nicely (more functional and less brittle than I sketched), and it gets rid of all those anonymous runner classes.

```
class AuthorizingViewRunner(
  view: View,
  registry: ViewRegistry
) : LayoutRunner<AuthorizingScreen>(view, registry) {
  private val messageView: TextView = view.findViewById(R.id.authorizing_message)

  override fun showValue(value: AuthorizingScreen) {
    messageView.text = value.message
  }

  companion object : ViewBinding<AuthorizingScreen> by LayoutBinding.of(
      R.layout.authorizing_layout, ::AuthorizingViewRunner
  )
}
```

zach-klippenstein: Having to manually plumb the `ViewRegistry` through is unfortunate, seems like boilerplate we should get rid of somehow.

rjrjr: Oh yeah, derp. Totally fixable.

rjrjr:
```
class HelloViewRunner(view: View) : LayoutRunner<Rendering>(view) {
  private val messageView: TextView = view.findViewById(R.id.hello_message)

  override fun showValue(value: Rendering) {
    messageView.text = value.message
    messageView.setOnClickListener { value.onClick(Unit) }
  }

  companion object : ViewBinding<Rendering> by binding(
      R.layout.hello_goodbye_layout, ::HelloViewRunner
  )
}
```

rjrjr: At this point the base class `LayoutRunner` is kind of pointless, even though it's a nice blank to fill in. And the `ViewRunner` interface has already evaporated. Let me see if I can tighten this up some more.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowLayout.kt
```diff
@@ -124,4 +124,29 @@ class WorkflowLayout(
...
   }
+
+  /**
+   * Subscribes [update] to [source] only while this [View] is attached to a window.
+   */
+  private fun <S : Any> View.takeWhileAttached(
+    source: Observable<S>,
+    update: (S) -> Unit
+  ) {
+    Coordinators.bind(this) {
```
rjrjr: Last Coordinator dependency. Should follow up to eliminate it.

rjrjr: Tracked in #384

rjrjr:
Update snips the last couple of kdoc references to `Coordinator`.

rjrjr:
Update reduces `Coordinator` gradle dep from `api` to `implementation`.

zach-klippenstein:
> Alert dialogs can be updated in place now.

Woah, missed this on the first pass. This seems like a huge win!

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ModalContainer.kt
```diff
@@ -206,30 +136,40 @@ abstract class ModalContainer<M : Any>
     }
   }
 
-  private class DialogRef<M : Any>(
-    val screen: M,
-    val dialog: Dialog
+  /**
+   * @param extra optional hook to allow subclasses to associate extra data with this dialog,
+   * e.g. its content view.
```
zach-klippenstein: Nit: Explain that `extra` is not used for equivalence.

zach-klippenstein:
Still LG in general, although @steveinflow asks some good questions.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/backstack/BackStackContainer.kt
```diff
@@ -51,45 +60,45 @@ class BackStackContainer(
...
-                  container = this,
-                  setUpNewView = updateTools::setUpNewView,
-                  direction = updateTools.direction
-              )
+        ?.let { oldView ->
+          updateTools.saveOldView(oldView)
+
+          val newScene = Scene(this, newView)
+
+          val (outEdge, inEdge) = when (updateTools.direction) {
```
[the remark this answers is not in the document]
rjrjr: At worst one implements a custom container by forking this one, which is why ViewStack is separate. But more realistically, I'm planning to simplify this in the next pr. Will keep an eye on that use case.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/LayoutRunner.kt
```diff
@@ -0,0 +1,96 @@
...
+            bindShowValueFunction(initialValue, runnerConstructor.invoke(this, registry)::showValue)
+          }
+    }
+  }
+
+  companion object {
+    inline fun <reified ValueT : Any> bind(
+      @LayoutRes layoutId: Int,
+      noinline constructor: (View, ViewRegistry) -> LayoutRunner<ValueT>
+    ): ViewBinding<ValueT> = Binding(ValueT::class, layoutId, constructor)
```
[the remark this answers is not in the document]
rjrjr: Probably the same thing we do today, an injectable factory. The semantics haven't really changed. Figure we'll port one or two and and see what reads best.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/LayoutRunner.kt
```diff
@@ -0,0 +1,96 @@
...
+import android.view.ViewGroup
+import kotlin.reflect.KClass
+
+/**
+ * An object that handles [View.showValue] calls for a view inflated
+ * from a layout resource in response to [ViewRegistry.buildView].
+ * (Use [BuilderBinding] if you want to build views from code rather than
+ * layouts.)
+ *
+ * Typical usage is to have a [LayoutRunner]'s `companion object` implement
```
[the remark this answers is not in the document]
rjrjr: Can't catch that at compile time, just like today. Very strict, descriptive exception handling is in place.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/LayoutRunner.kt
```diff
@@ -0,0 +1,96 @@
...
+@ExperimentalWorkflowUi
+interface LayoutRunner<ValueT : Any> {
+  fun showValue(value: ValueT)
+
+  class Binding<ValueT : Any>
+  constructor(
+    override val type: KClass<ValueT>,
+    @LayoutRes private val layoutId: Int,
+    private val runnerConstructor: (View, ViewRegistry) -> LayoutRunner<ValueT>
+  ) : ViewBinding<ValueT> {
```
Blisse: Are we intentionally calling this ValueT vs RenderingT? Any reason to prefer showValue() vs render()?

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/backstack/BackStackContainer.kt
```diff
@@ -51,45 +60,45 @@ class BackStackContainer(
...
+          TransitionManager.go(newScene, fullSet)
         }
-        ?: NoEffect.execute(
-            newScreen,
-            screens,
-            viewRegistry,
-            this,
-            updateTools::setUpNewView
-        )
+        ?.also { return }
```
Blisse: not important: could we use return@update here and above instead of the unqualified return?

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/LayoutRunner.kt
rjrjr: I kept going back and forth on that. If your intuition is leading the same way, I'll go through and make this `RenderingT` and `render()`.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/backstack/BackStackContainer.kt
rjrjr: Good thought, will do.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/LayoutRunner.kt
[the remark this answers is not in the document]
rjrjr: Maybe answering the wrong question. This will not compile w/o an implementation of `ViewBinding`, whether via delegating to a `bind` call or some other way.

```
val TicTacToeViewBuilders = ViewRegistry(
    NewGameLayoutRunner, GamePlayLayoutRunner, GameOverLayoutRunner
)
```

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/LayoutRunner.kt
rjrjr: Oh yeah, so much better. And `ViewShowRendering.kt`, `View.bindShowRendering`, `initialRendering`, etc.

rjrjr:
Update changes `value` to `rendering` throughout, gets rid of nasty `.also { return }` warts, punches up doc about DialogRef.extra.

rjrjr:
Update fixes terminology in the commit message.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/LayoutRunner.kt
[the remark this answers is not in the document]
rjrjr: Related: #390

Issue #385, opened 2019-06-01
Kotlin EventHandler probably too clever for prod.


rjrjr:
EventHandler does a crafty thing where all instances are equal to each other, because that was really convenient for testing. But it'll be a disaster for production if anyone decides to do some de-duping via `distinctUntilChanged`. It's kind of amazing that we haven't run into that already.

Seems like we could make RenderContext configurable, s.t. test code can set it up to do the hack, and production code will do the normal thing.

rjrjr:
EventHandler is now deprecated, it'll be gone soon.

PR #387, opened 2019-06-03
Change RenderContext.onEvent return type to be a raw function type.

zach-klippenstein:
Also added some additional kdoc on this method.

Prepares for #385, in which `onEvent` will return either a unique function or an always-equivalent function
depending on whether it's being used from a test or not.

### DIFF
--- a/kotlin/workflow-core/src/main/java/com/squareup/workflow/RenderContext.kt
+++ b/kotlin/workflow-core/src/main/java/com/squareup/workflow/RenderContext.kt
@@ -62,12 +62,28 @@ interface RenderContext<StateT, in OutputT : Any> {
    *      }
    *    )
    *
+   * ## Equivalence & Testing
+   *
+   * It is common in unit tests to get a rendering from a workflow, then (since renderings should be
+   * value types) create the expected rendering instance and compare them. However, when comparing
+   * renderings there is no meaningful way to compare event handlers (other than a null check if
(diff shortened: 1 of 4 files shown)


kotlin/workflow-core/src/main/java/com/squareup/workflow/RenderContext.kt
```diff
@@ -62,12 +62,28 @@ interface RenderContext<StateT, in OutputT : Any> {
...
+   * ## Equivalence & Testing
+   *
+   * It is common in unit tests to get a rendering from a workflow, then (since renderings should be
+   * value types) create the expected rendering instance and compare them. However, when comparing
+   * renderings there is no meaningful way to compare event handlers (other than a null check if
+   * they're nullable), and so two renderings that are identical but have different event handler
+   * functions should still be considered equal. In order to support this pattern, the functions
+   * returned by this method are all considered equal: `foo == bar` whenever `foo` and `bar` are
+   * values returned by this method. More precisely, this function returns instances of
+   * [EventHandler], and all [EventHandler] instances are considered equal.
```
vRallev: Is the last sentence still true?

zach-klippenstein: Yes. This doesn't change anything but the types. #385 requests changes to actual behavior, which will be done in a future PR.

rjrjr:
Belated LGTM

PR #388, opened 2019-06-03
Use new builder inference for Worker.create so types can be inferred from emitOutput calls.

zach-klippenstein:
While the `@BuilderInference` annotation is still experimental, I am not worried about the
risk of using here because the worst thing that can happen is the wrong type is inferred
and the call site needs to specify an explicit generic type, which is already required
right now.


kotlin/samples/hello-terminal/hello-terminal-app/src/main/java/com/squareup/sample/helloterminal/BlinkingCursorWorkflow.kt
```diff
@@ -34,7 +34,7 @@ class BlinkingCursorWorkflow(
 
   private val cursorString = cursor.toString()
 
-  private val intervalWorker = Worker.create<Boolean> {
```
kirillzh: 👋

PR #395, opened 2019-06-06
Allows WorkflowLayout to work from any stream of renderings.

### LINKED ISSUE #394: Make ViewRegistry accept just a stream of renderings
Need to just start showing a stream of renderings that are sourced from an arbitrary observable.

cc @ellapolo

rjrjr:
Closes #394.

### DIFF
--- a/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowLayout.kt
+++ b/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowLayout.kt
@@ -42,15 +42,28 @@ class WorkflowLayout(
   private val showing: View? get() = if (childCount > 0) getChildAt(0) else null
 
   /**
-   * Subscribes to [WorkflowRunner.renderings]. Uses [WorkflowRunner.viewRegistry]
-   * to [build a new view][ViewRegistry.buildView] each time a new type of rendering is received,
+   * Subscribes to [renderings], and uses [registry] to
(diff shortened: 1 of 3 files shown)


zach-klippenstein:
Cc @ellapolo

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowLayout.kt
```diff
@@ -42,15 +42,28 @@ class WorkflowLayout(
...
+    registry: ViewRegistry
+  ) {
+    takeWhileAttached(renderings) { show(it, registry) }
+  }
+
+  /**
+   * Convenience override to start this layout from [renderings][WorkflowRunner.renderings]
+   * and [viewRegistry][WorkflowRunner.viewRegistry] of [workflowRunner].
+   */
+  fun start(workflowRunner: WorkflowRunner<*>) {
```
rjrjr: @zach-klippenstein  Tempted to make this an extension in WorkflowRunner.kt. WDYT?

rjrjr: I guess that seems backward. Better for the layout to depend upon the model than vice versa.

zach-klippenstein: Sounds good!

rjrjr:
Oops, accidentally hit the re-review button. No changes, just waiting on Travis to merge.

PR #397, opened 2019-06-06
Adds LayoutRunner.bindNoRunner.

rjrjr:


kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/LayoutRunner.kt
```diff
@@ -95,5 +96,13 @@ interface LayoutRunner<RenderingT : Any> {
       @LayoutRes layoutId: Int,
       noinline constructor: (View) -> LayoutRunner<RenderingT>
     ): ViewBinding<RenderingT> = bind(layoutId) { view, _ -> constructor.invoke(view) }
+
+    inline fun <reified RenderingT : Any> bindNoRunner(
```
zach-klippenstein: What's the use case? Could use some kdoc.

rjrjr:
Update adds missing kdoc.

rjrjr:
Update makes the kdoc better.

PR #398, opened 2019-06-07
[DO NOT MERGE] Experimenting with Flow

zach-klippenstein:
Check out the individual commits.

### DIFF
--- a/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowRunnerViewModel.kt
+++ b/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowRunnerViewModel.kt
@@ -21,23 +21,23 @@ import android.os.Bundle
 import com.squareup.workflow.Snapshot
 import com.squareup.workflow.Workflow
 import com.squareup.workflow.WorkflowHost
-import io.reactivex.Flowable
-import io.reactivex.Observable
-import io.reactivex.disposables.CompositeDisposable
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Dispatchers.Unconfined
 import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.channels.ReceiveChannel
+import kotlinx.coroutines.flow.Flow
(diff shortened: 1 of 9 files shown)


kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowRunnerViewModel.kt
```diff
@@ -65,28 +68,20 @@ internal class WorkflowRunnerViewModel<OutputT : Any>(
...
-    // Buffer on backpressure so outputs don't get lost.
-    updates.toFlowable(BUFFER)
-        .filter { it.output != null }
-        .map { it.output!! }
-        // DON'T replay, outputs are events.
-        .publish()
-        // Subscribe upstream immediately so we immediately start getting notified about outputs.
-        // If [renderings] triggers the upstream subscription before we do, if the workflow emits
-        // an output immediately we might not see it.
-        .autoConnect(0) { subs.add(it) }
```
zach-klippenstein: @rjrjr see?

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowRunnerViewModel.kt
```diff
@@ -65,28 +68,20 @@ internal class WorkflowRunnerViewModel<OutputT : Any>(
...
   private val subs = CompositeDisposable()
 
-  @Suppress("EXPERIMENTAL_API_USAGE")
-  private val updates =
-    workflowUpdates.updates.asObservable(Dispatchers.Unconfined)
-        .doOnNext { lastSnapshot = it.snapshot }
-        .replay(1)
-        .autoConnect(1) { subs.add(it) }
+  init {
+    subs.add(workflowHost.renderingsAndSnapshots
```
rjrjr: Sketchy to do this in the init block, but I'm pretty sure we can do better.

zach-klippenstein: See the latest commit.

zach-klippenstein: But we could also give it an explicit start method, same thing.

zach-klippenstein:
Superseded by #426.

PR #403, opened 2019-06-11
BackStackScreen includes the entire backstack. Introduces Compatible.

### LINKED ISSUE #391: Move backstack into BackStackContainerScreen
The heuristics in `BackStackContainer` are nasty, a legacy of https://github.com/square/flow / `FragmentManager` thinking. We should instead build a list of screens into BackStackContainer, representing the visible screen and everything upstream of it — put the Workflow properly in charge.

This should allow us to eliminate the `value` notion entirely.

Note that Swift is already most of the way there.

rjrjr:
Closes #391.

`BackStackScreen.wrapped` pointer is replaced with a list, `stack`.
`ViewStateStack` is renamed `ViewStateCache` and is _much_ simpler.

The `keyExtension` mechanism from `BackStackScreen` is now an independent
utility provided by `Compatible.kt` and `Uniqued.kt`. It's used by both
`ViewShowRendering.canShowRendering` and `BackStackContainer`, and is
available to any other  container that may need to take advantage of it.
See the class doc on the `compatible` function and `Compatible` interface
for details.

`BackStackContainer` and `ViewStateCache` now rely on
`BackStackScreen.stack` to determine what view state to hold on to, and
thereby whether to run a push or pop effect. All the heuristics are gone.
See `AuthWorkflow` for an example of how much simpler life is when
the workflow is explicit about the stack.

Note  the `List.goTo` extension, expected to  be handy for  workflows that
manage back stacks — especially those replacing legacy code built around
Square's `flow.Flow`. This is pretty speculative, and if we don't find a
production use for it soon should be deleted. (I thought I'd need it
in the samples, but I didn't, which is interesting.)

`ViewRegistry` now includes default bindings for the standard screen
types.


kotlin/samples/tictactoe/android/src/main/java/com/squareup/sample/mainactivity/MainActivity.kt
```diff
@@ -53,11 +58,12 @@ class MainActivity : AppCompatActivity() {
...
+    super.onDestroy()
+  }
+
   private companion object {
-    val viewRegistry = ViewRegistry(
-        BackStackContainer,
-        ModalContainer.forAlertContainerScreen(),
-        PanelContainer
-    ) + AuthViewBindings + TicTacToeViewBindings
+    val viewRegistry = ViewRegistry(PanelContainer) + AuthViewBindings + TicTacToeViewBindings
```
rjrjr: `BackStackScreen` and `AlertContainerScreen` are now built in.

kotlin/samples/tictactoe/common/src/main/java/com/squareup/sample/panel/PanelContainerScreen.kt
```diff
@@ -34,17 +34,18 @@ import com.squareup.workflow.ui.HasModals
...
   override val modals: List<BackStackScreen<T>> = emptyList()
-) : HasModals<B, BackStackScreen<T>> {
-  constructor(
-    baseScreen: B,
-    panelBody: T
-  ) : this(baseScreen, listOf(BackStackScreen(panelBody)))
-}
+) : HasModals<B, BackStackScreen<T>>
 
+/**
```
rjrjr: Remember, this panel stuff is a sample.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/backstack/BackStackContainer.kt
```diff
@@ -49,42 +48,60 @@ import com.squareup.workflow.ui.showRendering
  * to displayed views that implement that interface themselves.
  */
 @ExperimentalWorkflowUi
-class BackStackContainer(
+open class BackStackContainer(
```
rjrjr: @steveinflow WDYT?

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/backstack/BackStackContainer.kt
```diff
@@ -49,42 +48,60 @@ import com.squareup.workflow.ui.showRendering
...
-  private val viewStateStack by lazy { restored ?: ViewStateStack() }
+  private var restored: ViewStateCache? = null
+  private val viewStateCache by lazy { restored ?: ViewStateCache() }
 
   private val showing: View? get() = if (childCount > 0) getChildAt(0) else null
 
   private lateinit var registry: ViewRegistry
 
   private fun update(newRendering: BackStackScreen<*>) {
+    val uniqued: BackStackScreen<Uniqued<*>> =
```
rjrjr: Had a comment explaining this, should bring it back.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/backstack/ViewStateCache.kt
```diff
@@ -0,0 +1,185 @@
...
+   */
+  fun save(currentView: View) {
+    val saved = SparseArray<Parcelable>().apply { currentView.saveHierarchyState(this) }
+    val newFrame = ViewStateFrame(currentView.uniquedKey, saved)
+    viewStates += newFrame.key to newFrame
+  }
+
+  /**
+   * Convenience for use in [View.onSaveInstanceState] and [View.onRestoreInstanceState]
+   * methods of container views that have no other state or their own to save.
```
rjrjr: …of their own…

kotlin/workflow-ui-core/src/main/java/com/squareup/workflow/ui/Compatible.kt
```diff
@@ -0,0 +1,60 @@
...
+ * that is [compatible] with [next]. If no matching frame is found, pushes [next]
+ * on [top].
+ */
+fun <T : Any> List<T>.goTo(next: T): List<T> {
+  val splicePoint = indexOfLast { compatible(it, next) }
+  return if (splicePoint == -1) this + next else slice(0..splicePoint)
+}
+
+/**
+ * Implemented by objects whose [compatibility][compatible] requires more nuance
```
rjrjr: implementation requires…

kotlin/workflow-ui-core/src/main/java/com/squareup/workflow/ui/Uniqued.kt
```diff
@@ -0,0 +1,45 @@
...
+  name: String
+) : Compatible<Uniqued<W>> {
+  init {
+    require(name.isNotBlank()) { "name must not be blank." }
+  }
+
+  /**
+   * Used as a comparison key by [isCompatibleWith]. Handy for use as a map key.
+   */
+  val key: String = ((wrapped as? Uniqued<*>)?.key ?: wrapped::class.jvmName) + "-$name"
```
rjrjr: Should look harder at how these names actually turn out, might not be very readable for debugging.

kotlin/workflow-ui-core/src/main/java/com/squareup/workflow/ui/Compatible.kt
rjrjr: nevermind, reads fine.

kotlin/workflow-ui-core/src/main/java/com/squareup/workflow/ui/Uniqued.kt
```diff
@@ -0,0 +1,45 @@
...
+package com.squareup.workflow.ui
+
+import kotlin.reflect.jvm.jvmName
+
+/**
+ * Allows renderings that do not implement [Compatible] themselves to be distinguished
+ * by more than just their type. Instances are [compatible] if they have the same name
+ * and have [compatible] [wrapped] fields.
+ */
+class Uniqued<W : Any>(
```
rjrjr: @zach-klippenstein  Would `Named` be a better name for this class?

kotlin/samples/tictactoe/common/src/main/java/com/squareup/sample/panel/PanelContainerScreen.kt
```diff
@@ -34,17 +34,18 @@ import com.squareup.workflow.ui.HasModals
...
 fun <B : Any, T : Any> BackStackScreen<T>.asPanelOver(baseScreen: B): PanelContainerScreen<B, T> {
   return PanelContainerScreen(baseScreen, listOf(this))
 }
 
+/**
+ * Shows the receiver as the only panel over [baseScreen], with no back stack.
+ */
 fun <B : Any, T : Any> T.asPanelOver(baseScreen: B): PanelContainerScreen<B, T> {
-  return PanelContainerScreen(baseScreen, this)
+  return PanelContainerScreen(baseScreen, listOf(BackStackScreen(this)))
```
zach-klippenstein: Nit: `BackStackScreen(this).asPanelOver(baseScreen)`?

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewRegistry.kt
```diff
@@ -80,10 +94,15 @@ class ViewRegistry private constructor(
     container: ViewGroup? = null
   ): View {
     @Suppress("UNCHECKED_CAST")
-    return (bindings[initialRendering::class] as? ViewBinding<RenderingT>)
-        ?.buildView(this, initialRendering, contextForNewView, container)
+    return (allBindings[initialRendering::class] as? ViewBinding<RenderingT>)
+        ?.buildView(this, initialRendering, contextForNewView, container)?.apply {
```
zach-klippenstein: Nit: This is a bit hard to read, consistent line breaks (before `?.`) might help?

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewRegistry.kt
```diff
@@ -80,10 +94,15 @@ class ViewRegistry private constructor(
...
+          check(showRenderingTag?.showing != null) {
```
zach-klippenstein: Nit: `checkNotNull`

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewShowRendering.kt
```diff
@@ -16,51 +16,70 @@
...
-      ShowRenderingTag(showRendering, initialRendering::class)
-  )
+  setTag(R.id.view_show_rendering_function, ShowRenderingTag(initialRendering, showRendering))
   showRendering.invoke(initialRendering)
 }
 
+/**
+ * True if this view is able to show [rendering].
+ *
+ * Returns`false` if [bindShowRendering] has not been called, so it is always safe to
```
zach-klippenstein: Nit: missing space

kotlin/workflow-ui-core/src/main/java/com/squareup/workflow/ui/Compatible.kt
```diff
@@ -0,0 +1,60 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.workflow.ui
+
+/**
+ * Returns true if [me] and [you] are instances of the same class, unless
+ * that class implements [Compatible]. In that case
```
zach-klippenstein: I don't thinks this is true? If `me` and `you` have different concrete types, the first case will fail and immediately return false.

kotlin/workflow-ui-core/src/main/java/com/squareup/workflow/ui/Compatible.kt
```diff
@@ -0,0 +1,60 @@
...
+ */
+fun compatible(
+  me: Any,
+  you: Any
+): Boolean {
+  return when {
+    me::class != you::class -> false
+    me !is Compatible<*> -> true
+    // If you see a casting error here, it's a lie. https://youtrack.jetbrains.com/issue/KT-31823
+    else -> me.isCompatibleWith(you as Compatible<*>)
```
zach-klippenstein: Could you avoid the cast (and address my above comment) by doing:
```kotlin
when {
  me is Compatible<*> && you is Compatible<*> -> me.isCompatibleWith(you)
  else -> me::class == you::class
}
```

kotlin/workflow-ui-core/src/main/java/com/squareup/workflow/ui/Uniqued.kt
zach-klippenstein: `Named`, `Keyed`, `Named/KeyedRendering`, any of those probably. Couldn't figure out what `Uniqued` actually meant till I read this kdoc.

kotlin/workflow-ui-core/src/test/java/com/squareup/workflow/ui/CompatibleTest.kt
```diff
@@ -0,0 +1,97 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.workflow.ui
+
+import org.assertj.core.api.Java6Assertions.assertThat
+import org.junit.Test
+
+// If you try to replace isTrue() with isTrue compilation fails.
```
zach-klippenstein: Neat.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewRegistry.kt
rjrjr: Much better.

kotlin/workflow-ui-core/src/main/java/com/squareup/workflow/ui/Compatible.kt
rjrjr: "that class" is the class that they are both instances of. I'll reword anyway.

rjrjr:
```
/**
 * Normally returns true if [me] and [you] are instances of the same class.
 * If that common class implements [Compatible], 
 * `me.`[canUpdateFrom][Compatible.isCompatibleWith]`(you)` must also be true.
 *
 * A convenient way to take control over the matching behavior of objects that
 * don't implement [Compatible] is to wrap them with [Named].
 */
```

kotlin/workflow-ui-core/src/main/java/com/squareup/workflow/ui/Compatible.kt
rjrjr: This isn't the contract. They must be of the same type _and_ return true from this method. Remember the recursive type param on `isCompatibleWith()`, we promise only to make the call if both left and right are instances of `T`.

Issue #401, opened 2019-06-07
Unit tests of BackStackContainer needed


rjrjr:
See if _Unmocked_ can get the job done, otherwise introduce _Robolectric_.

PR #406, opened 2019-06-12
Fixes BackStackContainer config change.

### LINKED ISSUE #400: BackstackContainer view state lost on config change
* Launch Tic Tac Toe sample
 * foo@2fa / password
 * rotate
 * go back

Transition is a push. Should be a pop. Indicates that viewStateCache was not saved

rjrjr:
Closes #400.

Our view persistence story got a lot simpler with the move to  `LayoutRunner`,
since we no longer try to render views before they have a complete rendering
to display. `BackStackContainer` didn't get the memo, and its assumption
that `onRestoreInstanceState` would come too late was breaking things.

Fix is simple, don't go like that.

### DIFF
--- a/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/backstack/BackStackContainer.kt
+++ b/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/backstack/BackStackContainer.kt
@@ -32,11 +32,10 @@ import com.squareup.workflow.ui.BackStackScreen
 import com.squareup.workflow.ui.BuilderBinding
 import com.squareup.workflow.ui.ExperimentalWorkflowUi
 import com.squareup.workflow.ui.HandlesBack
-import com.squareup.workflow.ui.R
 import com.squareup.workflow.ui.Named
+import com.squareup.workflow.ui.R
 import com.squareup.workflow.ui.ViewBinding
 import com.squareup.workflow.ui.ViewRegistry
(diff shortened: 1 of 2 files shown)


kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/backstack/BackStackContainer.kt
```diff
@@ -54,8 +53,7 @@ open class BackStackContainer(
 ) : FrameLayout(context, attributeSet), HandlesBack {
   constructor(context: Context) : this(context, null)
 
-  private var restored: ViewStateCache? = null
-  private val viewStateCache by lazy { restored ?: ViewStateCache() }
+  private val viewStateCache = ViewStateCache()
```
zach-klippenstein: 😍

rjrjr:
Still no unit / integration tests for this. I dug into that a bit and all the options still suck, so I'd prefer not to block the fix. Tracked at #401

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/backstack/BackStackContainer.kt
```diff
@@ -134,14 +132,13 @@ open class BackStackContainer(
   }
 
   override fun onSaveInstanceState(): Parcelable {
-    showing?.let { viewStateCache.save(it) }
```
rjrjr: No longer need the `save()` call, b/c Android built-in call is no longer busted.

PR #407, opened 2019-06-12
Cleans up persistence of ModalContainer.

rjrjr:
No longer necessary to save body state now that views are not created until
they have something to render. Use the recursive compatibility check from
`Named` when deciding whether or not to restore saved dialog state.

### DIFF
--- a/kotlin/workflow-ui-core/src/main/java/com/squareup/workflow/ui/Named.kt
+++ b/kotlin/workflow-ui-core/src/main/java/com/squareup/workflow/ui/Named.kt
@@ -33,7 +33,7 @@ data class Named<W : Any>(
   /**
    * Used as a comparison key by [isCompatibleWith]. Handy for use as a map key.
    */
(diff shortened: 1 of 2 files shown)


kotlin/workflow-ui-core/src/main/java/com/squareup/workflow/ui/Named.kt
```diff
@@ -42,4 +42,16 @@ data class Named<W : Any>(
   override fun toString(): String {
     return "${super.toString()}: $key"
   }
+
+  companion object {
+    /**
+     * Calculates the [compatibility key][Named.key] for a given [value] and [name].
```
zach-klippenstein: Nit: the fact that this uses `Named` is significant documentation, I wouldn't hide it behind a differently-named link. But not a big deal.

rjrjr: It's more that `Named` uses this method than the other way around. Inclined to leave it alone, I suspect I'll be back here for another round of tidying soon.

PR #409, opened 2019-06-13
Upgrade coroutines dep to 1.3.0-M1.

zach-klippenstein:
- Some things that were `@FlowPreview` have been graduated to `@ExperimentalCoroutinesApi`.
 - Breakage in `CoroutineWorkflow` was caused by https://github.com/Kotlin/kotlinx.coroutines/pull/1158.
 - Breakage in `ReactorAsWorkflowIntegrationTest` was probably caused by https://github.com/Kotlin/kotlinx.coroutines/pull/1239.

### DIFF
--- a/kotlin/legacy/legacy-workflow-core/src/test/java/com/squareup/workflow/legacy/ReactorAsWorkflowIntegrationTest.kt
+++ b/kotlin/legacy/legacy-workflow-core/src/test/java/com/squareup/workflow/legacy/ReactorAsWorkflowIntegrationTest.kt
@@ -103,6 +103,7 @@ class ReactorAsWorkflowIntegrationTest {
     subscribeToState(workflow)
     workflow.cancel()
 
+    assertEquals(SecondState("hello"), stateSub.poll())
     assertFailsWith<CancellationException> { stateSub.poll() }
     assertTrue(stateSub.isClosedForReceive)
   }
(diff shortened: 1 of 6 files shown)


kotlin/legacy/legacy-workflow-core/src/test/java/com/squareup/workflow/legacy/ReactorAsWorkflowIntegrationTest.kt
```diff
@@ -372,6 +375,7 @@ class ReactorAsWorkflowIntegrationTest {
     assertFalse(resultSub.isCompleted)
 
     workflow.sendEvent("foo")
+    assertEquals(SecondState("foo"), stateSub.poll())
```
rjrjr: Is this line for added coverage, or did the `poll()` call actually fix something?

kotlin/workflow-core/src/main/java/com/squareup/workflow/Worker.kt
```diff
@@ -328,7 +327,7 @@ inline fun <reified T> ReceiveChannel<T>.asWorker(
  * This **SHOULD NOT** be used in production code.
  */
 @VeryExperimentalWorkflow
-@FlowPreview
+@ExperimentalCoroutinesApi
```
rjrjr: 🍿

kotlin/legacy/legacy-workflow-core/src/test/java/com/squareup/workflow/legacy/ReactorAsWorkflowIntegrationTest.kt
zach-klippenstein: Conflated channels now hang onto their last value even after they're closed, so we have to pull it out first now.

See https://github.com/Kotlin/kotlinx.coroutines/issues/332

zach-klippenstein:
1.3.0-M2 was just released, adding a commit to this PR to bump again.

rjrjr:
Still LGTM

PR #410, opened 2019-06-13
Zachklipp/adventure sample

zach-klippenstein:


zach-klippenstein:
I had forgotten about `BackStackScreen` when I started writing this, and didn't think through the UI rendering vs for-parent rendering split enough.

I wrote down a better design (I think) in a separate issue: https://github.com/square/workflow/issues/411
Maybe one day I'll actually play with implementing it.

PR #419, opened 2019-06-25
Fix modules names.

zach-klippenstein:


zach-klippenstein:
The other modules are either Android- or Rx-specific, so can't ever be common modules.

PR #420, opened 2019-06-25
Rename WorkflowAction.noop() to noAction().

### LINKED ISSUE #412: Proposal: Rename noop() to noWorkflowAction()
Just seeing `noop()` inside a lambda is not very readable, you have to remember that it returns a `WorkflowAction`. `noWorkflowAction()` is much clearer.

Compare:

```kotlin
context.renderChild() {
  // Huh?
  noop()
}
```

```kotlin
context.renderChild() {
  noWorkflowAction()
}
```

Not sure this is any better, please comment!

zach-klippenstein:
Closes #412.

### DIFF
--- a/kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowAction.kt
+++ b/kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowAction.kt
@@ -41,9 +41,12 @@ interface WorkflowAction<StateT, out OutputT : Any> : (StateT) -> Pair<StateT, O
     }
 
     /**
-     * Returns a [WorkflowAction] that does nothing.
+     * Returns a [WorkflowAction] that does nothing: no output will be emitted, and `render` will be
+     * called again with the same `state` as last time.
+     *
+     * Use this to, for example, ignore the output of a child workflow or worker.
      */
-    fun <StateT, OutputT : Any> noop(): WorkflowAction<StateT, OutputT> =
+    fun <StateT, OutputT : Any> noAction(): WorkflowAction<StateT, OutputT> =
       WorkflowAction({ "noop" }) { Pair(it, null) }
 
     /**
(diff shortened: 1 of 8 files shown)


kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowAction.kt
```diff
@@ -43,7 +43,7 @@ interface WorkflowAction<StateT, out OutputT : Any> : (StateT) -> Pair<StateT, O
     /**
      * Returns a [WorkflowAction] that does nothing.
```
rjrjr: Could elaborate a bit in the kdoc, something like: 

```
Returns a [WorkflowAction] that does nothing: no output is emitted, and `render` is called again with the same `state` as last time.
```

zach-klippenstein:
@davidapgar @bencochran Does this change make sense in Swift too? Don't want to merge unless we agree on terminology.

davidapgar:
> @davidapgar @bencochran Does this change make sense in Swift too? Don't want to merge unless we agree on terminology.

If I understand what this is correctly, the closest we have in Swift is a `AnyWorkflowAction.identity`, which is still an action, but one that doesn't change the state at all.

The usage we have for it is where there is an Output from a child that we want to "ignore" in the parent (but still have to map the output to an action).

Is this the same use case? Does a `noAction` still result in a new render pass?

kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowAction.kt
zach-klippenstein: Done.

zach-klippenstein:
> The usage we have for it is where there is an Output from a child that we want to "ignore" in the parent (but still have to map the output to an action).
>
> Is this the same use case? Does a noAction still result in a new render pass?

@davidapgar Exactly. I don't like `identity` because that's even more opaque than `noAction` (it's not blindingly obvious that this thing returns a function, which is necessary to understand for that name to make sense).

davidapgar:
> > The usage we have for it is where there is an Output from a child that we want to "ignore" in the parent (but still have to map the output to an action).
> > Is this the same use case? Does a noAction still result in a new render pass?
> 
> @davidapgar Exactly. I don't like `identity` because that's even more opaque than `noAction` (it's not blindingly obvious that this thing returns a function, which is necessary to understand for that name to make sense).

Makes sense. I'm good with the name (will probably make it `AnyWorkflowAction.noAction` or a typealias to `NoAction` in Swift land).

PR #421, opened 2019-06-25
Added Maven Central and Apache License 2.0 badges to README

zach-klippenstein:
The version number is taken from the published `workflow-core` artifact, but the link points to the page for all packages under the `com.squareup.workflow` group.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 # workflow
 
+[![GitHub license](https://img.shields.io/badge/license-Apache%20License%202.0-blue.svg?style=flat)](https://www.apache.org/licenses/LICENSE-2.0)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,5 +1,8 @@
 # workflow
 
+[![GitHub license](https://img.shields.io/badge/license-Apache%20License%202.0-blue.svg?style=flat)](https://www.apache.org/licenses/LICENSE-2.0)
+[![Maven Central](https://img.shields.io/maven-central/v/com.squareup.workflow/workflow-core.svg?label=Maven%20Central)](https://search.maven.org/search?q=g:%22com.squareup.workflow%22)
```
rjrjr: Is there any similar treatment we can do for Cocoapods? If not a badge, an id / version number?

cc @davidapgar @segiddins

davidapgar: I think for Cocoapods it would be:
[CocoaPods compatible]

zach-klippenstein: Sweet! Adding…

zach-klippenstein: Done.

PR #424, opened 2019-06-26
Add instructions to README about consuming Maven artifacts.

zach-klippenstein:


zach-klippenstein:
Not sure if this table would work better as a definition list:

<dl>
<dt> workflow-core </dt>
<dd>
Depend on this if you are writing a library module/project that uses Workflows, and you don't need to interact directly with the runtime.

```groovy
implementation "com.squareup.workflow:workflow-core:${workflowVersion}"
```
</dd>

<dt> workflow-rx2 </dt>
<dd>
Depend on this if you need to interact with RxJava2 from your Workflows. 

```groovy
implementation "com.squareup.workflow:workflow-rx2:${workflowVersion}"
```
</dd>

<dt> workflow-testing </dt>
<dd>
Depend on this if you are writing tests. This should only be included as a test dependency.

```groovy
implementation "com.squareup.workflow:workflow-testing:${workflowVersion}"
```
</dd>

<dt> workflow-ui-android </dt>
<dd>
Depend on this if you are writing an Android app that uses Workflows. This is only needed for your main application code. If your app consists of multiple modules, most of your modules should only need to depend on `workflow-core`.

```groovy
implementation "com.squareup.workflow:workflow-ui-android:${workflowVersion}"
```
</dd>
</dl>

#### Lower-level Artifacts

Most code shouldn't need to depend on these directly. They should generally only be used to build higher-level integrations with UI frameworks.

<dl>
<dt> workflow-runtime </dt>
<dd>
Depend on this if you need to interact directly with the runtime, i.e. streams of renderings and outputs.

```groovy
implementation "com.squareup.workflow:workflow-runtime:${workflowVersion}"
```
</dd>

<dt> workflow-ui-core </dt>
<dd>
Depend on this if you are writing something like `workflow-ui-android` for another UI framework. Defines the core types used by that artifact. 

```groovy
implementation "com.squareup.workflow:workflow-ui:${workflowVersion}"
```
</dd>
</dl>

zach-klippenstein:
Remember you can click this button to see a rich diff of the rendered markdown.
Untitled

zach-klippenstein:
I think I like the table a bit better, the wrapping isn't ideal but it's more concise and this README is only going to get bigger in the near future.

README.md
```diff
@@ -15,6 +15,36 @@ This project is under active development. The workflow code is being piloted in
 
 More documentation and samples will be coming soon!
 
+## Getting the library
+
+### Swift
+
+[![CocoaPods compatible](https://img.shields.io/cocoapods/v/Workflow.svg)](https://cocoapods.org/pods/Workflow)
+
+Just consume the workflow CocoaPods.
```
zach-klippenstein: @davidapgar PTAL

zach-klippenstein:
I copied the badges for each repo in their sections, but also left them at the top of the file so they're super obvious for someone who just wants to check that they're on the latest version.

README.md
davidapgar: Maybe something more like:
If you use CocoaPods to manage your dependencies, simply add Workflow and WorkflowUI to your Podfile:
```
pod 'Workflow'
pod 'WorkflowUI'
```

zach-klippenstein: Done.

README.md
```diff
@@ -15,6 +15,41 @@ This project is under active development. The workflow code is being piloted in
...
+pod 'WorkflowUI'
+```
+
+### Kotlin
+
+[![Maven Central](https://img.shields.io/maven-central/v/com.squareup.workflow/workflow-core.svg?label=Maven%20Central)](https://search.maven.org/search?q=g:%22com.squareup.workflow%22)
+
+Artifacts are hosted on Maven Central. If you're using Gradle, ensure `mavenCentral()` appears in your `repositories` block, and then add dependencies on the following artifacts:
+
+| Subproject | Depend on this if… | Maven Coordinates |
```
rjrjr: This table is a bit wide, and so hard to read. WDYT of dropping the headers and making it just two cells, moving the coordinates to the left, under the names:

| Subproject        | Depend on this if… | 
|-------------------|--------------------|
| workflow-core<br>`com.squareup.workflow:workflow-core:${workflowVersion}`| You are writing a library module/project that uses Workflows, but you don't need to interact with the runtime from the outside. | 
| workflow-rx2<br>`com.squareup.workflow:workflow-rx2:${workflowVersion}` | You need to interact with RxJava2 from your Workflows. | 
| workflow-testing<br>`com.squareup.workflow:workflow-testing:${workflowVersion}` | You are writing tests. This should only be included as a test dependency. | 
| workflow-ui-android<br>`com.squareup.workflow:workflow-ui-android:${workflowVersion}` | You're writing an Android app that uses Workflows. |

rjrjr: Shoot, the coordinates are still wrapping, which is what bothered me most.

rjrjr: Also might be nicer as a simple dependencies block, with inline comments for the "depend on this" bit.

```
dependencies {
  // When writing a library module/project that uses Workflows, with no need to interact with the runtime from the outside.
  implementation com.squareup.workflow:workflow-core:${workflowVersion}

 // etc...
}
```

zach-klippenstein: Fixed the tables. Github supports basic HTML table syntax, including the `nowrap` attribute, so they wrap correctly now.

zach-klippenstein: I would like to avoid the giant code block, it's harder to scan.

rjrjr:
:shipit:

PR #426, opened 2019-06-27
Replace updates channel with rendering/snapshot and output Flows on WorkflowHost

zach-klippenstein:
_There are two commits, the first just cleans up the existing `WorkflowHost` implementation to make the main change easier to implement and review._

The new Flows are backed by `BroadcastChannels`. The `renderingsAndSnapshots` Flow
is backed by a `ConflatedBroadcastChannel`, so it doesn't get backpressure and will
always emit the latest rendering + snapshot on collection. The `outputs` Flow is
a regular `BroadcastChannel` with a capacity of 1 (one more than the old channel,
which was unbuffered). `BroadcastChannel`s aren't allowed to be unbuffered, so while
this channel experiences backpressure, it's no longer perfect lock-step.

This change also updates all the internal consumers of `WorkflowHost`s to use the Flows.
Namely, `WorkflowTester` and `WorkflowRunnerViewModel`, the latter of which now simply
converts the `Flow`s to `Observable`s/`Flowable`s with `asObservable()`/`asFlowable()`.

### DIFF
--- a/kotlin/build.gradle
+++ b/kotlin/build.gradle
@@ -41,6 +41,7 @@ buildscript {
       'mockito': '2.7.5',
       'mockitoKotlin': '1.5.0',
       'rxjava2Extensions': '0.20.4',
+      'truth': '1.0-rc1',
   ]
 
   ext.deps = [
@@ -95,6 +96,7 @@ buildscript {
       ],
       'test': [
           'assertj': "org.assertj:assertj-core:${versions.assertj}",
+          'truth': "com.google.truth:truth:${versions.truth}",
           'junit': "junit:junit:${versions.junit}",
           'mockito': "org.mockito:mockito-core:${versions.mockito}",
           'hamcrestCore': "org.hamcrest:hamcrest-core:${versions.hamcrest}"
--- a/kotlin/workflow-runtime/src/main/java/com/squareup/workflow/WorkflowHost.kt
+++ b/kotlin/workflow-runtime/src/main/java/com/squareup/workflow/WorkflowHost.kt
@@ -22,7 +22,6 @@ import com.squareup.workflow.internal.id
 import kotlinx.coroutines.CoroutineName
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.ExperimentalCoroutinesApi
-import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.channels.ReceiveChannel
 import kotlinx.coroutines.channels.consume
@@ -88,21 +87,17 @@ interface WorkflowHost<out OutputT : Any, out RenderingT> {
       inputs: () -> ReceiveChannel<InputT>,
       snapshot: Snapshot? = null,
       context: CoroutineContext = EmptyCoroutineContext
-    ): WorkflowHost<OutputT, RenderingT> =
-      object : WorkflowHost<OutputT, RenderingT> {
+    ): WorkflowHost<OutputT, RenderingT> = RealWorkflowHost(
         // Put the coroutine name first so the passed-in contexts can override it.
-        private val scope = CoroutineScope(DEFAULT_WORKFLOW_COROUTINE_NAME + baseContext + context)
-
-        override val updates: ReceiveChannel<Update<OutputT, RenderingT>> =
-          scope.produce(capacity = 0) {
-            runWorkflowTree(
-                workflow = workflow.asStatefulWorkflow(),
-                inputs = inputs,
-                initialSnapshot = snapshot,
-                onUpdate = ::send
-            )
-          }
-      }
+        context = DEFAULT_WORKFLOW_COROUTINE_NAME + baseContext + context
+    ) { onUpdate ->
+      runWorkflowTree(
+          workflow = workflow.asStatefulWorkflow(),
+          inputs = inputs,
+          initialSnapshot = snapshot,
+          onUpdate = onUpdate
+      )
+    }
 
     fun <OutputT : Any, RenderingT> run(
       workflow: Workflow<Unit, OutputT, RenderingT>,
@@ -124,22 +119,17 @@ interface WorkflowHost<out OutputT : Any, out RenderingT> {
       workflow: StatefulWorkflow<InputT, StateT, OutputT, RenderingT>,
       inputs: () -> ReceiveChannel<InputT>,
       initialState: StateT
-    ): WorkflowHost<OutputT, RenderingT> =
-      object : WorkflowHost<OutputT, RenderingT> {
-        override val updates: ReceiveChannel<Update<OutputT, RenderingT>> =
-          GlobalScope.produce(
-              capacity = 0,
-              context = DEFAULT_WORKFLOW_COROUTINE_NAME + baseContext
-          ) {
-            runWorkflowTree(
-                workflow = workflow.asStatefulWorkflow(),
-                inputs = inputs,
-                initialSnapshot = null,
-                initialState = initialState,
-                onUpdate = ::send
-            )
-          }
-      }
+    ): WorkflowHost<OutputT, RenderingT> = RealWorkflowHost(
+        context = DEFAULT_WORKFLOW_COROUTINE_NAME + baseContext
+    ) { onUpdate ->
+      runWorkflowTree(
+          workflow = workflow.asStatefulWorkflow(),
+          inputs = inputs,
+          initialSnapshot = null,
+          initialState = initialState,
+          onUpdate = onUpdate
+      )
+    }
 
     private fun <T> channelOf(value: T): () -> ReceiveChannel<T> {
       return {
@@ -148,6 +138,25 @@ interface WorkflowHost<out OutputT : Any, out RenderingT> {
       }
     }
   }
+}
+
+/**
+ * Implements the [WorkflowHost] interface using the [CoroutineScope.produce] builder.
+ */
+private class RealWorkflowHost<O : Any, R>(
+  context: CoroutineContext,
+  run: suspend (
+    onUpdate: suspend (Update<O, R>) -> Unit
+  ) -> Unit
+) : WorkflowHost<O, R> {
+
+  private val scope = CoroutineScope(context)
+
+  @UseExperimental(ExperimentalCoroutinesApi::class)
+  override val updates: ReceiveChannel<Update<O, R>> =
+    scope.produce(capacity = 0) {
+      run(channel::send)
+    }
 }
 
 /**
--- a/kotlin/workflow-runtime/src/main/java/com/squareup/workflow/WorkflowHost.kt
+++ b/kotlin/workflow-runtime/src/main/java/com/squareup/workflow/WorkflowHost.kt
@@ -16,17 +16,18 @@
 package com.squareup.workflow
 
 import com.squareup.workflow.WorkflowHost.Factory
-import com.squareup.workflow.WorkflowHost.Update
+import com.squareup.workflow.WorkflowHost.RenderingAndSnapshot
 import com.squareup.workflow.internal.WorkflowNode
 import com.squareup.workflow.internal.id
+import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineName
-import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.Job
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.channels.ReceiveChannel
 import kotlinx.coroutines.channels.consume
-import kotlinx.coroutines.channels.produce
 import kotlinx.coroutines.ensureActive
+import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.selects.select
 import org.jetbrains.annotations.TestOnly
 import kotlin.coroutines.CoroutineContext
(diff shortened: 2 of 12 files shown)


kotlin/workflow-runtime/src/main/java/com/squareup/workflow/WorkflowHost.kt
```diff
@@ -141,39 +188,75 @@ interface WorkflowHost<out OutputT : Any, out RenderingT> {
 }
 
 /**
- * Implements the [WorkflowHost] interface using the [CoroutineScope.produce] builder.
+ * Implements the [WorkflowHost] interface by [launching][CoroutineScope.launch] a new coroutine
+ * on [start].
+ *
+ * @param context The context used to launch the runtime coroutine.
+ * @param run Gets invoked from a new coroutine started with `context`, and passed functions
```
rjrjr: nit:

> _Gets invoked from a new coroutine started with `context`_ 

Maybe leaking a bit of implementation detail there? Would this be better?

_Gets invoked under the given [context]_

zach-klippenstein: This entire class is an implementation detail, and "invoked under the given context" implies a coroutine was started with context anyway.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/WorkflowHost.kt
```diff
@@ -36,27 +45,63 @@ import kotlin.coroutines.coroutineContext
...
+   * Start the workflow. [renderingsAndSnapshots] and [outputs] won't emit anything until this
+   * is called.
+   */
+  fun start()
+
+  /**
+   * Stop the workflow tree, cancelling every worker and workflow. This will cause the
+   * [renderingsAndSnapshots] and [outputs] Flows to complete.
+   */
+  fun cancel()
```
zach-klippenstein: An alternative design to having an explicit cancel method is to make `start` suspending and call it `run` or something. It suspends until the workflow throws an exception or the coroutine is cancelled. Feels a bit safer, since if you use structured concurrency you can't forget to cancel the runtime.

We could also then use the context from which `run` is called to determine the context for the session, although I'm not sure I like moving the context out of the factory – that's the whole reason the factory exists.

rjrjr: I don't really have the instincts for this yet, but this kind of sounds like an all or nothing situation: if we're not comfortable making `suspend fun run()` work in the calling context, it seems wrong to make it `suspend`. My gut is to declare victory and merge this.

zach-klippenstein: Yea agree on that first point.

I'm still going to change this API though. `start()` will return a `Job`, and I'll delete `cancel()`. Then up in `WorkflowRunner`, I'll give it its own start method that will hang onto the `Job` and cancel it when the view model is `onCleared`. `WorkflowRunner` needs a start method too so that consumers can subscribe to `output` before starting a workflow that could potentially emit an output right away.

kotlin/samples/hello-workflow/src/main/java/com/squareup/sample/helloworkflow/HelloWorkflowActivity.kt
```diff
@@ -30,6 +30,7 @@ class HelloWorkflowActivity : AppCompatActivity() {
   override fun onCreate(savedInstanceState: Bundle?) {
     super.onCreate(savedInstanceState)
     runner = setContentWorkflow(viewRegistry, HelloWorkflow, savedInstanceState)
```
rjrjr: I'm not thrilled with this extension method any more, but have no change to suggest for this PR. I'll follow up after you merge.

zach-klippenstein: Pass an optional (default true) parameter `startAutomatically` so you only have to tack on this extra `startWorkflow` call if you care about outputs? That feels kinda hacky…

zach-klippenstein: I'm also not a fan of the way this works right now because it means if there's any period of time when there are no activities/fragments subscribed, outputs sent during that time will be dropped.

zach-klippenstein: I _think_ what I kind of want is a multicasting operator that buffers items _when there are zero subscribers_, and is just a pass-through `publish`-type broadcast when there are _one or more subscribers_. Basically just ensures that _someone_, at least one observer, sees every element. But that's tricky to reason about, because if I'm subscribing to this stream, whether or not I miss an item can be decided anywhere else in the code by whether or not someone else beats me to subscribing or not.

Leaving outputs as a channel doesn't solve the multicasting problem, but just abdicates responsibility for even dealing with it. So it just pushes it downstream to `WorkflowRunner` which, if we want to keep it Rx-only, still needs to multicast that channel to turn it into an `Observable`.

zach-klippenstein: Maybe `WorkflowRunner` is the better place to solve this though. `WorkflowRunner` completely owns the `WorkflowHost`, so doesn't need to worry about arbitrary multicasting. If `WorkflowHost` exposes outputs as a `ReceiveChannel`, `WorkflowRunner` can require a statically-configured handler for outputs, instead of exposing them as a stream. So when you call `setContentWorkflow` (or `WorkflowRunner.of`), you have to pass in a function that handles outputs. This is probably a terrible idea because it would be super easy to accidentally close over your `Activity`/`Fragment` instance and create memory leaks, or at least end up trying to interact with dead components if outputs are emitted at the wrong time.

zach-klippenstein: `WorkflowLayout` could help with this. Since it knows when it's attached or not, it could be responsible for backpressure when nothing's attached. So `WorkflowRunner` could expose a `ReceiveChannel` of outputs, and `setContentWorkflow` could take the output handler function and pass it to `WorkflowLayout`. `WorkflowLayout`'s internal `Coordinator` could then subscribe to the channel only when there's a _valid_ observer.

zach-klippenstein: Making `outputs` a non-broadcast channel again would mean that we could drop the `start` method and tie `WorkflowHost` cancellation to the outputs channel, like it was to the `updates` channel before.

zach-klippenstein: Chatted about this offline, summarizing in #430.

zach-klippenstein:
@rjrjr  I got rid of `WorkflowRunner.startWorkflow`, went back to the lazy behavior we had before.

kotlin/build.gradle
```diff
@@ -41,6 +41,7 @@ buildscript {
       'mockito': '2.7.5',
       'mockitoKotlin': '1.5.0',
       'rxjava2Extensions': '0.20.4',
+      'truth': '1.0-rc1',
```
zach-klippenstein: Been meaning to switch to this. Might as well start using it for new code, I'll migrate the rest of our stuff over in a bit.

zach-klippenstein: Filed as #436

rjrjr:
> `BroadcastChannel`s aren't allowed to be unbuffered

Are we concerned about this at all? Should we be giving JetBrains the feedback that we'd like to be able to broadcast w/o being forced to provide an initial value to all subscribers? Seems like they're missing a `PublishSubject` analog.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowRunnerViewModel.kt
```diff
@@ -59,46 +72,67 @@ internal class WorkflowRunnerViewModel<OutputT : Any>(
...
-        .map { it.output!! }
-        // DON'T replay, outputs are events.
-        .publish()
-        // Subscribe upstream immediately so we immediately start getting notified about outputs.
-        // If [renderings] triggers the upstream subscription before we do, if the workflow emits
-        // an output immediately we might not see it.
-        .autoConnect(0) { subs.add(it) }
+  @UseExperimental(ExperimentalCoroutinesApi::class)
+  override val output: Flowable<out OutputT> = workflowHost.outputs
+      // Lazily start the workflow the first time someone starts listening. Like autoConnect(1).
```
rjrjr: I found this comment more confusing than helpful -- at first it made me think you meant that `onCollect` is analogous to `autoConnect(1)`, but really it's just `onSubscribe`, right? We get the `autoConnect(1)` behavior by virtue of calling `start()`, and because `start()` is idempotent. And that seems well enough documented by the comment below. If I have this right then perhaps delete this comment and change the one below to:

```
  // This call is idempotent and will always return the same Job, so we can just call it
  // every time. And so by making this call inside of onCollect(), we basically get the
  // nice effects of Rx's autoConnect(1) start up a shared stream the the first time
  // someone "subscribes" / collects.
```

And obv. the same for the identical block above.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowRunnerViewModel.kt
```diff
@@ -59,46 +72,67 @@ internal class WorkflowRunnerViewModel<OutputT : Any>(
...
+  @TestOnly
+  internal fun getLastSnapshotForTest() = lastSnapshot
+
   private companion object {
     val BUNDLE_KEY = WorkflowRunner::class.jvmName + "-workflow"
   }
 }
+
+/**
+ * Invokes `block` every time a new collector begins collecting this [Flow].
```
rjrjr: Whoa, I didn't realize you had built this method yourself. Neat. Seems like it should be built in: worth submitting?

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowRunnerViewModel.kt
```diff
@@ -59,46 +72,67 @@ internal class WorkflowRunnerViewModel<OutputT : Any>(
...
-        .doOnNext { lastSnapshot = it.snapshot }
-        .replay(1)
-        .autoConnect(1) { subs.add(it) }
+  private var workflowJob: Job? = null
 
   private var lastSnapshot: Snapshot = Snapshot.EMPTY
 
-  override val renderings: Observable<out Any> = updates.map { it.rendering }
+  @UseExperimental(ExperimentalCoroutinesApi::class)
+  override val renderings: Observable<out Any> = workflowHost.renderingsAndSnapshots
```
rjrjr: So this crafty `autoConnect()`-esque scheme means that we _don't_ have to worry about making any kind of start call around `setContentWorkflow()`, right? We still have the problem of the various races around `output()`, but that isn't made any worse by this PR.

zach-klippenstein: Correct. This gives us the same behavior we had before.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowRunnerViewModel.kt
zach-klippenstein: Yes, I plan to.

kotlin/workflow-runtime/src/test/java/com/squareup/workflow/internal/RealWorkflowHostTest.kt
```diff
@@ -0,0 +1,323 @@
...
+      emittedOutputs.send("two")
+      emittedOutputs.close()
+    }
+    host.start()
+
+    assertEquals("one", emittedOutputs.poll())
+    assertEquals("two", emittedOutputs.poll())
+  }
+
+  @Test fun `outputs flow honors backpressure`() {
```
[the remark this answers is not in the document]
zach-klippenstein: It is confusing, but I think it kind of has to be. Every collector gets its own buffer. Values sent to the channel **when there are no subscribers** are lost. Values sent to the channel **when one or more subscribers** exist are delivered to all subscribers' queues, and suspend if any of those queues are full.

> Do all collectors have to poll value _able_ before any collectors receive value _baker_? 

All collectors subscribed at the time _able_ was sent must receive _able_ before seeing _baker_. Collectors subscribing after _able_ was sent but before _baker_ was sent will only see _baker_.

zach-klippenstein:
> Are we concerned about this at all? 

I'm not worried. I'm not sure why they don't allow an unbuffered broadcast channel, but it still means we have backpressure, just there's more buffering.

> Should we be giving JetBrains the feedback that we'd like to be able to broadcast w/o being forced to provide an initial value to all subscribers? 

We are not forced to do that. It just means we won't suspend on sending the first value.

> Seems like they're missing a PublishSubject analog.

This is `PublishSubject`, it's just `PublishSubject` with buffered backpressure.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowRunnerViewModel.kt
zach-klippenstein: Done (extracted into a function too).

PR #432, opened 2019-06-29
Set up a teardown hook for dialogs.

### LINKED ISSUE #51: Fix leaky dialog warning in sample
Our old friend wants attention. I'm hoping we can use a lifecycle observer instead of the usual hacks, once #50 lands and we move to AndroidX.

```
android.view.WindowLeaked: Activity com.squareup.sample.authgameapp.ShellActivity has leaked window DecorView@9920615[] that was originally added here
    at android.view.ViewRootImpl.<init>(ViewRootImpl.java:485)
    at android.view.WindowManagerGlobal.addView(WindowManagerGlobal.java:346)
    at android.view.WindowManagerImpl.addView(WindowManagerImpl.java:93)
    at android.app.Dialog.show(Dialog.java:330)
    at com.squareup.sample.authgameapp.ShellCoordinator.ensureDialog(ShellCoordinator.kt:95)
    at com.squareup.sample.authgameapp.ShellCoordinator.access$ensureDialog(ShellCoordinator.kt:41)
    at com.squareup.sample.authgameapp.ShellCoordinator$attach$1.accept(ShellCoordinator.kt:65)
```

Maybe want this too:

```
private val Context.activity: AppCompatActivity
  get() {
    var wrapper = this as ContextWrapper
    var activity: Activity? = null
    do {
      if (wrapper is Activity) {
        activity = wrapper
      } else {
        wrapper = wrapper.baseContext as ContextWrapper
      }
    } while (activity == null)
    return activity as AppCompatActivity
  }
```

rjrjr:
This is to prevent logcat noise about leaked windows. Mind you, nothing bad
was actually happening. The windows still got closed. We just got scolded for
being presumptuous enough to ask the runtime to do us a tiny favor.

Closes #51.

### DIFF
--- a/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ModalContainer.kt
+++ b/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ModalContainer.kt
@@ -15,8 +15,10 @@
  */
 package com.squareup.workflow.ui
 
+import android.app.Activity
 import android.app.Dialog
 import android.content.Context
+import android.content.ContextWrapper
 import android.os.Bundle
 import android.os.Parcel
 import android.os.Parcelable
@@ -26,6 +28,10 @@ import android.support.annotation.StyleRes
 import android.util.AttributeSet
 import android.view.View
 import android.widget.FrameLayout
+import androidx.appcompat.app.AppCompatActivity
+import androidx.lifecycle.Lifecycle.Event.ON_DESTROY
+import androidx.lifecycle.LifecycleObserver
+import androidx.lifecycle.OnLifecycleEvent
 import com.squareup.workflow.ui.HandlesBack.Helper
 import com.squareup.workflow.ui.ModalContainer.Companion.forAlertContainerScreen
 import com.squareup.workflow.ui.ModalContainer.Companion.forContainerScreen
@@ -72,11 +78,18 @@ abstract class ModalContainer<ModalRenderingT : Any>
         dialogs[i].copy(modalRendering = modal)
             .also { updateDialog(it) }
       } else {
-        buildDialog(modal, registry).apply { dialog.show() }
+        buildDialog(modal, registry).apply {
+          dialog.show()
+          // Android makes a lot of logcat noise if it has to close the window for us. :/
+          // https://github.com/square/workflow/issues/51
+          dialog.activity?.lifecycle?.addObserver(OnDestroy {
+            dialog.dismiss()
+          })
+        }
       }
     }
 
-    (dialogs - newDialogs).forEach { it.dialog.hide() }
+    (dialogs - newDialogs).forEach { it.dialog.dismiss() }
     dialogs = newDialogs
   }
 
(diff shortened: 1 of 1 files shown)


kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ModalContainer.kt
```diff
@@ -242,3 +255,28 @@ abstract class ModalContainer<ModalRenderingT : Any>
...
+    var wrapper = this as ContextWrapper
+    var activity: Activity? = null
+    do {
+      if (wrapper is Activity) {
+        activity = wrapper
+      } else {
+        wrapper = wrapper.baseContext as ContextWrapper
+      }
+    } while (activity == null)
+    return activity as AppCompatActivity
```
rjrjr: Kind of late for this cast, should be using it in the loop.

zach-klippenstein: Could write this more concisely as a tailrec function:
```kotlin
tailrec fun Context.activity(): AppCompatActivity {
  if (this is Activity) return this
  return (this as ContextWrapper).baseContext.activity()
}
```

rjrjr:
```kotlin
tailrec fun Context.activity(): AppCompatActivity? = when {
  this is Activity -> this as? AppCompatActivity
  else -> (this as? ContextWrapper)?.baseContext?.activity()
}
```

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ModalContainer.kt
```diff
@@ -72,11 +78,18 @@ abstract class ModalContainer<ModalRenderingT : Any>
         dialogs[i].copy(modalRendering = modal)
             .also { updateDialog(it) }
       } else {
-        buildDialog(modal, registry).apply { dialog.show() }
+        buildDialog(modal, registry).apply {
+          dialog.show()
+          // Android makes a lot of logcat noise if it has to close the window for us. :/
+          // https://github.com/square/workflow/issues/51
+          dialog.activity?.lifecycle?.addObserver(OnDestroy {
```
pyricau: `dialog.activity?`  What's that API? Is this an extension function?

pyricau: oh, right, below.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ModalContainer.kt
```diff
@@ -242,3 +255,26 @@ abstract class ModalContainer<ModalRenderingT : Any>
...
+/**
+ * The [AppCompatActivity] for this context, or null if there isn't one.
+ *
+ * We keep all this very forgiving because we're just using it to keep some logging
+ * noise out of logcat. If someone manages to run this under the wrong type of
+ * activity, just return null and let the caller no-op.
+ */
+tailrec fun Context.activity(): AppCompatActivity? = when {
+  this is Activity -> this as? AppCompatActivity
+  else -> (this as? ContextWrapper)?.baseContext?.activity()
```
pyricau: This assumes there's only one context layer between the dialog and the activity. Is this guaranteed here? I've heard legends of a certain Square apps where there can be 10 layers of context and extracting the activity requires a while loop.

pyricau: and then I actually read the code and realized this is a recursive.. and tailrec so while loop. Oh well, all good.

PR #433, opened 2019-06-29
Convert InputT streams from channels to Flows.

zach-klippenstein:
Workflows accept _streams_ of inputs, which were previously passed around
as "channel producers" of the type `() -> ReceiveChannel<InputT>`.

Channels are not a good abstraction for this – they don't compose as well
as Flows, and they are inherently more expensive since they are meant to
be thread-safe, lock-free synchronization and queuing devices. Flows are
perfect for this API, since they are cold, easily composable, and map
directly to various reactive streams libraries (RxJava, Reactive Streams, etc.)

### DIFF
--- a/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowRunner.kt
+++ b/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowRunner.kt
@@ -28,9 +28,11 @@ import io.reactivex.Flowable
 import io.reactivex.Observable
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.ObsoleteCoroutinesApi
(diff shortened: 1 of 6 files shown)


kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowRunner.kt
```diff
@@ -95,23 +98,23 @@ interface WorkflowRunner<out OutputT> {
      * It's probably more convenient to use [FragmentActivity.setContentWorkflow]
      * rather than calling this method directly.
      */
-    @UseExperimental(ObsoleteCoroutinesApi::class)
-    fun <InputT, OutputT : Any> of(
+    @UseExperimental(ObsoleteCoroutinesApi::class, ExperimentalCoroutinesApi::class)
+    fun <InputT : Any, OutputT : Any> of(
```
zach-klippenstein: Had to add this constraint because the `Flowable.asFlow()` extension is only defined for non-nullable types. And it's just more correct anyway.

zach-klippenstein:
The wonders of Kotlin never cease.

PR #441, opened 2019-07-04
Added builder functions for creating stateful workflows.

zach-klippenstein:
Also made the `Workflow.stateless` function inline.

This lets you define stateful workflows with a function call:
```kotlin
Workflow.stateful<Input, State, Output, Rendering>(
  initialState = { input, snapshot -> TODO() },
  render = { input, state -> renderChild(TODO()) },
  snapshot = { state -> TODO() }
)
```

This is just an idea – does it really make anything better? These test workflows are a horrible test case for ergonomics because they're all very unusually short and usually only implement a subset of the functionality that a real workflow would implement.


rjrjr:
It's hard to judge until we try to use it in earnest. I can imagine backing away from it; I can imagine deprecating the abstract classes and going with this entirely. I like how it emphasizes the functional nature of a workflow.

I think we should merge it and find a couple of interesting spots to try it out -- see how it gets along with injection, for example. So long as we're willing to throw it away again, now is the time to give it a try.

PR #442, opened 2019-07-05
Expect child workflows and outputs in render tester

### LINKED ISSUE #438: RenderTester should allow expecting child workflows
Right now, when using the RenderTester, if a child workflow spawns any workers, the tester will fail (due to it having a separate RenderTesterContext that doesn't expect any workers). This prevents testing against a parent workflow that has a child that spawns a worker in the initial state (ie: a child that "loads" some data).

Paths:
1) Add support for "expecting" a child workflow with controlling what it's rendering is (ie: fail if the child isn't awaited)
2) Allow expectations for the children as well (this becomes more of a gray-box/black-box test, where using the `WorkflowHost` is likely a more accurate simulation.

davidapgar:
Closes #438

Adds support to validated a workflow has a child in render, as well as that it outputs

### DIFF
--- a/swift/WorkflowTesting/Sources/WorkflowRenderTester.swift
+++ b/swift/WorkflowTesting/Sources/WorkflowRenderTester.swift
@@ -39,6 +39,12 @@ extension Workflow {
 ///                 worker: TestWorker(),
 ///                 output: TestWorker.Output.success),
 ///             ...,
+///         ]
+///         expectedWorkflows: [
+///             ExpectedWorkflow(
+///                 type: ChildWorkflow.self,
+///                 rendering: "rendering",
+///                 output: ChildWorkflow.Output.success
 ///         ]),
 ///         assertions: { rendering in
 ///             XCTAssertEqual("expected text on rendering", rendering.text)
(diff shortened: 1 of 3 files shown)


swift/WorkflowTesting/Sources/WorkflowRenderTester.swift
```diff
@@ -114,6 +149,26 @@ public final class RenderTester<WorkflowType: Workflow> {
         return self
     }
 
+    /// Convenience method for testing without creating an explicit RenderExpectation.
+    @discardableResult
+    public func render(
```
davidapgar: Added this after discussion with @timdonnelly - improves the ergonomics (instead of making a full `RenderExpectations`, can just specify which expectations on the call to `render` with the tester)

swift/WorkflowTesting/Tests/WorkflowRenderTesterTests.swift
```diff
@@ -121,7 +168,6 @@ private struct TestWorkflow: Workflow {
     }
 
     func workflowDidChange(from previousWorkflow: TestWorkflow, state: inout TestWorkflow.State) {
-        // TODO: Add a behavior when it changes to validate.
```
davidapgar: Testing `workflowDidChange` can (and should) be done directly against a workflow, not with the render tester - so this comment doesn't make sense.

swift/WorkflowTesting/Sources/WorkflowRenderTester.swift
```diff
@@ -114,6 +149,26 @@ public final class RenderTester<WorkflowType: Workflow> {
...
+        expectedState: ExpectedState<WorkflowType>? = nil,
+        expectedOutput: ExpectedOutput<WorkflowType>? = nil,
+        expectedWorkers: [ExpectedWorker] = [],
+        expectedWorkflows: [ExpectedWorkflow] = [],
+        assertions: (WorkflowType.Rendering) -> Void,
+        file: StaticString = #file, line: UInt = #line
+        ) -> RenderTester<WorkflowType> {
```
bencochran: annoying whitespace nit: outdent this line

PR #447, opened 2019-07-07
Replace WorkflowHost with the launchWorkflowIn function.

### LINKED ISSUE #439: Replace WorkflowHost with a static runWorkflow function
`WorkflowHost.Factory` isn't useful, and `WorkflowHost.start` isn't quite idiomatic. Instead:

```
suspend fun <I, O, R> runWorkflow(
  workflow: Workflow<I, O, R>,
  inputs: Flow<I>,
  block: CoroutineScope.(Flow<RS>, Flow<O>) -> Unit
): Nothing
```

Note that block _does not_ `suspend`. It is a setup function. The workflow is not started until block exits. `runWorkflow` blocks forever, since workflows have no way to end themselves.

zach-klippenstein:
First commit just does some refactoring to make the next commit diff cleaner.

The new API is a single function:
```kotlin
fun <InputT, OutputT : Any, RenderingT, ResultT> launchWorkflowIn(
  scope: CoroutineScope,
  workflow: Workflow<InputT, OutputT, RenderingT>,
  inputs: Flow<InputT>,
  initialSnapshot: Snapshot? = null,
  beforeStart: CoroutineScope.(
    renderingsAndSnapshots: Flow<RenderingAndSnapshot<RenderingT>>,
    outputs: Flow<OutputT>
  ) -> ResultT
): ResultT
```
See the kdoc on that function for details.

Closes #439.


kotlin/workflow-runtime/src/main/java/com/squareup/workflow/RunWorkflow.kt
```diff
@@ -0,0 +1,140 @@
...
+ * more obvious how to use it.
+ */
+@UseExperimental(ExperimentalCoroutinesApi::class)
+internal typealias Configurator <O, R, T> = CoroutineScope.(
+  renderingsAndSnapshots: Flow<RenderingAndSnapshot<R>>,
+  outputs: Flow<O>
+) -> T
+
+/**
+ * TODO write documentation
```
zach-klippenstein: I should probably write some documentation.

zach-klippenstein: Done.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/LaunchWorkflow.kt
```diff
@@ -0,0 +1,182 @@
...
+ * decides what on what threads to run workflow code.
+ *
+ * This function creates a child [Job] in [scope] and uses it as the parent for the workflow
+ * runtime, and as the job passed to the [beforeStart] function. This means that if [beforeStart]
+ * calls `CoroutineScope.cancel`, it will cancel the workflow runtime, but will not cancel the
+ * [scope] passed into this function.
+ *
+ * @param scope The [CoroutineScope] in which to launch the workflow runtime. Any exceptions thrown
+ * in any workflows will be reported to this scope, and cancelling this scope will cancel the
+ * workflow runtime.
```
rjrjr: [beforeStart] will be called against a child [Job] of this scope.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/LaunchWorkflow.kt
```diff
@@ -0,0 +1,182 @@
...
+ * @param workflow The root workflow to start.
+ * @param inputs Stream of input values for the root workflow. The first value emitted is passed to
+ * the root workflow's [StatefulWorkflow.initialState], and subsequent emissions are passed as
```
rjrjr: _first value_ explanation is a little confusing if root is not stateful, but I don't have a great alternative yet. Seems worth keeping though. Will think harder.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/LaunchWorkflow.kt
```diff
@@ -0,0 +1,182 @@
...
+ * input updates to the root workflow.
+ * @param initialSnapshot If not null, used to restore the workflow.
+ * @param beforeStart Called exactly once with the flows for renderings/snapshots and outputs.
```
rjrjr: Insert after first sentence: Called on a newly created child [Job] of the given [scope], which defines the lifetime of the launched workflow tree. Cancelling that job ends the new workflow session.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/LaunchWorkflow.kt
```diff
@@ -0,0 +1,182 @@
...
+ * Note that the receiver scope will already be cancelled, and the flows will complete immediately
+ * if this function was invoked from a cancelled [scope].
+ *
+ * @return The value returned by [beforeStart].
+ */
+@UseExperimental(ExperimentalCoroutinesApi::class)
+fun <InputT, OutputT : Any, RenderingT, ResultT> launchWorkflowIn(
```
rjrjr: How about s/ResultT/RunnerT/? `ResultT` seems too easy to confuse with `OutputT`.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/Workers.kt
```diff
@@ -37,9 +38,14 @@ import kotlinx.coroutines.selects.SelectBuilder
 @UseExperimental(FlowPreview::class, ExperimentalCoroutinesApi::class)
 internal fun <T> CoroutineScope.launchWorker(worker: Worker<T>): ReceiveChannel<Output<T>> =
   worker.run()
-      // TODO(https://github.com/square/workflow/issues/434) Remove this map to allow operator
-      // fusion to occur.
       .map { Output(it) }
```
rjrjr: You removed the TODO. Did something fix fusion for us?

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/WorkflowLoop.kt
```diff
@@ -27,79 +27,94 @@ import kotlinx.coroutines.flow.Flow
...
-                input = newInput
+          // Tick _might_ return an output, but if it returns null, it means the state or a child
+          // probably changed, so we should re-render/snapshot and emit again.
+          output = select {
+            // Stop trying to read from the inputs channel after it's closed.
+            if (!inputsClosed) {
+              @Suppress("EXPERIMENTAL_API_USAGE")
+              inputsChannel.onReceiveOrNull { newInput ->
+                if (newInput == null) {
+                  inputsClosed = true
```
rjrjr: Have we documented that closing inputs does not end the workflow session?

kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/WorkflowTester.kt
```diff
@@ -58,22 +59,23 @@ import kotlin.coroutines.EmptyCoroutineContext
...
+  private val renderingsAndSnapshotsFlow: Flow<RenderingAndSnapshot<RenderingT>>,
+  private val outputsFlow: Flow<OutputT>
 ) {
 
   private val renderings = Channel<RenderingT>(capacity = UNLIMITED)
   private val snapshots = Channel<Snapshot>(capacity = UNLIMITED)
   private val outputs = Channel<OutputT>(capacity = UNLIMITED)
 
-  internal fun start() {
+  internal fun subscribeToWorkflow() {
```
rjrjr: Nit: should we start saying "collectFrom" instead of "subscribeTo"?

kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/WorkflowTester.kt
```diff
@@ -212,11 +222,23 @@ fun <T, InputT, OutputT : Any, RenderingT>
       snapshot: Snapshot? = null,
       context: CoroutineContext = EmptyCoroutineContext,
       block: WorkflowTester<InputT, OutputT, RenderingT>.() -> T
-    ): T = test(block, context) { factory, inputs ->
-      inputs.offer(input)
-      factory.run(this, inputs.asFlow(), snapshot)
-    }
+    ): T
 // @formatter:on
```
rjrjr: Did you mean to move the `@formatter:on` above the body?

kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/WorkflowTester.kt
```diff
@@ -267,51 +302,3 @@ fun <StateT, OutputT : Any, RenderingT>
       block: WorkflowTester<Unit, OutputT, RenderingT>.() -> Unit
     ) = testFromState(Unit, initialState, context, block)
 // @formatter:on
-
```
rjrjr: 🔥 !

Such a good sign.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/Workers.kt
zach-klippenstein: No, the `catch` operator means we can't get fusion anymore.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/WorkflowLoop.kt
zach-klippenstein: We did at one point. I'll add it to the docs on `launchWorkflowIn`.

- [ ] Document input-closing behavior in `launchWorkflowIn`

kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/WorkflowTester.kt
zach-klippenstein: Yes.

- [ ] s/subscribeToWorkflow/collectFromWorkflow/

kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/WorkflowTester.kt
zach-klippenstein: Yea, would like to minimize the amount of code affected by those things.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/LaunchWorkflow.kt
```diff
@@ -0,0 +1,182 @@
...
+ * you should probably be using a higher-level entry point that will also let you define UI bindings
+ * for your renderings.
+ *
+ * ## Initialization
+ *
+ * Before starting the actual workflow runtime, this function will invoke [beforeStart] and pass
+ * it the [Flow]s of renderings, snapshots, and outputs, as well as a [CoroutineScope] that the
+ * runtime will be hosted in. The workflow runtime will not be started until after this function
+ * returns. This is to allow the output flow to start being collected before any outputs can
+ * actually be emitted. Collectors that start _after_ [beforeStart] returns may not receive outputs
```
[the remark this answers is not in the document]
zach-klippenstein: It does make it difficult, but I want to be really clear about it. Does it make the docs more confusing?

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/LaunchWorkflow.kt
```diff
@@ -0,0 +1,182 @@
...
+ *
+ * The workflow runtime makes use of
+ * [structured concurrency](https://medium.com/@elizarov/structured-concurrency-722d765aa952).
+ * The runtime is started in a specific [CoroutineScope], which defines the context for the entire
+ * workflow tree – most importantly, the [Job] that governs the runtime's lifetime and exception
+ * reporting path, and the [CoroutineDispatcher][kotlinx.coroutines.CoroutineDispatcher] that
+ * decides what on what threads to run workflow code.
+ *
+ * This function creates a child [Job] in [scope] and uses it as the parent for the workflow
+ * runtime, and as the job passed to the [beforeStart] function. This means that if [beforeStart]
```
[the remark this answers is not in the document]
zach-klippenstein: Not a fan of using "receives" as a verb like that – it makes sense grammatically, but I've never seen it used that way and I had to stop and think about it. `beforeStart` is passed a `Job`. That's the important thing this sentence is talking about, the fact it comes in through a `CoroutineScope` receiver is not relevant.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/LaunchWorkflow.kt
```diff
@@ -0,0 +1,182 @@
...
+ * calls `CoroutineScope.cancel`, it will cancel the workflow runtime, but will not cancel the
```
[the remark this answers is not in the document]
zach-klippenstein: I tried that, but it ends up linking to the extension on `CoroutineCancel`, not `CoroutineScope`, and neither of those actually have their own kdoc anyway. I also tried linking to `Job.cancel`, but _that_ links to the deprecated nullary function. Extremely unfortunate that kdoc completely removed javadoc's ability to differentiate between overloads. This issue is filed in https://github.com/Kotlin/dokka/issues/80, but that project seems almost abandoned.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/LaunchWorkflow.kt
zach-klippenstein: Reworded a bit, WDYT?

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/LaunchWorkflow.kt
zach-klippenstein: Agreed about the confusion, `RunnerT` seems at least as good as anything else.

zach-klippenstein:
@rjrjr  Just pushed to address your comments, PTAL.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/LaunchWorkflow.kt
rjrjr: I still find "It also gets a sub-scope" confusing since the scope is `this`, not a parameter. That's what got me talking about _receiver_ elsewhere, although I take your point that that's a weird term too -- this ain't Smalltalk. _Called on_ doesn't work for you?

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/LaunchWorkflow.kt
zach-klippenstein: Gonna leave this in for now, we can always come back later.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/LaunchWorkflow.kt
zach-klippenstein: I'd be curious what the general opinion is on this. I still think of receivers as pretty much regular parameters, with slightly different syntax rules, but I have no reason to think that's a widely-shared mental model.  "Called on" doesn't seem great to me though, since the receiver doesn't know anything about the method call – it makes sense from the _caller's_ perspective, but not from the _implementer's_.

PR #448, opened 2019-07-08
Less hacky lifecycle search in ModalContainer.

rjrjr:


kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ModalContainer.kt
```diff
@@ -261,20 +261,20 @@ private class OnDestroy(private val block: () -> Unit) : LifecycleObserver {
...
  * We keep all this very forgiving because we're just using it to keep some logging
- * noise out of logcat. If someone manages to run this under the wrong type of
- * activity, just return null and let the caller no-op.
+ * noise out of logcat. If someone manages to run this under a strange context whose
+ * [Lifecycle] we can't find, just return null and let the caller no-op.
  */
-tailrec fun Context.activity(): AppCompatActivity? = when {
-  this is Activity -> this as? AppCompatActivity
-  else -> (this as? ContextWrapper)?.baseContext?.activity()
+tailrec fun Context.lifecycle(): Lifecycle? = when {
```
zach-klippenstein: I'm really surprised these don't come out of the box… seems like it would be a really common use case?

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ModalContainer.kt
```diff
@@ -261,20 +261,20 @@ private class OnDestroy(private val block: () -> Unit) : LifecycleObserver {
   fun onDestroy() = block()
 }
 
-private val Dialog.activity: AppCompatActivity?
-  get() = decorView?.context?.activity()
+private val Dialog.lifecycle: Lifecycle?
```
zach-klippenstein: Let's be consistent about whether `lifecycle` is a property or a function. I could go either way – it returns consistently for a given receiver, but also involves a potentially expensive search. But either way, let's be consistent.

rjrjr: Fair enough. Looks like `tailrec` isn't available for property getters, so I'll make them all functions.

zach-klippenstein: Ah, I forgot about that. That's a good enough reason to make them inconsistent, just document it.

rjrjr: Renamed to `lifecycleOrNull()`.

PR #451, opened 2019-07-08
Bump Kotlin to latest version (1.3.41), along with Dokka and detekt.

zach-klippenstein:


kotlin/legacy/legacy-workflow-core/src/test/java/com/squareup/workflow/legacy/CoroutineWorkflowTest.kt
```diff
@@ -132,7 +132,10 @@ class CoroutineWorkflowTest : CoroutineScope {
       throw ExpectedException
     }
     testContext.triggerActions()
-    assertFailsWith<ExpectedException> { workflow.getCompleted() }
+    assertFailsWith<ExpectedException> {
+      @Suppress("IMPLICIT_NOTHING_AS_TYPE_PARAMETER")
+      workflow.getCompleted()
```
zach-klippenstein: This was giving a warning about implicit type parameter `Nothing`, but we don't actually care.

PR #453, opened 2019-07-09
[WIP] Use Flows for TerminalWorkflow.

zach-klippenstein:
Converted the TerminalWorkflowRunner to flows much more directly.

Two commits, one for events and one for inputs.

Given the custom operators I had to write, I'm not sure this makes for so great a sample anymore. But it might serve as a good demo use case for the argument to add these operators to the coroutines library.

### DIFF
--- a/kotlin/samples/hello-terminal/terminal-workflow/src/main/java/com/squareup/sample/helloterminal/terminalworkflow/KeyStroke.kt
+++ b/kotlin/samples/hello-terminal/terminal-workflow/src/main/java/com/squareup/sample/helloterminal/terminalworkflow/KeyStroke.kt
@@ -27,10 +27,15 @@ import com.googlecode.lanterna.input.KeyType.Enter
 import com.squareup.sample.helloterminal.terminalworkflow.KeyStroke.KeyType
 import com.squareup.sample.helloterminal.terminalworkflow.KeyStroke.KeyType.Unknown
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.CoroutineStart.UNDISPATCHED
 import kotlinx.coroutines.ExperimentalCoroutinesApi
-import kotlinx.coroutines.channels.BroadcastChannel
-import kotlinx.coroutines.channels.broadcast
+import kotlinx.coroutines.coroutineScope
+import kotlinx.coroutines.ensureActive
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.isActive
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.suspendCancellableCoroutine
 import com.googlecode.lanterna.input.KeyStroke as LanternaKeystroke
 
 /**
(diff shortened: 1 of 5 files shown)


kotlin/samples/hello-terminal/terminal-workflow/src/main/java/com/squareup/sample/helloterminal/terminalworkflow/ShareOperator.kt
```diff
@@ -0,0 +1,72 @@
...
+ * The flow is collected when the first downstream collector starts, and cancelled after the last
+ * downstream collector is cancelled. To control the dispatcher or other context elements of the
+ * coroutine on which this flow is collected, use the `flowOn` operator before calling `share`.
+ * To control buffering, use the `buffer` operator before calling `share`.
+ *
+ * Should be removed once the
+ * [official share operator](https://github.com/Kotlin/kotlinx.coroutines/issues/1261) lands.
+ */
+@UseExperimental(ExperimentalCoroutinesApi::class, FlowPreview::class)
+internal fun <T> Flow<T>.share(): Flow<T> {
```
zach-klippenstein: Unfortunately, I had to write a custom `share` operator for this, since the keystroke worker is used by two separate workflows simultaneously, and without it they each get alternating keystrokes. There is an issue open on the coroutines project to implement a standard operator (https://github.com/Kotlin/kotlinx.coroutines/issues/1261).

kotlin/samples/hello-terminal/terminal-workflow/src/main/java/com/squareup/sample/helloterminal/terminalworkflow/TerminalSize.kt
```diff
@@ -36,18 +36,25 @@ data class TerminalSize(
...
+    awaitClose { removeResizeListener(resizeListener) }
   }
+      // Conflate first, so it fuses with the channelFlow operator.
+      .conflate()
+      .startWith(terminalSize.toSize())
 
 internal fun LanternaTerminalSize.toSize(): TerminalSize = TerminalSize(rows, columns)
+
+@UseExperimental(ExperimentalCoroutinesApi::class)
+private fun <T> Flow<T>.startWith(value: T): Flow<T> = flow {
```
zach-klippenstein: Had to write a custom `startWith` as well, but this is super trivial and if we merge this I will submit a PR to add it to the coroutines library.

kotlin/samples/hello-terminal/terminal-workflow/src/main/java/com/squareup/sample/helloterminal/terminalworkflow/KeyStroke.kt
```diff
@@ -55,19 +60,72 @@ data class KeyStroke(
...
+}
+
+/**
+ * Sends and interrupt to the calling thread when the [CoroutineScope] is cancelled. Intended to
+ * make cancellation work correctly when [block] performs blocking IO – the blocked thread will not
+ * get the cancellation signal, so we interrupt it to cancel the IO, then catch the interrupt
+ * exception and throw the cancellation exception instead.
+ */
+@ExperimentalCoroutinesApi
+private fun <T> CoroutineScope.interruptThreadOnCancel(block: () -> T): T {
```
zach-klippenstein: This was fun to solve, although I'm not sure how I feel about this solution. The nice thing is that it should work with just about any JVM IO as long as it supports interrupts (I think all of it does?). The gross thing is that it launches a new coroutine for every iteration, but we could get around that using an internal API (see the comments).

kotlin/samples/hello-terminal/terminal-workflow/src/main/java/com/squareup/sample/helloterminal/terminalworkflow/TerminalSize.kt
zach-klippenstein: This implementation is broken. The upstream flow won't get collected until the first emit returns. If the upstream flow is hot (as it is for resizes) and an event occurs while the first emit is suspended, those events will be dropped. Needs to open a channel to the upstream before emitting the initial value.

kotlin/samples/hello-terminal/terminal-workflow/src/main/java/com/squareup/sample/helloterminal/terminalworkflow/KeyStroke.kt
```diff
@@ -55,19 +60,72 @@ data class KeyStroke(
...
-internal fun InputProvider.listenForKeyStrokesOn(
-  scope: CoroutineScope
-): BroadcastChannel<KeyStroke> = scope.broadcast {
-  while (isActive) {
-    val keyStroke = readInput()
-    if (keyStroke.keyType === EOF) {
-      // EOF indicates the terminal input was closed, and we won't receive any more input, so
-      // close the channel instead of sending the raw event down.
-      return@broadcast
+internal fun InputProvider.keyStrokes(): Flow<KeyStroke> = flow {
```
rjrjr: `createKeyStrokesFlow()`?

kotlin/samples/hello-terminal/terminal-workflow/src/main/java/com/squareup/sample/helloterminal/terminalworkflow/KeyStroke.kt
```diff
@@ -55,19 +60,72 @@ data class KeyStroke(
...
+        // close the channel instead of sending the raw event down.
+        break
+      }
+      emit(keyStroke.toKeyStroke())
+    }
+  }
+}
+
+/**
+ * Sends and interrupt to the calling thread when the [CoroutineScope] is cancelled. Intended to
```
rjrjr: s/and/an/

kotlin/samples/hello-terminal/terminal-workflow/src/main/java/com/squareup/sample/helloterminal/terminalworkflow/KeyStroke.kt
rjrjr: They've been bragging so much about how cheap coroutines are, I feel good putting it to the test. Especially in a sample.

kotlin/samples/hello-terminal/terminal-workflow/src/main/java/com/squareup/sample/helloterminal/terminalworkflow/TerminalSize.kt
rjrjr: Are you going to document that or fix it?

kotlin/samples/hello-terminal/terminal-workflow/src/main/java/com/squareup/sample/helloterminal/terminalworkflow/TerminalWorkflowRunner.kt
```diff
@@ -92,41 +90,22 @@ private suspend fun runTerminalWorkflow(
...
-              // Sending that new input invalidated the lastRendering, so we don't want to
-              // re-iterate until we have a new rendering with a fresh event handler. It also
-              // triggered a render pass, so we can just retrieve that immediately.
-              return@onReceive renderings.receive()
-            }
-
-            renderings.onReceive { it }
-          }
-
+  val result = launchWorkflowIn(this, workflow, inputs) { renderingsAndSnapshots, outputs ->
```
rjrjr: def. think that calling these things `runner` rather than `result` is easier to understand.

kotlin/samples/hello-terminal/terminal-workflow/src/main/java/com/squareup/sample/helloterminal/terminalworkflow/TerminalSize.kt
zach-klippenstein: Fix

kotlin/samples/hello-terminal/terminal-workflow/src/main/java/com/squareup/sample/helloterminal/terminalworkflow/ShareOperator.kt
zach-klippenstein: Ray filed as #454

PR #458, opened 2019-07-11
Updates to sdk 29, and tidies dependencies.

rjrjr:
`androidx` deps are now grouped, and `versions` is now consistent: it only
holds shared values, others are inlined (as an increasing number already
were). Benefit, now the IDE nudges us when SDK gets stale.

### DIFF
--- a/kotlin/build.gradle
+++ b/kotlin/build.gradle
@@ -15,45 +15,44 @@
  */
 buildscript {
   ext.versions = [
-      'agp': '3.5.0-beta05',
-      'buildTools': '28.0.3',
-      'compileSdk': '28',
-      'lifecycle': '2.0.0',
-      'minSdk': 21,
-      'targetSdk': 28,
-      'sourceCompatibility': JavaVersion.VERSION_1_8,
-      'targetCompatibility': JavaVersion.VERSION_1_8,
-
-      'coordinators': '0.4',
-      'rxandroid2': '2.1.0',
-      'timber': '4.7.1',
-
-      'detektPlugin': '1.0.0-RC16',
+      'targetSdk': 29,
       'dokka': '0.9.18',
-      'hamcrest': '1.3',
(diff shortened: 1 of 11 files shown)


kotlin/gradle.properties
```diff
@@ -16,6 +16,7 @@
 org.gradle.jvmargs='-Dfile.encoding=UTF-8'
 org.gradle.parallel=true
 android.useAndroidX=true
+#android.debug.obsoleteApi=true
```
zach-klippenstein: Delete?

kotlin/build.gradle
```diff
@@ -15,45 +15,44 @@
...
+      versionCode 1
+      versionName "1.0"
+    }
+  }
+
   ext.deps = [
-      'android_gradle_plugin': "com.android.tools.build:gradle:${versions.agp}",
+      'android_gradle_plugin': "com.android.tools.build:gradle:3.5.0-beta05",
+
+      'androidx': [
```
zach-klippenstein: Thanks! I had made a mental note to do this the other day and forgot to write it down.

kotlin/build.gradle
```diff
@@ -123,23 +123,6 @@ rootProject.ext.defaultDokkaConfig = {
     outputDirectory = "$buildDir/docs"
 }
 
-rootProject.ext.defaultAndroidConfig = {
```
zach-klippenstein: It doesn't look like you inlined this block, or moved it to somewhere – were we just not using it?

kotlin/gradle.properties
rjrjr: Prefer to keep it, I'll add a comment.

kotlin/build.gradle
rjrjr: I moved it toward the top, s.t. the android version #s stayed prominent.

PR #459, opened 2019-07-12
Make the WorkflowRunnerViewModel constructor public so it can be wired up to already-running workflows.

zach-klippenstein:


zach-klippenstein:
Nope, don't need this after all.

PR #460, opened 2019-07-13
Make lifecycleOrNull private.

zach-klippenstein:
It's only used in this file, and we shouldn't pollute the auto-complete space for
`Context` and `Dialog` without really good reason.


kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ModalContainer.kt
```diff
@@ -272,7 +272,7 @@ private val Dialog.decorView: View?
  * noise out of logcat. If someone manages to run this under a strange context whose
  * [Lifecycle] we can't find, just return null and let the caller no-op.
  */
-tailrec fun Context.lifecycleOrNull(): Lifecycle? = when {
+private tailrec fun Context.lifecycleOrNull(): Lifecycle? = when {
```
rjrjr: Oops, thanks.

PR #461, opened 2019-07-13
wip: neato

rjrjr:


kotlin/workflow-runtime/src/main/java/com/squareup/workflow/LaunchWorkflow.kt
```diff
@@ -140,11 +140,11 @@ fun <InputT, OutputT : Any, RenderingT, ResultT> launchSingleOutputWorkflowIn(
     output: Deferred<OutputT>
   ) -> ResultT
 ): ResultT = launchWorkflowIn(scope, workflow, inputs, initialSnapshot) { renderings, outputs ->
-  val workflowScope = this
+//  val workflowScope = this
```
rjrjr: @zach-klippenstein We're nearly there, but there's a teardown race or something going on. If I don't comment out this block, the new `output.await()` call in `WorfklowRunnerViewModel` never fires, and so the call to `finish()` in `HelloWorkflowActivity` is never made, and the app hangs.

rjrjr: Maybe we should just implement this behavior in the view model.

rjrjr: Yeah, that works nicely. I think we'll want to revert the `launchWorkflowIn` PR, just lost our only use case.

zach-klippenstein: Did I miss a test case?

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowRunnerViewModel.kt
```diff
@@ -74,16 +72,27 @@ internal class WorkflowRunnerViewModel<OutputT : Any>(
...
-      WorkflowRunnerViewModel(viewRegistry, renderings, outputs, this) as T
+      WorkflowRunnerViewModel(renderings, output, this) as T
     }
   }
 
-  private val snapshotJob = scope.launch {
-    renderingsFlow
-        .map { it.snapshot }
-        .collect { lastSnapshot = it }
+  private val _result = BehaviorSubject.create<OutputT>()
```
zach-klippenstein: Use `SingleSubject`

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/LaunchWorkflow.kt
zach-klippenstein: Ok, this doesn't actually surprise me. The cancellation does feel pretty racy. It's easy enough to do this manually that providing an API just to do it is redundant and over-engineered anyway.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowRunnerViewModel.kt
zach-klippenstein: Or better yet, `scope.rxSingle`

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowRunnerViewModel.kt
```diff
@@ -74,16 +72,27 @@ internal class WorkflowRunnerViewModel<OutputT : Any>(
...
+  override val result: Single<out OutputT> = _result.firstOrError()
+
+  init {
+    scope.launch {
+      renderingsFlow
+          .map { it.snapshot }
+          .collect { lastSnapshot = it }
+    }
```
zach-klippenstein: Slightly more idiomatic/less nesting:
```kotlin
renderingsFlow.map { it.snapshot }
  .onEach { lastSnapshot = it }
  .launchIn(scope)
```

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowRunnerViewModel.kt
zach-klippenstein: Btw looks like these builders are about to lose their scope receiver:

https://github.com/Kotlin/kotlinx.coroutines/commits/reactive-scopeless

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/LaunchWorkflow.kt
rjrjr: https://github.com/square/workflow/pull/462

Issue #466, opened 2019-07-15
Fix modal back button handling


rjrjr:
The sample PanelContainer doesn't offer back button handling, but we'll need to make that work. This may force us to jump on #250 even though the AndroidX back button stuff is still in alpha. 

I tried putting this on the dialog built in `ModalViewContainer`, but I'm seeing it get called repeatedly for a single back button press. :/

```
          setOnKeyListener { _, keyCode, _ ->
            keyCode == KeyEvent.KEYCODE_BACK && HandlesBack.Helper.onBackPressed(view)
          }
```

PR #467, opened 2019-07-15
WorkflowRunner now delivers single result instead of output stream.

### LINKED ISSUE #430: Workflows started with setContentWorkflow may drop early output emissions.
I'm not thrilled with this extension method any more, but have no change to suggest for this PR. I'll follow up after you merge.

_Originally posted by @rjrjr in https://github.com/square/workflow/pull/426_

rjrjr:
Android's configuration changes make it very difficult to craft an API that
can catch all output values and also avoid double-processing of them, and each
such API we've come up with is very confusing. The cognitive cost  outweighs
the benefit of allowing the Activity or Fragment to consume a stream of
outputs — if that's even a benefit! So far we have no use cases for a stream
v. a single result code that can't be better handled inside the root workflow.

Note that we update the Tic Tac Toe sample to take advantage. We also
(nearly) make it's back button work properly on the initial screen, but
that is blocked by #466.

Closes #430.

### DIFF
--- a/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowRunnerViewModel.kt
+++ b/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowRunnerViewModel.kt
@@ -20,23 +20,19 @@ import androidx.lifecycle.ViewModel
 import androidx.lifecycle.ViewModelProvider
 import com.squareup.workflow.RenderingAndSnapshot
 import com.squareup.workflow.Snapshot
-import com.squareup.workflow.Workflow
 import com.squareup.workflow.launchWorkflowIn
-import io.reactivex.Flowable
 import io.reactivex.Observable
-import kotlinx.coroutines.CoroutineDispatcher
+import io.reactivex.Single
 import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.ExperimentalCoroutinesApi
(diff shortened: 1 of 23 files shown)


kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowRunnerViewModel.kt
```diff
@@ -20,70 +20,64 @@ import androidx.lifecycle.ViewModel
...
+      }
     }
   }
 
-  private val snapshotJob = scope.launch {
+  override val result: Single<out OutputT> = scope
+      .rxSingle { output.first() }
+      .doAfterTerminate {
+        scope.cancel(CancellationException("WorkflowRunnerViewModel delivered result"))
+      }
```
zach-klippenstein: I think we should add a `.cache()` after this, so multiple subscribers to the `Single` don't end up trying to cancel the scope multiple times.

rjrjr:
Nearly there, busted test `WorkflowRunnerViewModelTest#hostCancelledOnCleared` reveals that `Single<OutputT>` is too strict, blows up if we cancel w/no result. Going to replace it with `Maybe<OutputT>`. 

Closing this in the meantime to reduce spam.

rjrjr:
Cool, can't re-open after force push. :/

zach-klippenstein:
@rjrjr In the future, don't worry about closing for noise – if I don't want to get spammed I can always unwatch until it's ready to go.

PR #468, opened 2019-07-16
WorkflowRunner now delivers single result instead of output stream.

### LINKED ISSUE #430: Workflows started with setContentWorkflow may drop early output emissions.

rjrjr:
Android's configuration changes make it very difficult to craft an API that
can catch all output values and also avoid double-processing of them, and each
such API we've come up with is very confusing. The cognitive cost  outweighs
the benefit of allowing the Activity or Fragment to consume a stream of
outputs — if that's even a benefit! So far we have no use cases for a stream
v. a single result code that can't be better handled inside the root workflow.

Note that we update the Tic Tac Toe sample to take advantage. We also
(nearly) make it's back button work properly on the initial screen, but
that is blocked by #466.

Closes #430.

### DIFF
--- a/kotlin/workflow-ui-android/src/test/java/com/squareup/workflow/ui/WorkflowRunnerViewModelTest.kt
+++ b/kotlin/workflow-ui-android/src/test/java/com/squareup/workflow/ui/WorkflowRunnerViewModelTest.kt
@@ -3,16 +3,27 @@ package com.squareup.workflow.ui
 import com.google.common.truth.Truth.assertThat
 import com.squareup.workflow.RenderingAndSnapshot
 import com.squareup.workflow.Snapshot
+import com.squareup.workflow.Workflow
+import com.squareup.workflow.WorkflowAction.Companion.emitOutput
+import com.squareup.workflow.asWorker
+import com.squareup.workflow.onWorkerOutput
+import com.squareup.workflow.stateless
+import com.squareup.workflow.ui.WorkflowRunner.Config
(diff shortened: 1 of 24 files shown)


kotlin/workflow-ui-android/src/test/java/com/squareup/workflow/ui/WorkflowRunnerViewModelTest.kt
```diff
@@ -39,18 +52,83 @@ class WorkflowRunnerViewModelTest {
...
+          onWorkerOutput(outputs.asWorker()) { emitOutput(it) }
+          Unit
+        }
+        .run()
+
+    val tester = runner.result.test()
+
+    tester.assertNotComplete()
+    runBlocking { outputs.send("fnord") }
+    tester.assertComplete()
```
rjrjr: @zach-klippenstein Halp, this assertion fails and I can't figure out why. Especially confusing b/c the less-integrationy `hostCancelledOnResultAndNoSooner` test above passes.

rjrjr:
Forgot to run `build connectedCheck` myself. Last couple of updates fix style issues, broken sample `MainWorkflowTest`.

zach-klippenstein:
I'll merge this as soon as I cut 0.18.0.

PR #59, opened 2018-12-07
Migrate to AndroidX

charbgr:
Per title.

It also opens the way for: https://github.com/square/workflow/issues/51

Closes issue: https://github.com/square/workflow/issues/50, https://github.com/square/workflow/issues/66

PR #469, opened 2019-07-16
Replace support annotations with AndroidX annotations.

zach-klippenstein:
Missed this in #59.

### DIFF
--- a/kotlin/gradle.properties
+++ b/kotlin/gradle.properties
@@ -15,6 +15,7 @@
 #
 org.gradle.jvmargs='-Dfile.encoding=UTF-8'
 org.gradle.parallel=true
+android.enableJetifier=true
 android.useAndroidX=true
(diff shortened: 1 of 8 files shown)


kotlin/gradle.properties
```diff
@@ -15,6 +15,7 @@
 #
 org.gradle.jvmargs='-Dfile.encoding=UTF-8'
 org.gradle.parallel=true
+android.enableJetifier=true
```
zach-klippenstein: Something is still bringing in support annotations, this fixes that.

rjrjr: This might also fix the androidx test rule for fixing the main dispatcher if we try to use it again: I noticed that JetBrains is using a pre-AndroidX jar for some of their integration, which I think was hitting the wrong assertMainThread.

PR #473, opened 2019-07-17
Use single object for noAction instead of creating it on every call

0legg:
`WorkflowAction.noAction()` can be highly utilized across consumers, and it's not depending on any particular input or output types. We can assume having the same object across invocations is safe, and it reduces the memory footprint.

### DIFF
--- a/kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowAction.kt
+++ b/kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowAction.kt
@@ -46,8 +46,9 @@ interface WorkflowAction<StateT, out OutputT : Any> : (StateT) -> Pair<StateT, O
      *
      * Use this to, for example, ignore the output of a child workflow or worker.
      */
+    @Suppress("UNCHECKED_CAST")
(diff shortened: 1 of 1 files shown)


kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowAction.kt
```diff
@@ -100,5 +101,7 @@ interface WorkflowAction<StateT, out OutputT : Any> : (StateT) -> Pair<StateT, O
       output: OutputT
     ): WorkflowAction<StateT, OutputT> =
       WorkflowAction({ "emitOutput($name, $output)" }) { Pair(it, output) }
+
+    private val NO_ACTION = WorkflowAction<Any, Any>({ "noop" }) { Pair(it, null) }
```
zach-klippenstein: Nit: Oops, I forgot to change this to "noAction" – would you mind doing that here right away? But it's not a regression, so I'm also fine merging this as-is.

0legg: 👍

PR #475, opened 2019-07-18
Fix WorkflowPool to work with ConflatedBroadcastChannel.

zach-klippenstein:
Before, the renderings stream came from a regular `Channel`, and the replay/caching
happened downstream. In the legacy integration code, this meant that when a V2 workflow emitted
an output and got cancelled, the pool would immediately see the renderings channel was closed on the next
`onWorkflowUpdate` and emit `Finished`. Now that it's backed by a `ConflatedBroadcastChannel`, it will
always emit the last-emitted value on subscription before sending the close signal. This means the
pool will always see the last state, and emit `Running`, which means the v2 workflow just gets restarted
on the next iteration. Fixed by explicitly checking the `isActive` flag on the workflow's job before
emitting `Running`.


zach-klippenstein:
Note this is merging into the release branch for 0.18.0.

PR #485, opened 2019-07-19
Update to Xcode 10.2/Swift 5.0

timdonnelly:
@davidapgar @bencochran are we ready to move to Swift 5?

Feels great to get rid of all the `import Result` lines!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,7 @@ gen/
 # Swift Package Manager
 .build/
 Package.resolved
+.swiftpm/
 
 # CocoaPods
 Pods/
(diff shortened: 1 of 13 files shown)


davidapgar:
Ideally should stay at ReactiveSwift 5.0.0, for backward compatibility with Swift 4.2 consumer code (which supports both Swift 4.2 and Swift 5). `Result` is also compatible across both.

swift/Samples/SampleApp/Podfile.lock
```diff
@@ -1,25 +1,21 @@
 PODS:
-  - ReactiveSwift (5.0.0):
+  - ReactiveSwift (5.0.1):
```
davidapgar: This is gone on master, so you'll have to rebase to not accidentally re-add it.

davidapgar:
FYI @timdonnelly I rebased this so it can merge cleanly. Will merge when builds go green. Thanks again!

PR #490, opened 2019-07-19
DNM: Java 7 to 8, and converts non-legacy jvm tests to kotlintest.

rjrjr:
So this is what kotlintest looks like. I'm a bit freaked out for a couple of reasons:

 * `MainWorkflowTest` and a couple of others pass in the command lie but fail in the IDE with `NoSuchMethodError CompletableJob`
 * A couple of tests in `WorkflowRunnerViewModelTest` flat out fail now that really shouldn't have changed.

I'll look a little harder on Monday, but I'm wary.

### DIFF
--- a/kotlin/build.gradle
+++ b/kotlin/build.gradle
@@ -17,6 +17,7 @@ buildscript {
   ext.versions = [
       'targetSdk': 29,
       'dokka': '0.9.18',
+      'junit5': '5.5.0',
       'kotlin': '1.3.41',
       'kotlinCoroutines': '1.3.0-M2',
   ]
@@ -75,6 +76,7 @@ buildscript {
               'common': "org.jetbrains.kotlin:kotlin-test-common",
               'annotations': "org.jetbrains.kotlin:kotlin-test-annotations-common",
               'jdk': "org.jetbrains.kotlin:kotlin-test-junit",
+              'kotlintest': "io.kotlintest:kotlintest-runner-junit5:3.3.3",
               'mockito': "com.nhaarman:mockito-kotlin-kt1.1:1.5.0"
           ]
       ],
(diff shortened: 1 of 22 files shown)


kotlin/build.gradle
```diff
@@ -75,6 +76,7 @@ buildscript {
               'common': "org.jetbrains.kotlin:kotlin-test-common",
               'annotations': "org.jetbrains.kotlin:kotlin-test-annotations-common",
               'jdk': "org.jetbrains.kotlin:kotlin-test-junit",
+              'kotlintest': "io.kotlintest:kotlintest-runner-junit5:3.3.3",
```
zach-klippenstein: Are there no common modules we can use for `workflow-core`, `workflow-runtime`, etc?

kotlin/samples/tictactoe/common/src/test/java/com/squareup/sample/gameworkflow/PlayerInfoTest.kt
```diff
@@ -15,14 +15,15 @@
...
-class PlayerInfoTest {
-  @Test fun readWriteTurn() {
+class PlayerInfoTest : StringSpec({
+  "read write turn" {
     val before = PlayerInfo("able", "baker")
     val out = before.toSnapshot()
     val after = PlayerInfo.fromSnapshot(out.bytes)
-    assertThat(after).isEqualTo(before)
+
+    after shouldBe before
```
zach-klippenstein: 🤔🤔🤔

kotlin/samples/tictactoe/common/src/test/java/com/squareup/sample/gameworkflow/TakeTurnsWorkflowTest.kt
```diff
@@ -25,31 +24,31 @@ import com.squareup.sample.gameworkflow.Player.O
...
         TakeTurnsInput.newGame(PlayerInfo("higgledy", "piggledy"))
     ) {
       val (x, o) = awaitNextRendering().playerInfo
 
-      assertThat(x)
-          .isEqualTo("higgledy")
-      assertThat(o)
-          .isEqualTo("piggledy")
+      x shouldBe "higgledy"
+      o shouldBe "piggledy"
```
zach-klippenstein: I do appreciate that this reads like a children's book now.

rjrjr:
I like this even less than you do. I really don't see any benefit to the Junit4 to Junit5 switch. For me it offers a single interesting feature, the ability autoload Extensions, and I can't get that to work. Other than it's just a case of "new must be better," which I simply don't believe.

And neither Junit5 nor kotlintest get us any closer to a decent solution for cross platform testing. I think they're distractions.

kotlin/build.gradle
rjrjr: Nope. It builds strictly on Junit5.

zach-klippenstein:
The only other advantage I can think of off the top of my head is better support for parameterized tests, but we don't have an immediate need for that anyway.

PR #494, opened 2019-07-23
Publish Swift API docs as well, re-org the nav menu.

zach-klippenstein:


.buildscript/build_swift_docs.sh
```diff
@@ -0,0 +1,57 @@
...
+# gem install bundler jazzy
+
+# Usage: build_swift_docs SHA REPO_URL_OR_PATH OUTPUT_DIR SPACE_SEPARATED_SCHEMES
+function build_swift_docs() {
+    SHA=$1
+    REPO="$2"
+    OUTPUT_DIR="$(realpath "$3")"
+    SCHEMES="$4"
+    ORIG_DIR="$(pwd)"
+    DIR=swiftdocs-clone
```
rjrjr: `/tmp/swiftdocs-clone`?

.buildscript/build_swift_docs.sh
```diff
@@ -0,0 +1,57 @@
...
+        jazzy \
+            --xcodebuild-arguments "-scheme,$scheme,-workspace,SampleApp.xcworkspace" \
+            --author Square \
+            --author_url https://developer.squareup.com \
+            --github_url https://github.com/square/workflow \
+            --output "$OUTPUT_DIR/$scheme"
+    done
+
+    # Delete our temp folder.
+    cd "$ORIG_DIR"
```
rjrjr: Could just use pushd and popd

deploy_website.sh
```diff
@@ -22,11 +22,19 @@
...
+#
+# Usage deploy_website.sh SHA_OR_REF_TO_DEPLOY
 
 set -ex
 
+# Get the actual SHA, even if a ref was passed.
+SHA=$(git rev-parse $1)
 REPO="git@github.com:square/workflow.git"
-DIR=temp-clone
+DIR=mkdocs-clone
```
rjrjr: /tmp?

rjrjr: You `cd ..` in this script v. `cd $ORIG_DIR` in the other one. Be good to be consistent, and again I'd suggest pushd / popd.

mkdocs.yml
```diff
@@ -57,15 +57,21 @@ nav:
...
+  - 'API Reference':
+    - 'Kotlin':
       - 'workflow-core': 'kotlin/api/workflow-core/com.squareup.workflow/index.md'
       - 'workflow-runtime': 'kotlin/api/workflow-runtime/com.squareup.workflow/index.md'
       - 'workflow-ui-core': 'kotlin/api/workflow-ui-core/com.squareup.workflow.ui/index.md'
       - 'workflow-ui-android': 'kotlin/api/workflow-ui-android/com.squareup.workflow.ui/index.md'
       - 'workflow-testing': 'kotlin/api/workflow-testing/com.squareup.workflow.testing/index.md'
       - 'workflow-rx2': 'kotlin/api/workflow-rx2/com.squareup.workflow.rx2/index.md'
+    - 'Swift':
+      # mkdocs will warn about these not existing, that's fine, deploy_website.sh manually adds them.
```
rjrjr: Perhaps echo this reassurance before the warning shows up.

zach-klippenstein: I can't echo it from here, and I don't want to duplicate documentation everywhere. If you try fixing it, you'll see this comment.

deploy_website.sh
zach-klippenstein: Moved to pushd/popd.

.buildscript/build_swift_docs.sh
zach-klippenstein: Easier to troubleshoot if something fails if you just stay in the same directory tree.

.buildscript/build_swift_docs.sh
```diff
@@ -0,0 +1,56 @@
...
+# It requires ruby, bundler (2.x), CocoaPods, and Jazzy to build.
+# gem install bundler jazzy
+
+# Usage: build_swift_docs SHA REPO_URL_OR_PATH OUTPUT_DIR SPACE_SEPARATED_SCHEMES
+function build_swift_docs() {
+    local SHA=$1
+    local REPO="$2"
+    local OUTPUT_DIR="$(realpath "$3")"
+    local SCHEMES="$4"
+    local DIR=swiftdocs-clone
```
zach-klippenstein: These were overwriting the variables in the main deploy script because they weren't local. Bash is great.

davidapgar: (only partially trolling) Port this to python, ruby, perl, or <insert your favorite scripting language>, since any of them are generally safer to write correctly than bash?

zach-klippenstein: Yea, Python would be good. Not gonna do that right now though.

PR #495, opened 2019-07-23
Add analytics to docs site.

zach-klippenstein:


mkdocs.yml
```diff
@@ -77,3 +77,8 @@ nav:
   - 'Change Log ⏏': CHANGELOG.md
   - 'Contributing': CONTRIBUTING.md
   - 'Code of Conduct': CODE_OF_CONDUCT.md
+
+# Google Analytics. Add export WORKFLOW_GOOGLE_ANALYTICS_KEY="UA-XXXXXXXXX-X" to your ~/.bashrc
```
rjrjr: What happens if we haven't done so?

zach-klippenstein: I will add an error to the deploys cript.

zach-klippenstein: Added.

Issue #443, opened 2019-07-05
Relax restriction on using a stale `sink`


davidapgar:
The most recent update to concurrency in Workflow made sink's very strict, in only accepting a single `Action` per render pass.

Unfortunately, a lot of consumers on the UI side do not synchronously update their callbacks which will result in actions being sent to invalid sinks for fast user input (see https://github.com/square/Blueprint/issues/13 for the example in Blueprint).

This restriction should be relaxed, piping old sink's action into the new sink created in each render pass.

Additionally, should introduce a "strict" sink that has the same current restrictions.

zach-klippenstein:
What if the new rendering doesn't have a sink?

davidapgar:
If the new rendering does not declare the same type of sink, then it will still fail with a hard assert.

Issue #351, opened 2019-05-16
Discuss: Allow "input" to update for a workflow in a `ContainerViewController`/`WorkflowHost`


davidapgar:
When rendering a child workflow, the "input" (ie: the `Workflow`) is updated each time (allowing it to handle an update to its input, via `workflowDidChange`).

Currently, the root workflow hosted in a `WorkflowHost` (and `ContainerViewController`) does not provide any means to update the `Workflow` - it is just the root workflow, so the outside doesn't have a path to update the input.

We should consider adding this (I believe the Android version *does* have support for this)

zach-klippenstein:
We do, our `WorkflowHost` takes a `ReceiveChannel` (`Signal`) of inputs. Super handy because it also lets the first input be provided asynchronously, which turns out to be a pretty common use case given the way our internal integration works.

PR #496, opened 2019-07-23
[Swift] Allow reuse of sinks across render passes

davidapgar:
The UI receiving the screen or rendering from a workflow *should* synchronously update to use the newest sinks and closures on every render pass.

However, some declarative UI *may* not always update synchronously (for instance, will wait for the next render pass) which for very rapid UI originated events may result in an old sink being reused.

This change relaxes the strictness internally - as long as a subsequent render pass declares a sink of the same action type, it will be "reused" and thus proxy events from the prior sink through. If a sink of the same type is *not* declared, it will still be an error to send an action to a stale sink. This does not change the restriction for there only being a single action sent to a sink for each render pass.

Additionally, adds and `update` to `WorkflowHost` to allow the input to change (#351) as this facilitated some of the test cases (and is a current gap between the Swift and Kotlin implementations). Providing an update to the input in `ContainerViewController` will come in a followup PR.

Issue #443

### DIFF
--- a/swift/Workflow/Sources/SubtreeManager.swift
+++ b/swift/Workflow/Sources/SubtreeManager.swift
@@ -13,6 +13,9 @@ extension WorkflowNode {
         /// Sinks from the outside world (i.e. UI)
         private var eventPipes: [EventPipe] = []
 
+        /// Reusable sinks from the previous render pass
+        private var previousSinks: [ObjectIdentifier:AnyReusableSink] = [:]
+
         /// The current array of children
         private (set) internal var childWorkflows: [ChildKey:AnyChildWorkflow] = [:]
 
@@ -34,6 +37,7 @@ extension WorkflowNode {
 
             /// Create a workflow context containing the existing children
             let context = Context(
+                previousSinks: previousSinks,
                 originalChildWorkflows: childWorkflows,
                 originalChildWorkers: childWorkers)
 
@@ -52,8 +56,12 @@ extension WorkflowNode {
             /// Merge all of the signals together from the subscriptions.
             self.subscriptions = Subscriptions(eventSources: context.eventSources, eventPipe: EventPipe())
 
+            /// Captured the reusable sinks from this render pass.
+            self.previousSinks = context.sinkStore.usedSinks
+
             /// Capture all the pipes to be enabled after render completes.
             self.eventPipes = context.eventPipes
+            self.eventPipes.append(contentsOf: context.sinkStore.eventPipes)
             self.eventPipes.append(self.subscriptions.eventPipe)
 
             /// Set all event pipes to `pending`.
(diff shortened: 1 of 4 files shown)


swift/Workflow/Sources/WorkflowHost.swift
```diff
@@ -53,6 +53,19 @@ public final class WorkflowHost<WorkflowType: Workflow> {
 
     }
 
+    /// Update the input for the workflow. Will cause a render pass.
+    public func update(workflow: WorkflowType) {
+        rootNode.update(workflow: workflow)
+
+        // Treat the update as an "output" from the workflow as an external event to force a render pass.
```
zach-klippenstein: Nit: "as an" twice – intentional? Not sure how to parse.

zach-klippenstein: Does this mean that the host emits an output every time the root workflow "input" is updated?

davidapgar: Whoops. Will update to hopefully make this more readable.

davidapgar: At the `WorkflowHost` level, it "outputs" every time something in the tree changes. Sometimes that's an actual output with the `Output` type on the workflow, otherwise it's an event that includes the debugging information and is the signal that a render pass needs to occur.

swift/Workflow/Sources/SubtreeManager.swift
```diff
@@ -227,6 +240,70 @@ extension WorkflowNode.SubtreeManager {
...
+
+    fileprivate class AnyTypedSink {
+        var eventPipe: EventPipe
+
+        init() {
+            eventPipe = EventPipe()
+        }
+    }
+
+    fileprivate final class TypedSink<Action: WorkflowAction>: AnyTypedSink where Action.WorkflowType == WorkflowType {
```
bencochran: Maybe `ReusableSink` rather than `TypedSink`

swift/Workflow/Sources/SubtreeManager.swift
```diff
@@ -227,6 +240,70 @@ extension WorkflowNode.SubtreeManager {
...
+        func eventPipes() -> [EventPipe] {
+            let eventPipes = usedSinks.values.map { typedSink -> EventPipe in
+                typedSink.eventPipe
+            }
+
+            return eventPipes
+        }
+    }
+
+    fileprivate class AnyTypedSink {
```
bencochran: Consider a `/// Type-erased base class` kind of comment

swift/Workflow/Sources/SubtreeManager.swift
```diff
@@ -227,6 +240,70 @@ extension WorkflowNode.SubtreeManager {
...
+                // Create a new typed sink.
+                typedSink = TypedSink<Action>()
+            }
+
+            usedSinks[key] = typedSink
+
+            return typedSink
+        }
+
+        func eventPipes() -> [EventPipe] {
```
bencochran: nit: consider `var eventPipes: [EventPipe]`?

swift/Workflow/Sources/SubtreeManager.swift
davidapgar: Good call on the renaming, must better description of what is does, not how it happens to be used internally.

PR #497, opened 2019-07-23
Remove podfile.lock from sample app.

davidapgar:
Since it uses the local version numbers, it's
out of sync every time a release is done. Less
noise this way (and the Podfile configuration should
keep the versions such that it will build)


timdonnelly:
Ah, good call.

Also, remember that we can move all of this to Swift Package Manager once we upgrade to Xcode 11 later this year.

PR #498, opened 2019-07-23
Add analytics to swift API docs.

zach-klippenstein:


zach-klippenstein:
Deployed from this SHA and verified that it's working.

PR #516, opened 2019-08-02
Better errors for render tester convenience method

davidapgar:
Plumbs the file and line to convenience render tester method.

This will cause errors to be reported at the calling site (in the test case) instead of within the library (was missed on the original addition)


zach-klippenstein:
So… jealous…

PR #517, opened 2019-08-06
Give TicTacToe sample modules unique names.

zach-klippenstein:
Gradle requires all project (module) names to be unique. Project names are not
namespaced to their parent, meaning only one project can be named, e.g., "common",
in the entire "build". All the other sample modules have their sample name prepended
to their modules, so this makes the tictactoe sample consistent with that pattern.


zach-klippenstein:
I'm super dumb, this is a maven config issue that we can fix, not a gradle issue.

PR #529, opened 2019-08-09
Simple dungeon crawler sample app.

zach-klippenstein:
Screenshot_20190808-173322


kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/GameWorkflow.kt
```diff
@@ -0,0 +1,244 @@
...
+  /**
+   * @param board Should not change while the game is running.
+   */
+  data class Input(
+    val board: Board,
+    val ticksPerSecond: Int = 15
+  )
+
+  /**
+   * @param finishedSnapshot If non-null, the game is finished and this was the last rendering
```
rjrjr: No such param

kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/GameWorkflow.kt
```diff
@@ -0,0 +1,244 @@
...
+        aiLocations = aiWorkflows.map { random.nextEmptyLocation(board) }
+    ))
+  }
+
+  override fun onInputChanged(
+    old: Input,
+    new: Input,
+    state: State
+  ): State {
+    check(old.board == new.board) { "Expected board to not change during the game." }
```
rjrjr: nit: `require` (payback, man)

kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/GameWorkflow.kt
```diff
@@ -0,0 +1,244 @@
...
+  }
+
+  override fun render(
+    input: Input,
+    state: State,
+    context: RenderContext<State, Output>
+  ): GameRendering {
+    val running = !state.game.isPlayerEaten
+    // Stop actors from ticking if the game is paused or finished.
+    val ticker = if (running) createTickerWorker(input.ticksPerSecond) else PAUSED_TICKER
```
rjrjr: nit: seems odd to create the ticker worker over and over again. Seems like it could be a private var on the input, s.t. it only changes if the tick rate actually changes. Something like that. (Or maybe that's over engineering even for us.)

kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/GameWorkflow.kt
```diff
@@ -0,0 +1,244 @@
...
+    val playerRendering = context.renderChild(playerWorkflow, playerInput)
+
+    // Render all the other actors.
+    val aiRenderings = aiWorkflows.zip(game.aiLocations)
+        .mapIndexed { index, (aiWorkflow, aiLocation) ->
+          val aiInput = ActorInput(board, aiLocation, ticker)
+          aiLocation to context.renderChild(aiWorkflow, aiInput, key = index.toString())
+        }
+
+    // If the game is paused or finished, just render the board without ticking.
```
rjrjr: nit: this kind of thing is a little easier to read if the trivial case is first. That is, `if (!running)…` — makes your comment easier to follow too.

kotlin/samples/dungeon/README.md
```diff
@@ -0,0 +1,12 @@
+# Dungeon Crawler Sample
+
+This sample app is a simple 2D dungeon crawler game.
+
+The game logic is all in the `common` module. The `app` module contains a simple Android app
+that runs the game.
+
+The core game logic is all defined by `GameWorkflow`. Each AI and player actor is implemented as a
+child workflow of `GameWorkflow`. Each actor workflow renders its avatar, and outputs directions
+to move. Actor workflows inject a `GameTicker` that they use to perform movement on a regular
```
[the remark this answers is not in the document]
rjrjr: Didn't you change this s.t. the actors no longer output their direction, instead it's part of their rendering?

kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/GameWorkflow.kt
```diff
@@ -0,0 +1,244 @@
...
+        // Check if AI captured player.
+        return@onWorkerOutput if (newGame.isPlayerEaten) {
+          enterState(
+              state.copy(game = newGame),
+              emittingOutput = PlayerWasEaten
+          )
+        } else {
+          enterState(
+              state.copy(game = newGame),
+              emittingOutput = output
```
rjrjr: Worth noting that null output means no output.

kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/GameWorkflow.kt
```diff
@@ -0,0 +1,244 @@
...
+  // Don't let the player leave the board.
+  x = x.coerceIn(0 until board.width)
+  // Don't allow collisions with obstacles on the board.
+  if (!board[x, y].isEmpty) {
+    collisionDetected = true
+    x = this.x
+  }
+
+  if (UP in movement) y -= 1
+  if (DOWN in movement) y += 1
```
rjrjr: Oh, neat -- hadn't noticed how the opposite buttons cancel each other.

kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/GameWorkflow.kt
zach-klippenstein: We already do this anywhere you use `asWorker`. We could do that but then we'd need to store the last `ticksPerSecond` or the worker itself in the state. This is less code.

EtherBit:
Nice

PR #533, opened 2019-08-10
Create a Worker.transform operator.

zach-klippenstein:


kotlin/workflow-core/src/main/java/com/squareup/workflow/Worker.kt
```diff
@@ -36,6 +37,8 @@ import kotlin.reflect.KClass
 /**
  * Represents a unit of asynchronous work that can have zero, one, or multiple outputs.
  *
+ * [Worker]s are simply [Flow]s that can be compared to determine equivalence.
```
rjrjr: I think you mean _effectively_. :trollface:

kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/AiWorkflow.kt
```diff
@@ -95,12 +96,12 @@ private fun <T : Enum<T>> Random.nextEnum(enumClass: KClass<T>): T {
...
 private fun Worker<Long>.createDirectionTicker(random: Random): Worker<Unit> =
-  Worker.create(key = UUID.randomUUID().toString()) {
-    run()
-        .collect { tick ->
-          if (tick % random.nextInt(2, 5) == 0L) {
-            emit(Unit)
-          }
-        }
+  transform {
+    it.transform { tick ->
```
rjrjr: Maybe more readable if you s/it/flow/

kotlin/workflow-core/src/main/java/com/squareup/workflow/Worker.kt
```diff
@@ -184,6 +187,11 @@ interface Worker<out T> {
       block: suspend () -> Unit
     ): Worker<Nothing> = TypedWorker(Nothing::class, key, flow { block() })
 
+    /**
+     * Returns a [Worker] that finishes immediately without emitting anything.
+     */
+    fun finished(): Worker<Nothing> = FinishedWorker
```
rjrjr: Unused? Should at least be covered by a smoke test.

zach-klippenstein: Everything needs unit tests here.

zach-klippenstein: It's not unused, see GameWorkflow above, line 105.

I'm wondering if this should be a function though, or we should just expose it as a val. No need to do any casting here so no requirement to have it as a function.

rjrjr: +1 to making it a val

zach-klippenstein: I'm going to keep it a function to be consistent with what Kotlin does: `emptyFlow()` is a function. I'm also going to give it a generic, so you can say `Worker.finished<String>()`.

https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/empty-flow.html

zach-klippenstein:
Added unit tests.

PR #537, opened 2019-08-15
Adds Kotlin Sink, makeActionSink, makeEventSink.

rjrjr:
This PR introduces:

 * a Kotlin `Sink` class matching Swift's
 * `RenderContext.makeActionSink` matching Swift's `makeSink(ofActionType)`
 * `RenderContext.makeEventSink` matching Swift's `makeSink(ofEventType)`

and deprecates:

 * `WorkflowAction.enterState`
 * `WorkflowAction.emitOutput`
 * `WorkflowAction.modifyState`
 * `RenderContext.onEvent`
 * `EventHandler`

Kotlin apps are now able to make custom WorkflowAction class hierarchies in
the style common in Swift apps. This is made possible by the new
`WorkflowAction.apply` method, analagous to the same method in Swift.
Where the Swift version has an `inout StateT` function argument,
Kotlin's is called against a `Mutator<StateT>` receiver.

This allows us to implement action types like the following, where the null
returns indicate that no output is to be emitted.

```
  sealed class Action : WorkflowAction<Movement, Nothing> {
    class StartMoving(private val direction: Direction) : Action() {
      override fun Mutator<Movement>.apply(): Nothing? {
        state += direction
        return null
      }
    }

    class StopMoving(private val direction: Direction) : Action() {
      override fun Mutator<Movement>.apply(): Nothing? {
        state -= direction
        return null
      }
    }
  }
```

The convention is to declare these `Action` types inside workflows, instead of
`Event` type trees in rendering classes. And instead of creating `EventHandler`
instances to be used by renderings, we create `Sink` instances:

```
  override fun render(
    input: TakeTurnsInput,
    state: Turn,
    context: RenderContext<Turn, CompletedGame>
  ): GamePlayScreen {
    val sink = context.makeActionSink<Action>()

    return GamePlayScreen(
        playerInfo = input.playerInfo,
        gameState = state,
        onQuit = { sink.send(Quit) },
        onClick = { row, col -> sink.send(TakeSquare(row, col)) }
    )
  }
```


kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/PlayerWorkflow.kt
```diff
@@ -57,15 +65,15 @@ class PlayerWorkflow(
     state: Movement,
     context: RenderContext<Movement, Nothing>
   ): Rendering {
+    val sink = context.onEvent<Action> {
+      it.state = state
```
zach-klippenstein: Isn't this already initialized to the current state?

rjrjr:
Not in this proof of concept, but it will be.

kotlin/workflow-core/src/main/java/com/squareup/workflow/Sink.kt
```diff
@@ -0,0 +1,34 @@
...
+ *
+ *    sink.contraMap(transform).send(value)
+ *
+ *  **Trivia**: Why is this called `contraMap`?
+ *     - `map` turns `Type<T>` into `Type<U>` via `(T)->U`.
+ *     - `contraMap` turns `Type<T>` into `Type<U>` via `(U)->T`
+ *
+ * Another way to think about this is: `map` transforms a type by changing the
+ * output types of its API, while `contraMap` transforms a type by changing the
+ *  *input* types of its API.
```
zach-klippenstein: Nit: extra space at beginning of line

kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/PlayerWorkflow.kt
```diff
@@ -34,17 +34,27 @@ class PlayerWorkflow(
...
-  sealed class Event {
-    data class StartMoving(val direction: Direction) : Event()
-    data class StopMoving(val direction: Direction) : Event()
+  sealed class Action : WorkflowAction<Movement, Nothing> {
+    class StartMoving(val direction: Direction) : Action()
+
+    class StopMoving(val direction: Direction) : Action()
+
+    override fun Mutator<Movement>.apply(): Nothing? {
+      when (this@Action) {
```
zach-klippenstein: Ugh this is gross, didn't think about how this would interact with the receiver. Don't have any better ideas though

kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/PlayerWorkflow.kt
```diff
@@ -57,15 +67,12 @@ class PlayerWorkflow(
...
     return Rendering(
         actorRendering = ActorRendering(avatar = avatar, movement = state),
-        onEvent = context.onEvent { event ->
-          val newMovement = when (event) {
-            is StartMoving -> state + event.direction
-            is StopMoving -> state - event.direction
-          }
-          enterState(newMovement)
-        })
+        onEvent = sink
```
zach-klippenstein: So I thought the swift style here was more like:

```
onEvent = { sink.send(it) }
```

Which keeps the screen types as pure functions. Don't know why they don't do this though. 

@apgar @benconchran?

kotlin/samples/tictactoe/common/src/main/java/com/squareup/sample/gameworkflow/TakeTurnsWorkflow.kt
```diff
@@ -64,32 +86,15 @@ class RealTakeTurnsWorkflow : TakeTurnsWorkflow,
...
     return GamePlayScreen(
         playerInfo = input.playerInfo,
         gameState = state,
-        onEvent = context.onEvent { event ->
-          when (event) {
-            is TakeSquare -> onTakeSquare(state, event)
-            Quit -> emitOutput(CompletedGame(Quitted, state))
-          }
-        }
+        onQuit = { sink.send(Quit) },
```
zach-klippenstein: Oh you're doing that here, why the variation?

kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowAction.kt
```diff
@@ -54,54 +62,70 @@ interface WorkflowAction<StateT, out OutputT : Any> : (StateT) -> Pair<StateT, O
...
     ): WorkflowAction<StateT, OutputT> =
-      WorkflowAction({ "emitOutput($name, $output)" }) { Pair(it, output) }
+      WorkflowAction({ "emitOutput($name, $output)" }) { output }
 
-    private val NO_ACTION = WorkflowAction<Any, Any>({ "noAction" }) { Pair(it, null) }
+    private val NO_ACTION = WorkflowAction<Any, Any>({ "noAction" }) { null }
   }
 }
+
+fun <StateT, OutputT : Any> WorkflowAction<StateT, OutputT>.fire(
```
zach-klippenstein: `applyTo`?

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/RealRenderContext.kt
```diff
@@ -67,15 +68,25 @@ class RealRenderContext<StateT, OutputT : Any>(
...
-      ): ChildRenderingT {
-    // @formatter:on
+  override fun <A : WorkflowAction<StateT, OutputT>> makeSink(): Sink<A> {
+    checkNotFrozen()
+
+    return object : Sink<A> {
+      override fun send(value: A) {
+        // If this returns false, we lost the race with another event being sent.
+        check(nextUpdateFromEvent.complete(value)) {
+          "Expected to successfully deliver $value. Are you using an old rendering?"
```
zach-klippenstein: Don't know why I didn't do this initially, but I think it would be helpful for debugging to include the deferred itself in this message, to see what we lost to.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/WorkflowNode.kt
```diff
@@ -119,7 +120,7 @@ internal class WorkflowNode<InputT, StateT, OutputT : Any, RenderingT>(
     handler: (OutputT) -> T?
   ) {
     fun acceptUpdate(action: WorkflowAction<StateT, OutputT>): T? {
-      val (newState, output) = action(state)
+      val (newState, output) = action.fire(state)
```
zach-klippenstein: Could save a Pair allocation by creating the Mutator here directly to close over the state property. This prevents any leaks of the sink from surprisingly mutating the property though so the safety is nice.

swift/Workflow/Sources/Sink.swift
```diff
@@ -1,4 +1,5 @@
-/// Sink is a type that receives incoming values (commonly events or actions)
+/// Sink is a type that receives incoming values (commonly events or `WorkflowAction`)
+/// Use `RenderContext.makeSink` to create instances.
```
zach-klippenstein: Nit: I think it is conventional in swift docs to put a blank line between these.

kotlin/samples/tictactoe/common/src/main/java/com/squareup/sample/gameworkflow/TakeTurnsWorkflow.kt
rjrjr: Because I'm not done yet and it's too soon to review.

kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/PlayerWorkflow.kt
rjrjr: I was just slavishly doing the Swift thing — this is their only option. In another spot I have each action implement it. Leagues better, will adjust this.

kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/PlayerWorkflow.kt
rjrjr: Not done yet, day is very fragmented. I'll make noise when the dust settles.

kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/PlayerWorkflow.kt
rjrjr: It's a little verbose, but super easy to type thanks to the IDE, and easy to read.

```
  sealed class Action : WorkflowAction<Movement, Nothing> {
    class StartMoving(val direction: Direction) : Action() {
      override fun Mutator<Movement>.apply(): Nothing? {
        state += direction
        return null
      }
    }

    class StopMoving(val direction: Direction) : Action() {
      override fun Mutator<Movement>.apply(): Nothing? {
        state -= direction
        return null
      }
    }
  }
```

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/RealRenderContext.kt
rjrjr: https://github.com/square/workflow/issues/539

swift/Workflow/Sources/Sink.swift
rjrjr: Do you mean btw the comment and the code? Not seeing that.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/WorkflowNode.kt
rjrjr: I'm inclined to err on the side of safety. That's why I like the `Mutator` API in general. We can always optimize if this somehow turns out to be a hot spot.

kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/PlayerWorkflow.kt
rjrjr: Backed away from this and went back to `when (this@Action)`. It's really the smallest amount of boilerplate. 

Main wart is occasional cast like `val oldState = state`, but that pretty consistently seems like a clue that a workflow is too big and should be factored down into screen-sized bites.

swift/Workflow/Sources/Sink.swift
zach-klippenstein: Summary and description (just like java/kdoc)

kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowAction.kt
```diff
@@ -13,33 +13,52 @@
...
+   *
+   * @return an output value for the workflow to emit, or null to emit nothing.
+   */
+  fun Mutator<StateT>.apply(): OutputT?
 
   companion object {
 
     /**
-     * Creates a [WorkflowAction] from the [update] lambda.
+     * Convenience to create a [WorkflowAction] from the [apply] lambda.
```
rjrjr: @davidapgar @bencochran @timdonnelly I don't believe that Swift has an analog to this, and I'm curious to hear your reactions to it. 

I've been pretty happy with how it rolled out across the Kotlin demos (I'll highlight a few). It allows us some flexibility when full fledged hierarchy of action types are more trouble than they're worth, especially in spots where it allowed discrete action functions to get the job done more concisely.

kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/AiWorkflow.kt
```diff
@@ -78,7 +78,10 @@ class AiWorkflow(
         LEFT -> UP
       }
       println("AI changing direction from ${state.direction} to $newDirection")
-      return@runningWorker enterState(state.copy(direction = newDirection))
+      return@runningWorker WorkflowAction {
```
rjrjr: Probably better to pull this whole block out into a `handleDirectionChange` method, but I'd still write that method using `WorkflowAction {`.

kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/GameWorkflow.kt
```diff
@@ -156,16 +154,12 @@ class GameWorkflow(
...
-          enterState(
-              state.copy(game = newGame),
-              emittingOutput = PlayerWasEaten
-          )
-        } else {
-          enterState(
-              state.copy(game = newGame),
-              emittingOutput = output
-          )
+        return@runningWorker if (newGame.isPlayerEaten) WorkflowAction {
```
rjrjr: Another spot to consider moving this all out into a single bigger action method.

kotlin/samples/hello-terminal/hello-terminal-app/src/main/java/com/squareup/sample/helloterminal/HelloTerminalWorkflow.kt
```diff
@@ -81,4 +75,14 @@ class HelloTerminalWorkflow : TerminalWorkflow,
   }
 
   override fun snapshotState(state: State): Snapshot = Snapshot.EMPTY
+
+  private fun onKeystroke(key: KeyStroke): HelloTerminalAction = WorkflowAction {
```
rjrjr: This is the kind of place where I especially think `WorkflowAction {` carries its weight — these actions are all tied to worker calls, and so we never make a sink. An `enum` / `sealed class` would simply be boilerplate, and worse introduce an unused mechanism.

rjrjr:
The main driver here is that we want to stop interfering with Swift and Kotlin team-members to be useful to each other.
Without this change it just isn’t practical for Kotlin workflows to define their own own internal action / event types. I think this is the main reason that Kotlin devs are less like to write single-screen workflows, while that’s the norm in iOS. And our strong impression is that single screen is usually a win.

kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowAction.kt
davidapgar: If I'm understanding this correctly, we do have an equivalent in Swift:
https://github.com/square/workflow/blob/master/swift/Workflow/Sources/WorkflowAction.swift#L42

Specifically, it's a constructor to create an `AnyWorkflowAction` (type erased) with a lambda for the `apply`, instead of having to make a full type.

Not something I've seen used in practice (aside from in a few test cases in the infrastructure), but I can see the convenience.

rjrjr: Oh, nice! I totally missed that.

kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/GameWorkflow.kt
rjrjr: Backing away from this one. Its `tick` lambda leans heavily on the child renderings.

kotlin/workflow-testing/src/test/java/com/squareup/workflow/testing/RenderTesterTest.kt
```diff
@@ -138,21 +140,19 @@ class RenderTesterTest {
...
+    val workflow = Workflow.stateful<Unit, String, String, Sink<String>>(
         initialState = { _, _ -> fail() },
         render = { _, state ->
-          onEvent { event ->
-            enterState(
-                newState = "from $state on $event",
-                emittingOutput = "event: $event"
-            )
+          makeEventSink { event ->
+            this@makeEventSink.state = "from $state on $event"
```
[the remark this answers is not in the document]
rjrjr: Yeah, I tried to avoid that in non-test code. Didn't care so much here.

Rationalization: at least it encourages you to break actions out of your `render` method, which usually seems to be a win.

kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/AiWorkflow.kt
```diff
@@ -78,7 +78,10 @@ class AiWorkflow(
         LEFT -> UP
       }
       println("AI changing direction from ${state.direction} to $newDirection")
-      return@onWorkerOutput enterState(state.copy(direction = newDirection))
+      return@onWorkerOutput WorkflowAction {
+        this.state = state.copy(direction = newDirection)
```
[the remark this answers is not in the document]
zach-klippenstein: Oh, because `state` will always refer to closed-over variables before members of receiver. :(

kotlin/samples/tictactoe/common/src/main/java/com/squareup/sample/authworkflow/LoginScreen.kt
```diff
@@ -17,15 +17,6 @@ package com.squareup.sample.authworkflow
...
-  sealed class Event {
-    data class SubmitLogin(
-      val email: String,
-      val password: String
-    ) : Event()
-
-    object Cancel : Event()
-  }
-}
+  val onLogin: (email: String, password: String) -> Unit = { _, _ -> Unit },
```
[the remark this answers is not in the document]
rjrjr: TIL

kotlin/workflow-core/src/main/java/com/squareup/workflow/RenderContext.kt
```diff
@@ -40,51 +41,21 @@ import com.squareup.workflow.WorkflowAction.Companion.noAction
...
     handler: (EventT) -> WorkflowAction<StateT, OutputT>
   ): (EventT) -> Unit
 
+  /**
+   * Creates a sink that will a single [WorkflowAction] of the given type.
+   * Invokes that action by calling [WorkflowAction.apply] to update the current
+   * state, and optionally emit the returned output value if it is non-null.
+   *
+   * Note that only a single action can be processed the sink (or sinks) created
+   * during a `render` call. Redundant calls to [Sink.send] will
```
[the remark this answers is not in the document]
rjrjr:
```
   * Note that only a single action can be processed by the sink (or sinks) created
   * during a `render` call. Redundant calls to [Sink.send] will result in exceptions
   * being thrown.
```

True for now, and we can update when we get around to #541

kotlin/samples/dungeon/app/src/main/java/com/squareup/sample/dungeon/DungeonAppWorkflow.kt
```diff
@@ -42,6 +44,13 @@ class DungeonAppWorkflow(
     data class Running(val board: Board) : State()
   }
 
+  private class StartRunning(val board: Board) : DungeonAction {
```
rjrjr: Not sure why I made this a class rather than a function.

kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/AiWorkflow.kt
```diff
@@ -69,24 +69,28 @@ class AiWorkflow(
...
-    }
+    context.runningWorker(state.directionTicker) { updateDirection() }
 
     return ActorRendering(avatar, Movement(state.direction, cellsPerSecond = cellsPerSecond))
   }
 
   override fun snapshotState(state: State): Snapshot = Snapshot.EMPTY
 }
 
+private fun updateDirection() = WorkflowAction<State, Nothing> {
```
rjrjr: Should be a `val`

kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/GameWorkflow.kt
```diff
@@ -74,7 +70,9 @@ class GameWorkflow(
 
   data class GameRendering(
     val board: Board,
-    val onPlayerEvent: ((PlayerWorkflow.Event) -> Unit)?
+    val onStartMoving: (Direction) -> Unit,
+    val onStopMoving: (Direction) -> Unit,
+    val gameOver: Boolean = false
```
rjrjr: Move `gameOver` above event handlers.

kotlin/workflow-core/src/main/java/com/squareup/workflow/RenderContext.kt
```diff
@@ -40,51 +41,22 @@ import com.squareup.workflow.WorkflowAction.Companion.noAction
...
-   * @param handler A function that returns the [WorkflowAction] to perform when the event handler
-   * is invoked.
-   */
+  @Deprecated("Use makeActionSink.")
   fun <EventT : Any> onEvent(
     handler: (EventT) -> WorkflowAction<StateT, OutputT>
   ): (EventT) -> Unit
 
+  /**
+   * Creates a sink that will a single [WorkflowAction] of the given type.
```
rjrjr: a word

kotlin/workflow-core/src/main/java/com/squareup/workflow/Sink.kt
```diff
@@ -0,0 +1,34 @@
+package com.squareup.workflow
```
rjrjr: copyright

rjrjr: @zach-klippenstein I wonder if we can add a lint check for that.

kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowAction.kt
```diff
@@ -13,33 +13,52 @@
...
-      crossinline update: (StateT) -> Pair<StateT, OutputT?>
+      crossinline name: () -> String = { "" },
+      crossinline apply: Mutator<StateT>.() -> OutputT?
     ): WorkflowAction<StateT, OutputT> = object : WorkflowAction<StateT, OutputT> {
-      override fun invoke(state: StateT): Pair<StateT, OutputT?> = update(state)
+      override fun Mutator<StateT>.apply() = apply.invoke(this)
       override fun toString(): String = "WorkflowAction(${name()})@${hashCode()}"
     }
 
+    inline operator fun <StateT, OutputT : Any> invoke(
```
rjrjr: Not sure why I added this.

kotlin/samples/hello-terminal/hello-terminal-app/src/main/java/com/squareup/sample/helloterminal/HelloTerminalWorkflow.kt
loganj: +1 these are super clean

kotlin/workflow-testing/src/test/java/com/squareup/workflow/testing/RenderTesterTest.kt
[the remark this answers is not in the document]
loganj: worth doing in the test, to set a good example?

rjrjr:
Update fixes signature of `makeActionSink` -- `block` return type should have been `OutputT?`, not `OutputT`. Added related unit test.

PR #540, opened 2019-08-19
wip: Converts Todo sample from onEvent to makeSink.

rjrjr:


kotlin/samples/todo-android/common/src/main/java/com/squareup/sample/todo/TodoEditorWorkflow.kt
```diff
@@ -36,50 +38,84 @@ data class TodoRow(
...
-              copy(done = !done)
-            }))
-            is TextChanged -> emitOutput(ListUpdated(input.updateRow(it.index) {
-              copy(text = it.text)
-            }))
-            is DeleteClicked -> emitOutput(ListUpdated(input.removeRow(it.index)))
-          }
-        }
+        input.copy(rows = input.rows + TodoRow("")),
+        { sink.send(TitleChanged(input, it)) },
```
zach-klippenstein: These should use named params.

kotlin/samples/todo-android/common/src/main/java/com/squareup/sample/todo/TodoEditorWorkflow.kt
```diff
@@ -36,50 +38,84 @@ data class TodoRow(
...
+
+    // Make event handling idempotent.
+    var eventFired = false
+    val sink = object : Sink<TodoAction> {
+      override fun send(value: TodoAction) {
+        if (eventFired) return
+        eventFired = true
+        rawSink.send(value)
+      }
+    }
```
zach-klippenstein: File an issue with a TODO here to implement the same "if the sink type is the same across render passes, let multiple calls to the same sink instance work" behavior that Swift has.

kotlin/samples/todo-android/common/src/main/java/com/squareup/sample/todo/TodoListsWorkflow.kt
```diff
@@ -29,8 +30,8 @@ class TodoListsWorkflow : StatelessWorkflow<List<TodoList>, Int, TodoListsRender
     input: List<TodoList>,
     context: RenderContext<Nothing, Int>
   ): TodoListsRendering {
-    return TodoListsRendering(input, context.onEvent { row ->
-      emitOutput(row)
-    })
+    val sink = context.makeSink<WorkflowAction<Nothing, Int>>()
+
+    return TodoListsRendering(input) { sink.send(emitOutput(it)) }
```
zach-klippenstein: Discussed offline: `makeSink` variant with anonymous action lambda.

kotlin/samples/todo-android/common/src/main/java/com/squareup/sample/todo/TodoEditorWorkflow.kt
rjrjr: https://github.com/square/workflow/issues/541

kotlin/samples/todo-android/common/src/main/java/com/squareup/sample/todo/TodoListsWorkflow.kt
rjrjr:
```
    // A sink that emits the given index as the result of this workflow.
      emitOutput(row)	    val sink: Sink<Int> = context.makeEventSink { index: Int ->
    })	      index
    }
    return TodoListsRendering(lists = input, onRowClicked = sink::send)
```

PR #546, opened 2019-08-20
Rename onWorkerOutput to runningWorker (and friends).

### LINKED ISSUE #545: Kotlin: change onWorkerResult to runningWorker
Swift and Kotlin have inconsistent names for this and they're both kind of lame. We've settled on runningWorker(handler) for Kotlin, and running(worker mumble) for Swift.

Related to #544

zach-klippenstein:
Closes #545.

### DIFF
--- a/kotlin/workflow-core/src/main/java/com/squareup/workflow/RenderContext.kt
+++ b/kotlin/workflow-core/src/main/java/com/squareup/workflow/RenderContext.kt
@@ -32,7 +32,7 @@ import com.squareup.workflow.WorkflowAction.Companion.noAction
  *
  * ## Performing Asynchronous Work
  *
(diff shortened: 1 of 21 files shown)


kotlin/workflow-core/src/main/java/com/squareup/workflow/RenderContext.kt
```diff
@@ -119,11 +119,27 @@ interface RenderContext<StateT, in OutputT : Any> {
    *
    * @param key An optional string key that is used to distinguish between identical [Worker]s.
    */
-  fun <T> onWorkerOutputOrFinished(
+  fun <T> runningWorkerUntilFinished(
```
zach-klippenstein: I'm not sure about this one. Alternatives welcome.

PR #551, opened 2019-08-20
Remove all direct dependencies on Kotlin Reflect. 

vRallev:
Kotlin Reflect is a heavy dependency and the modules barely make use of it to justify the
dependency.

### DIFF
--- a/kotlin/build.gradle
+++ b/kotlin/build.gradle
@@ -71,7 +71,6 @@ buildscript {
               'rx2': "org.jetbrains.kotlinx:kotlinx-coroutines-rx2:${versions.kotlinCoroutines}",
               'test': "org.jetbrains.kotlinx:kotlinx-coroutines-test:${versions.kotlinCoroutines}"
           ],
-          'reflect': "org.jetbrains.kotlin:kotlin-reflect:${versions.kotlin}",
           'test': [
               'common': "org.jetbrains.kotlin:kotlin-test-common",
               'annotations': "org.jetbrains.kotlin:kotlin-test-annotations-common",
@@ -134,8 +133,15 @@ subprojects {
 
   apply plugin: "org.jlleitschuh.gradle.ktlint"
   apply plugin: "io.gitlab.arturbosch.detekt"
-  afterEvaluate {
+  afterEvaluate { project ->
     project.tasks.findByName('check')?.dependsOn 'detekt'
+
+    project.configurations.configureEach {
+      // No module depends on on Kotlin Reflect directly, but there could be transitive dependencies
+      // in tests with a lower version. This could cause problems with a newer Kotlin version that
+      // we use.
+      resolutionStrategy.force "org.jetbrains.kotlin:kotlin-reflect:${versions.kotlin}"
+    }
   }
 
   tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
--- a/kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/WorkflowId.kt
+++ b/kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/WorkflowId.kt
@@ -22,7 +22,6 @@ import com.squareup.workflow.writeUtf8WithLength
 import okio.Buffer
 import okio.ByteString
 import kotlin.reflect.KClass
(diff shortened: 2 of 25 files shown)


kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/WorkflowId.kt
```diff
@@ -46,9 +45,14 @@ internal constructor(
...
 
+private val KClass<*>.jClass: Class<*>
+  get() = this::class.java.getMethod("getJClass").invoke(this) as Class<*>
+
 internal fun WorkflowId<*, *, *>.toByteString(): ByteString = Buffer()
     .also { sink ->
-      sink.writeUtf8WithLength(type.jvmName)
+      // Unfortunately, we can't use type.java.name. There might be issue with inner class names
+      // and inlined types. Instead, we run the same code as Kotlin reflect would do.
+      sink.writeUtf8WithLength(type.jClass.name)
```
vRallev: That's my only concern.

vRallev:
@zach-klippenstein Do you know why Travis is failing? 
```
\["\[One of the secure variables in your .travis.yml has an invalid format.\]", "\[One of the secure variables in your .travis.yml has an invalid format.\]"\]
```

rjrjr:
@vRallev That message is a red herring, it always says that. Click through to see the real logs.

kotlin/build.gradle
```diff
@@ -134,8 +134,15 @@ subprojects {
...
   apply plugin: "org.jlleitschuh.gradle.ktlint"
   apply plugin: "io.gitlab.arturbosch.detekt"
-  afterEvaluate {
+  afterEvaluate { project ->
     project.tasks.findByName('check')?.dependsOn 'detekt'
+
+    project.configurations.configureEach {
+      // No module depends on on Kotlin Reflect directly, but there could be transitive dependencies
+      // in tests with a lower version. This could cause problems with a newer Kotlin version that
+      // we use. To
```
zach-klippenstein: To…

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/WorkflowId.kt
zach-klippenstein: What is the potential issue?

kotlin/build.gradle
```diff
@@ -134,8 +134,15 @@ subprojects {
...
+      resolutionStrategy.force deps.kotlin.reflect
```
zach-klippenstein: To prevent adding the dependency back in the future by accident, can you hard-code the artifact name here and delete it from the map?

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/WorkflowId.kt
vRallev: There was an issue in a unit test where type was `com.squareup.workflow.internal.WorkflowNodeTest$snapshots non-empty with children$$inlined$stateful$1`. But this happened because I used `type::java.name` instead of `type.java.name`. It's fixed now.

kotlin/build.gradle
vRallev: Ohh

kotlin/build.gradle
vRallev: I thought about this, but personally would have preferred the current way. But I'm changing it.

zach-klippenstein: If we actually need the dependency, we can add it back to the list.

vRallev:
Rebased on master because of a conflict.

kotlin/build.gradle
zach-klippenstein: Did you happen to note which modules include it as a transitive dep? Is dropping it as a direct dependency _actually_ saving us anything?

vRallev: Kotlin Mockito depends on it. Yes, it's saving us a lot, because you need it only for tests. I could have added a dependency on the latest version with a `testImplementation`, but forcing the concrete version is the cleaner solution IMO.

zach-klippenstein: Ok good.

PR #556, opened 2019-08-22
Never pass an empty snapshot to initialState.

### LINKED ISSUE #506: Do not restore empty snapshots
When a restored snapshot is empty, we should pass `null` to the `snapshot` param of `initialState`. Otherwise we're forcing every implementation to make that check, which is silly.

zach-klippenstein:
Two commits, I'd recommend reviewing individually.

Fixes #506

### DIFF
--- a/kotlin/workflow-core/src/main/java/com/squareup/workflow/StatefulWorkflow.kt
+++ b/kotlin/workflow-core/src/main/java/com/squareup/workflow/StatefulWorkflow.kt
@@ -75,7 +75,9 @@ abstract class StatefulWorkflow<
    * initial state.
    *
    * @param snapshot
-   * If the workflow is being created fresh, [snapshot] will be null.
+   * If the workflow is being created fresh, OR the workflow is being restored from an empty
+   * [Snapshot], [snapshot] will be null. A snapshot is considered "empty" if [Snapshot.bytes]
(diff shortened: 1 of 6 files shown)


kotlin/workflow-core/src/main/java/com/squareup/workflow/StatefulWorkflow.kt
```diff
@@ -75,7 +75,8 @@ abstract class StatefulWorkflow<
    * initial state.
    *
    * @param snapshot
-   * If the workflow is being created fresh, [snapshot] will be null.
+   * If the workflow is being created fresh, OR the workflow is being restored from an empty
+   * [Snapshot] (i.e. [snapshotState] returned [Snapshot.EMPTY]), [snapshot] will be null.
```
rjrjr: i.e. [Snapshot.bytes] returned an empty [ByteString], probably because [snapshotState] returned [Snapshot.EMPTY]

zach-klippenstein: Pro tip: use the "Insert a suggestion" button and then I can apply your suggestion with a button click 😉 
image

zach-klippenstein: Ugh, way too hard to figure out where the line break boundary is, that's a useless feature here.

PR #564, opened 2019-08-24
Tidies Hello samples.

rjrjr:
Vestigial  event snipped, less ViewRegistry instantiation.


rjrjr:
Update adds missing `UsesExperimental` annotation.

PR #574, opened 2019-08-28
Update ReactiveSwift to 6.0.0

lechristian:
Updating ReactiveSwift to 6.0.0
- Remove the dependency on `antitypical/Result` in favor of Swift 5's `Swift.Result`
- Replace `NoError` with `Swift.Never` since `Never` now conforms to `Swift.Error` [Swift Evolution Proposal](https://github.com/apple/swift-evolution/blob/master/proposals/0215-conform-never-to-hashable-and-equatable.md)

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -11,7 +11,7 @@ let package = Package(
             targets: ["Workflow"]),
     ],
     dependencies: [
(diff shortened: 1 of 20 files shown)


rjrjr:
Thanks again for the contribution!

PR #587, opened 2019-09-02
Updates README.md to be less redundant and less qualified.

rjrjr:
We're really not experimental any more.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,19 +4,20 @@
 [![CocoaPods compatible](https://img.shields.io/cocoapods/v/Workflow.svg)](https://cocoapods.org/pods/Workflow)
 [![Maven Central](https://img.shields.io/maven-central/v/com.squareup.workflow/workflow-core-jvm.svg?label=Maven%20Central)](https://search.maven.org/search?q=g:%22com.squareup.workflow%22)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,19 +4,13 @@
...
 
-An architecture that allows composable state machines to drive UI navigation and content, where the
- state machines are cleanly separated from UI code.
+A unidirectional data flow library for Kotlin and Swift, emphasizing modularity and the ability to scale. 
 
 _**This project is currently experimental and the API subject to breaking changes without notice.**
 Follow Square's engineering blog, [The Corner](https://developer.squareup.com/blog/), to see when
 this project becomes stable._
 
-Workflow is a library for making composable state machines, and UIs driven by those state machines.
```
zach-klippenstein: This paragraph is still relevant, isn't it?

rjrjr: It doesn't feel like our primary focus any more, just one use case of many. But let me try to rephrase and bring it back a bit.

Also forgot to link to the slide deck, which was why I opened this in the first place. :/

davidapgar: +1 on having something about composable state machines. It's a major focus from my standpoint of what makes this valuable

Issue #512, opened 2019-07-31
Replace receiveOrNull with receiveOrClosed once released.


zach-klippenstein:
Starting in 1.3.0-RC, `receiveOrClosed` exists – however it's marked as `@InternalCoroutinesApi`. There are plans to make it public, but it's blocked by inline class stability (https://github.com/Kotlin/kotlinx.coroutines/pull/1343#discussion_r304082103).

zach-klippenstein:
Tried using the preview version of `receiveOrClosed`, but turns out it's broken (https://github.com/Kotlin/kotlinx.coroutines/issues/1584) and caused a set of bugs with odd stacktraces (#626).

zach-klippenstein:
The referenced coroutines bug was fixed some time ago. Might want to give this another try.

PR #589, opened 2019-09-03
Remove runningWorkerUntilFinished.

### LINKED ISSUE #434: Eliminate map call to enable channel operator fusion for Kotlin Workers.
Flow implements operator fusion for operators that rely on channels. E.g., if you start with an RxJava2 `Flowable`, and then do:
```kotlin
val subscriptionChannel = myFlowable
  .flowOn(Dispatchers.IO) // Subscribe/observe on a background thread.
  .buffer(50) // Buffer up to 50 items when there's backpressure.
  .produceIn(scope) // Subscribe to the Flowable with a ReceiveChannel.
```
Then only a single channel will be created and shared between all the operators to own the subscription to the `Flowable`, communicate between threads, and store the buffer. And that same channel will be what is returned from `produceIn`. This is just an optimization, since each of those declarative operators describe a different aspect of channels' behavior. If you put another, non-channel operator between any of those operators, two separate channels will be created. The behavior of the `Flow` shouldn't change, but it will use a bit more memory.

The Workflow runtime uses channels to subscribe to Workers. In practice, at least for now, many workers are likely to be direct wrappers around other reactive stream types.  So while we can't force Workers to use operator fusion (nor would we want to), we should at least not _prevent_ fusion from occurring when possible.

Currently, to keep the change to using Flow for the Worker API simple, call `map` just before `produceIn` in order to wrap values in the `Output` sentinel type. This prevents any operator fusion from ever occurring in the common case of, for example `myObservable.asWorker()`. We should remove that map to allow fusion to occur.

### LINKED ISSUE #558: Remove runningWorkerUntilFinished
The need for an explicit "finished" signal came from before we were using a real stream API for Workers. Now that we're using Flow, if you care about when some worker finishes, you can just use stream operators (with [`transform`](https://square.github.io/workflow/kotlin/api/workflow-core/com.squareup.workflow/transform/)) to emit your own trailer sentinel.

zach-klippenstein:
Closes #558, removes 1/2 of #512 TODOs, and closes #434.

### DIFF
--- a/kotlin/workflow-core/src/main/java/com/squareup/workflow/RenderContext.kt
+++ b/kotlin/workflow-core/src/main/java/com/squareup/workflow/RenderContext.kt
@@ -17,9 +17,6 @@
 
 package com.squareup.workflow
 
-import com.squareup.workflow.Worker.OutputOrFinished
-import com.squareup.workflow.Worker.OutputOrFinished.Finished
-import com.squareup.workflow.Worker.OutputOrFinished.Output
 import com.squareup.workflow.WorkflowAction.Companion.noAction
 import com.squareup.workflow.WorkflowAction.Mutator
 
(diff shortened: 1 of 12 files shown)


kotlin/workflow-core/src/main/java/com/squareup/workflow/RenderContext.kt
```diff
@@ -86,32 +83,17 @@ interface RenderContext<StateT, in OutputT : Any> {
   ): ChildRenderingT
 
   /**
-   * Ensures [worker] is running. When the [Worker] emits an output or finishes, [handler] is called
-   * to determine the [WorkflowAction] to take.
+   * Ensures [worker] is running. When the [Worker] emits an output, [handler] is called
+   * to determine the [WorkflowAction] to take. When the worker finishes, nothing happens (although
+   * another render pass may be triggered).
```
rjrjr: Maybe document how to use `transform` to react to finish if needed?

zach-klippenstein: Almost nobody cares, and I don't see the need to duplicate the Flow documentation.

PR #599, opened 2019-09-05
Add swift tutorial on unit/integration testing workflows

davidapgar:
Relevant things to review are:
`*Tests.swift`
Tutorial5.md

The other additions are effectively a copy of the resulting code from Tutorial 4 (which would be the starting point for this tutorial).

### DIFF
--- a/swift/Samples/Tutorial/Tutorial5.md
+++ b/swift/Samples/Tutorial/Tutorial5.md
@@ -1,0 +1,843 @@
+# Step 5
+
+_Unit and Integration Testing Workflows_
+
+## Setup
+
+To follow this tutorial:
+- Open your terminal and run `bundle exec pod install` in the `swift/Samples/Tutorial` directory.
+- Open `Tutorial.xcworkspace` and build the `Tutorial` Scheme.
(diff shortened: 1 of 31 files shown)


swift/Samples/Tutorial/Tutorial5.md
```diff
@@ -0,0 +1,838 @@
+# Step 5
+
+_Unit and Integration Testing Workflows_
+
+## Setup
+
+Start from the completed code after Tutorial 4, or if you're jumping ahead, the completed `Tutorial4` pod. The default scheme of `Tutorial` will run unit tests from `TutorialBase` and `Tutorial4` when pressing `cmd+shift+u`
```
timdonnelly: I would leave explicit instructions in here: "to follow this tutorial, open your terminal and do X to get Xcode into a buildable state".

It's common to jump straight to a tutorial when it's relevant, so it would be nice to make this as easy as possible.

davidapgar: Updated in this tutorial as well as all the previous ones. Good call.

PR #612, opened 2019-09-14
Introduce debug snapshotting like Swift has

### LINKED ISSUE #343: Copy Swift's debug snapshots mechanism for tracing workflows
See the `DebugSnapshot` code in Swift. Do that.

See also #129 and #130.

zach-klippenstein:
3 commits, recommend reviewing individually. First one cleans up the `launchWorkflowIn` API to prepare to add another stream, then state snapshots are introduced, and finally update info. 

This PR introduces a new stream to the function passed to `launchWorkflowIn`, one that emits a tuple of two things on every render pass:
- `WorkflowHierarchyDebugInfo` – A simple tree data structure that mirrors the workflow tree and, at each node, describes the type of the workflow at that node and the state of that workflow at the time of that render pass.
- `WorkflowUpdateDebugInfo` – A description of a path down the tree that indicates which workflow triggered the render pass and why. Includes information about which nodes up the tree executed `WorkflowAction`s as part of the update. Null for the first render pass, since it is not triggered by any update.

See the kdoc on those types for more information.

Closes #343.


kotlin/workflow-runtime/src/main/java/com/squareup/workflow/debugging/WorkflowHierarchyDebugSnapshot.kt
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * @param workflowType A string representation of the type of this workflow.
+ * @param stateDescription A (lazy) string representation of the state of this workflow after the
+ * last render pass.
+ * @param children All the child workflows that were rendered by this workflow in the last render
+ * pass. See [Child].
+ */
+data class WorkflowHierarchyDebugSnapshot(
+  val workflowType: String,
+  val stateDescription: LazyString,
```
zach-klippenstein: I was too squeamish about the performance impact of converting every state value in the tree to a string on every render pass to put the string in this class directly. The idea is that if a particular collector of the debug stream needs to get the string representation en mass, it can do it on a background thread.

zach-klippenstein: Why is this a string representation of the state and not just the state itself (as the convenience constructor below might even suggest it should be)? I'm not entirely sure, but I've done it this way for now since that's what Swift does:
https://github.com/square/workflow/blob/8acff486c2fc80b1de2e1956170819f03fa0eafd/swift/Workflow/Sources/Debugging.swift#L129-L133

Would appreciate if @timdonnelly could provide the rationale for eagerly converting to a string representation.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/debugging/WorkflowUpdateDebugInfo.kt
```diff
@@ -0,0 +1,106 @@
...
+  /**
+   * A sealed class that indicates whether a workflow actually executed a `WorkflowAction`, or
+   * was just the ancestor of a workflow that did.
+   *
+   * Contains two subclasses, see their documentation for details:
+   * - [DidUpdate]
+   * - [ChildDidUpdate]
+   */
+  sealed class Kind {
+    // TODO more kotliny names
```
zach-klippenstein: "DidUpdate" is a very idiomatic name in Swift/ObjC, not so much in Kotlin. I'm thinking just `Updated` and `ChildUpdated`?

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/debugging/WorkflowUpdateDebugInfo.kt
```diff
@@ -0,0 +1,106 @@
...
+     * [Sink][com.squareup.workflow.Sink] (see
+     * [makeActionSink][com.squareup.workflow.RenderContext.makeActionSink]).
+     */
+    object External : Source()
+
+    /**
+     * Indicates that the update was triggered by an output emitted from a
+     * [Worker][com.squareup.workflow.Worker] being run by this workflow.
+     */
+    data class Worker(val key: String) : Source()
```
zach-klippenstein: Should this class (and `Subtree` below) also include the actual result of the update (i.e. the output value that was emitted)? Seems like that could be helpful for debugging, but again I'm not sure why Swift was reluctant to include raw values in these types.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/RealRenderContext.kt
```diff
@@ -93,7 +95,9 @@ class RealRenderContext<StateT, OutputT : Any>(
     val case: WorkflowOutputCase<ChildPropsT, ChildOutputT, StateT, OutputT> =
       WorkflowOutputCase(child, id, props, handler)
     childCases += case
-    return renderer.render(case, child, id, props)
+    val (rendering, debugSnapshot) = renderer.render(case, child, id, props)
+    childDebugSnapshots += Child(key, debugSnapshot)
```
zach-klippenstein: TODO Needs kdoc.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/SubtreeManager.kt
```diff
@@ -81,13 +84,24 @@ internal class SubtreeManager<StateT, OutputT : Any>(
...
+    selector: SelectBuilder<OutputEnvelope<T>>,
+    handler: (WorkflowAction<StateT, OutputT>, Kind) -> OutputEnvelope<T>
   ) {
     for ((case, host) in hostLifetimeTracker.lifetimes) {
       host.tick(selector) { output ->
-        val componentUpdate = case.acceptChildOutput(output)
-        return@tick handler(componentUpdate)
+        if (output.output != null) {
+          val componentUpdate = case.acceptChildOutput(output.output)
+          // This workflow's child emitted a non-null output, this workflow's action was invoked,
```
zach-klippenstein: Sentence reads awkwardly, needs more words.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/SubtreeManager.kt
```diff
@@ -81,13 +84,24 @@ internal class SubtreeManager<StateT, OutputT : Any>(
...
+          // meaning we did update ourselves.
+          return@tick handler(
+              componentUpdate,
+              Kind.DidUpdate(Source.Subtree(case.id.name, output.debugInfo))
+          )
+        } else {
+          // The child didn't actually emit an output, which means this workflow has nothing to do,
+          // which means we will only report our child updating.
+          return@tick handler(noAction(), Kind.ChildDidUpdate(output.debugInfo))
```
zach-klippenstein: There should probably be some kdoc somewhere saying that if the kind is `ChildDidUpdate`, `noAction` is the only valid workflow action.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/WorkflowId.kt
```diff
@@ -39,6 +39,8 @@ internal constructor(
     workflow: Workflow<PropsT, OutputT, RenderingT>,
     name: String = ""
   ) : this(workflow::class, name)
+
+  val typeDebugString: String = type.java.name
```
zach-klippenstein: Needs kdoc

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/WorkflowNode.kt
```diff
@@ -150,11 +160,14 @@ internal class WorkflowNode<PropsT, StateT, OutputT : Any, RenderingT>(
     // Listen for any events.
     with(selector) {
       behavior!!.nextActionFromEvent.onAwait { update ->
-        acceptUpdate(update)
+        acceptUpdate(update, Kind.DidUpdate(Source.External))
       }
     }
   }
 
+  private fun createDebugInfo(kind: Kind): WorkflowUpdateDebugInfo =
```
zach-klippenstein: kdoc

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/WorkflowSession.kt
```diff
@@ -0,0 +1,32 @@
...
+ * A tuple of [Flow]s representing all the emissions from the workflow runtime.
+ *
+ * Passed to the function taken by [launchWorkflowIn].
+ *
+ * @param debugSnapshots A stream of [diagnostic information][WorkflowDebugInfo] about the runtime.
+ */
+class WorkflowSession<out OutputT : Any, out RenderingT>(
+  val renderingsAndSnapshots: Flow<RenderingAndSnapshot<RenderingT>>,
+  val outputs: Flow<OutputT>,
+  val debugSnapshots: Flow<WorkflowDebugInfo>
```
zach-klippenstein: `debugInfos`? `debugger`?

zach-klippenstein: Going with `debugInfo`.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/SubtreeManager.kt
zach-klippenstein: Added a bunch of kdoc to this method.

zach-klippenstein:
Pushed to address some of my own comments.

kotlin/samples/tictactoe/app/src/main/java/com/squareup/sample/mainactivity/MainActivity.kt
```diff
@@ -48,7 +49,10 @@ class MainActivity : AppCompatActivity() {
       finish()
     }
 
-    loggingSub = workflowRunner.renderings.subscribe { Timber.d("rendering: %s", it) }
+    loggingSub = CompositeDisposable(
+        workflowRunner.renderings.subscribe { Timber.d("rendering: %s", it) },
+        workflowRunner.debugSnapshots.subscribe { Timber.d("debug snapshot: %s", it) }
```
rjrjr: Nit: `Timber.v` for the debug snapshot

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/debugging/LazyString.kt
```diff
@@ -0,0 +1,30 @@
...
+ */
+package com.squareup.workflow.debugging
+
+import kotlin.LazyThreadSafetyMode.PUBLICATION
+
+/**
+ * Wrapper around a string provided lazily by a function that implements equals/hashcode/toString
+ * using the lazy string.
+ */
+class LazyString(provider: () -> String) {
```
rjrjr: Does this really have to be public?

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/debugging/WorkflowHierarchyDebugSnapshot.kt
rjrjr: If Tim doesn't reply by the time you merge I'd argue to just hold the state. Hard pressed to imagine and argument against it.

Would you still want to cache the rendered strings?

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/debugging/WorkflowDebugInfo.kt
```diff
@@ -0,0 +1,25 @@
...
+package com.squareup.workflow.debugging
+
+/**
+ * A pair of [WorkflowHierarchyDebugSnapshot] and optional [WorkflowUpdateDebugInfo].
+ *
+ * The [WorkflowUpdateDebugInfo] will be null on the first render pass and non-null on every
+ * subsequent pass, since the first pass is the only one that is not triggered by some kind of
+ * update.
+ */
+typealias WorkflowDebugInfo = Pair<WorkflowHierarchyDebugSnapshot, WorkflowUpdateDebugInfo?>
```
rjrjr: Why not a `data class`?

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/debugging/WorkflowUpdateDebugInfo.kt
rjrjr: SG

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/debugging/WorkflowUpdateDebugInfo.kt
rjrjr: Is Swift always doing this eager string rendering in production?

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/SubtreeManager.kt
```diff
@@ -79,15 +82,36 @@ internal class SubtreeManager<StateT, OutputT : Any>(
...
     for ((case, host) in hostLifetimeTracker.lifetimes) {
       host.tick(selector) { output ->
-        val componentUpdate = case.acceptChildOutput(output)
-        return@tick handler(componentUpdate)
+        if (output.output != null) {
+          val componentUpdate = case.acceptChildOutput(output.output)
+          // This workflow's child emitted a non-null output: the WorkflowNode will execute the
+          // WorkflowAction, and we need to pass DidUpdate to indicate that this workflow actually
+          // got updated.
+          return@tick handler(
```
rjrjr: Can you hoist the `return@tick` out to the `if`?

kotlin/samples/tictactoe/app/src/main/java/com/squareup/sample/mainactivity/MainActivity.kt
zach-klippenstein: Fine, but why?

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/debugging/LazyString.kt
zach-klippenstein: Nope. Will fix…

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/debugging/WorkflowDebugInfo.kt
zach-klippenstein: Meant to change that.

kotlin/samples/tictactoe/app/src/main/java/com/squareup/sample/mainactivity/MainActivity.kt
rjrjr: Because I actually use those rendering lines when debugging, and they'll be harder to read if I can't easily filter out the debug snapshots.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/SubtreeManager.kt
zach-klippenstein: Probably

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/debugging/WorkflowHierarchyDebugSnapshot.kt
zach-klippenstein: Yea I might just do that. Is this too over-engineered?

```kotlin
data class WorkflowHierarchyDebugSnapshot(
  val state: Any?,
  …
) {
  val stateDescription: String by lazy(PUBLICATION, state::toString)
}
```

rjrjr: LGTM

rjrjr: What does `PUBLICATION` do?

zach-klippenstein: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-lazy-thread-safety-mode/index.html#enum-values

> Initializer function can be called several times on concurrent access to uninitialized Lazy instance value, but only the first returned value will be used as the value of Lazy instance.

Doesn't require a full lock on the field to read, uses optimistic/CAS locking I believe. Cheaper to read the property if the initializer function doesn't have side effects.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/debugging/WorkflowUpdateDebugInfo.kt
zach-klippenstein: I believe so.

kotlin/samples/tictactoe/app/src/main/java/com/squareup/sample/mainactivity/MainActivity.kt
zach-klippenstein: You can: just filter on `/rendering: /`.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/debugging/WorkflowHierarchyDebugSnapshot.kt
zach-klippenstein: Now I can delete `LazyString`, yay.

zach-klippenstein: I think I will also add `rendering` and `renderingDescription` to this class, and `output` and `outputDescription` to `WorkflowUpdateDebugInfo` then.

Only rationale for providing `*Description` properties is caching, which I'm not really convinced is important enough to bother with, and giving us some indirection in case we want to allow customizing how that description is generated in the future – but I'm also not convinced overriding `toString` isn't good enough for that.

kotlin/samples/tictactoe/app/src/main/java/com/squareup/sample/mainactivity/MainActivity.kt
rjrjr: Then I'll only see the renderings. Having the full debug output seems verbose to me, so `.v` is my knee-jerk.

zach-klippenstein: Oh, I thought `d` was lower priority than `v` – it's the other way around. Makes sense now.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/debugging/WorkflowHierarchyDebugSnapshot.kt
zach-klippenstein: Just talked to @bencochran offline and everything makes sense now. He's slightly worried about the eager stringification we're doing in Swift, and there doesn't seem to be any reason we couldn't just store the state directly in the snapshot and do the encoding lazily in Swift too. The current implementation was likely used simply because it prevented having to manually implement `Codable` in `WorkflowHierarchyDebugSnapshot`.

zach-klippenstein:
All feedback addressed, merging on green.

PR #614, opened 2019-09-17
Hello View.backButtonPressed

### LINKED ISSUE #250: Drop HandlesBack and use OnBackPressedCallback instead
Requires AndroidX, so blocked by #50.

### LINKED ISSUE #466: Fix modal back button handling

rjrjr:
Much simpler back button scheme. Closes #466, closes #250.


kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ModalViewContainer.kt
```diff
@@ -46,14 +47,28 @@ internal class ModalViewContainer
...
+              true
+            } else {
+              false
+            }
+          }
 
           setCancelable(false)
           setContentView(modalDecorator(view))
+              .apply {
+                backPressedHandler = {
```
rjrjr: Should not clobber an existing handler if there was one.

Perhaps shouldn't do this here at all, rely on PanelContainer and its ilk to do the right thing? That'll be so easy to screw up, though. :/

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewBackHandler.kt
```diff
@@ -0,0 +1,65 @@
+package com.squareup.workflow.ui
+
+import android.content.Context
+import android.content.ContextWrapper
+import android.view.View
+import android.view.View.OnAttachStateChangeListener
+import androidx.activity.ComponentActivity
+import androidx.activity.OnBackPressedCallback
+
+typealias BackPressHandler = () -> Unit
```
rjrjr: kdoc

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewBackHandler.kt
```diff
@@ -0,0 +1,65 @@
...
+
+var View.backPressedHandler: BackPressHandler?
```
rjrjr: kdoc

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewShowRendering.kt
```diff
@@ -60,8 +67,8 @@ fun View.canShowRendering(rendering: Any): Boolean {
 fun <RenderingT : Any> View.showRendering(rendering: RenderingT) {
   showRenderingTag
       ?.let { tag ->
-        check(tag.showing.matches(rendering)) {
-          "Expected $this to be able to update of ${tag.showing} from $rendering"
+        check(tag.initialRendering.matches(rendering)) {
```
rjrjr: I want to follow up and start updating this value. In the meantime, `initialRendering` is a more accurate description of how things actually work at the moment.

kotlin/workflow-ui-android/src/main/res/values/ids.xml
```diff
@@ -17,8 +17,10 @@
...
   <!-- View Tag that holds the function to update a view created via ViewRegistry. -->
   <item type="id" name="view_show_rendering_function"/>
-  <!-- View Tag used for back button handlers. See HandlesBack.Helper -->
-  <item type="id" name="workflow_back_handler"/>
+  <!-- View Tag used for back button handlers. -->
+  <item type="id" name="view_back_handler"/>
+  <!-- View Tag storing the OnBackPressedCallback that drives view_back_handler. -->
+  <item type="id" name="view_on_back_pressed_callback"/>
   <!-- View Tag holding the BackStackScreen.Key associated with a view. -->
   <item type="id" name="workflow_back_stack_key"/>
```
rjrjr: Unused, delete.

kotlin/workflow-ui-core/src/main/java/com/squareup/workflow/ui/BackStackScreen.kt
```diff
@@ -16,7 +16,8 @@
 package com.squareup.workflow.ui
 
 /**
- * @param stack: screens that have are / have been displayed, ending in the current screen
+ * @param stack: screens that have are / have been displayed, ending in the current screen.
```
rjrjr: This is gibberish.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewBackHandler.kt
```diff
@@ -0,0 +1,65 @@
...
+        }
+    }
+  }
+
+private var View.onBackPressedCallback: AttachedOnBackPressedCallback?
+  get() {
+    return getTag(R.id.view_on_back_pressed_callback) as? AttachedOnBackPressedCallback
+  }
+  set(value) {
+    onBackPressedCallback?.remove()
```
zach-klippenstein: Needs a comment explaining why this is necessary - I think I know, but so many things are being added and removed in this file that I feel like I'm reading a riddle.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewBackHandler.kt
```diff
@@ -0,0 +1,65 @@
...
+import android.view.View.OnAttachStateChangeListener
+import androidx.activity.ComponentActivity
+import androidx.activity.OnBackPressedCallback
+
+typealias BackPressHandler = () -> Unit
+
+var View.backPressedHandler: BackPressHandler?
+  get() {
+    val handler = getTag(R.id.view_back_handler)
+    check(handler is Function0<*>?) {
```
zach-klippenstein: Why is this check necessary here? Isn't it enough to just return null like all the other getters? Even this getter uses Elvis right below.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewBackHandler.kt
```diff
@@ -0,0 +1,65 @@
...
+  }
+  set(value) {
+    onBackPressedCallback?.remove()
+    setTag(R.id.view_on_back_pressed_callback, value)
+  }
+
+private class AttachedOnBackPressedCallback(val view: View) : OnBackPressedCallback(true),
+    OnAttachStateChangeListener {
+  override fun handleOnBackPressed() {
+    view.backPressedHandler!!.invoke()
```
zach-klippenstein: Can't the handler function just also be passed to this class directly, so we don't need this null assertion?

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewBackHandler.kt
```diff
@@ -0,0 +1,65 @@
...
+private var View.onBackPressedCallback: AttachedOnBackPressedCallback?
+  get() {
+    return getTag(R.id.view_on_back_pressed_callback) as? AttachedOnBackPressedCallback
+  }
+  set(value) {
+    onBackPressedCallback?.remove()
+    setTag(R.id.view_on_back_pressed_callback, value)
+  }
+
+private class AttachedOnBackPressedCallback(val view: View) : OnBackPressedCallback(true),
```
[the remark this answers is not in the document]
zach-klippenstein: Actually this whole class just needs kdoc too

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewBackHandler.kt
zach-klippenstein: Ah, it doesn't because if the callback is already set we only bother to update the tag. Instead of that tricky layer of indirection, can we just store this class in the tag and make the handler function a mutable property on this class?

zach-klippenstein:
> Isn't there a new jetpack library that is supposed to let you run UI tests as unit tests?

Indeed there is, and I just filed #615.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewBackHandler.kt
rjrjr: It's an invariant check, e.g. to make sure some other code hasn't stomped our field. Anyway, point could be moot if I can move it out of the view tag like you suggested elsewhere.

rjrjr:
We have automated tests for this internally, including existing tests that will fail if this is broken. I can add more to test the modal fix, and I'll make sure to run them before merging. 

We can / should add the unit tests you suggest in #615, but those will never be adequate. The most interesting problems here are in the integration, especially the modal bits. I no longer believe that Espresso tests can be made stable under Travis. I suppose it's worth seeing if Robolectric can test the integration. :(

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ModalViewContainer.kt
rjrjr: I was over thinking when I wrote this comment -- this is just another uncaught exception handler type of situation. If there is one on the view already, we just have to make sure that this backstop calls it.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewBackHandler.kt
zach-klippenstein: None of the other getters do invariant checks though, they just do the safe cast, which is why I'm wondering why this one is special.

rjrjr: I was stupidly overreacting to the loss of the un-reified return type. I'm going to follow up and move this all into a single view tag like you suggested, that should moot the issue anyway.

rjrjr:
Oh my. I had no idea how ambitious the new test classes are. I sure hope they work, shifting to #615 before doing more on this.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ModalViewContainer.kt
rjrjr: It's even simpler than that: if this view, the root of the modal container, has an back press handler, do nothing. It will fire and prevent any upstream ones, which should be blocked by the modal, from doing anything. The no-op is only needed if the handler is null.

rjrjr:
Much cleaner now, and documented. Still need to add more tests before merging, though.

rjrjr:
Update adds test coverage and fixes kdoc in BackStackScreen.kt

PR #618, opened 2019-09-18
Introduces TicTacToeEspressoTest.

rjrjr:
Also:

 * Fixes some bitrot in the app, introduced during WorkflowAction converion.
 * Converts AuthService to Kotlin.


kotlin/build.gradle
```diff
@@ -15,8 +15,10 @@
  */
 buildscript {
   ext.versions = [
-      'targetSdk': 29,
+      // Can't move to 29 until Robolectric does.
+      'targetSdk': 28,
```
rjrjr: undo

kotlin/samples/tictactoe/app/src/main/java/com/squareup/sample/mainactivity/MainComponent.kt
```diff
@@ -32,7 +39,24 @@ import timber.log.Timber
  */
 internal class MainComponent {
 
-  private val authService = AuthService()
+  private val countingIdlingResource  = CountingIdlingResource("fred")
```
rjrjr: "fred"

zach-klippenstein:
Looks really good so far!

PR #620, opened 2019-09-20
Update the Worker "Network request" example

SalvatoreT:
It still had the `#performWork` example which needed to change to `#run`. I'm also using this as a chance to use the much easier coroutine syntax.


zach-klippenstein:
Thanks!

PR #625, opened 2019-09-23
Add running workers to WorkflowHierarchyDebugSnapshots.

zach-klippenstein:


kotlin/workflow-runtime/src/main/java/com/squareup/workflow/debugging/WorkflowHierarchyDebugSnapshot.kt
```diff
@@ -53,11 +56,33 @@ data class WorkflowHierarchyDebugSnapshot(
    * @param key The string key that was used to render this child.
    * @param snapshot The [WorkflowHierarchyDebugSnapshot] that describes the state of this child.
    */
-  data class Child(
+  data class ChildWorkflow(
```
rjrjr: Thanks for this rename.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/debugging/WorkflowHierarchyDebugSnapshot.kt
```diff
@@ -53,11 +56,33 @@ data class WorkflowHierarchyDebugSnapshot(
...
+   * Represents a worker that is being run by the workflow represented by a particular
+   * [WorkflowHierarchyDebugSnapshot].
+   *
+   * @param key The string key that was used to render this child.
+   * @param describe A function that returns a string that describes the worker. Includes the output
+   * of the worker's `toString` method.
+   */
+  class ChildWorker(
+    val key: String,
+    val describe: () -> String
```
zach-klippenstein: This should be private.

zach-klippenstein:
Just rewrote this infra, this PR is obsolete.

PR #633, opened 2019-10-01
Add IDEA dictionary file to git.

zach-klippenstein:


zach-klippenstein:
Rx test keeps flaking (#591), this can't possibly break tests so just merging.

PR #638, opened 2019-10-04
Remove racy hack in WorkerTester.

### LINKED ISSUE #591: One of the RxWorker tests might be flaky
Failed in CI, doesn't fail locally.

https://travis-ci.com/square/workflow/jobs/231198759

zach-klippenstein:
I think this was causing some test flakes, so this should fix #591.

Also removed the context parameter to `Worker.test`, since it was only there
to use with `TestCoroutineContext` but that apparently doesn't support `yield`
so it's no longer necessary.


zach-klippenstein:
@rjrjr feel free to merge if this LG

PR #647, opened 2019-10-07
Add a section to RELEASING.md about publishing snapshot artifacts.

zach-klippenstein:


zach-klippenstein:
Oops, we already had a section for this at the bottom of the file (thanks docs lint!!). I moved the enhanced content down there.

PR #650, opened 2019-10-07
Add an RxWorker class so Workers can be implemented without using experimental Flow APIs.

zach-klippenstein:
As I was upgrading us to 0.21.0 internally, realized that some code needing custom workers was calling `asFlow()`, which requires an experimental annotation. This is just to avoid that.


zach-klippenstein:
I'm not sure this is the best approach, and if we do land it then as soon as `asFlow` becomes stable I want to delete it.

PR #653, opened 2019-10-08
Introduce an anonymous, lambda-based LifecycleWorker builder.

zach-klippenstein:
I saw someone making a worker out of a `Completable` just to dispose a `CompositeDisposable` – this feels more kotlin-y than an explicit anonymous `LifecycleWorker` object, but maybe it's overkill?


zach-klippenstein:
Discussed the value of this change internally, and decided not to do it.


@kirillzh: 
> alternatively we could make LifecycleWorker open with lambdas as constructor arguments? 

@zach-klippenstein:
> I'm looking at this from an API maintenance perspective - every little thing like this is one thing that has to live forever once we hit 1.0. Does it provide enough value to justify that? Is making an anonymous object so gross that we need this function to eliminate it?
My instinct is that I haven't heard that to be true for anyone, so it's too early to land this.

@qrazhan:
>   it’s also simple enough that someone could write it if it became a pain

PR #654, opened 2019-10-09
Make RxWorker and Flowable.asWorker use Publisher instead of Flowable.

zach-klippenstein:
`Publisher` is the non-rx-specific interface from the `reactive-streams` library
that `Flowable` implements, and that all the `Flow` extensions for `Flowable`
are actually defined on.


zach-klippenstein:
Technically, to be more pure, we "should" have a `workflow-reactivestreams` module or something that does this, and depend on it from the rx module. This is what kotlinx.coroutines does. But that feels like overkill for now since (afaik) nobody's using this with reactive streams without Rx, and it will be easy to extract if the need arises.

PR #657, opened 2019-10-10
Introduces ViewRegistry.updateView.

rjrjr:
Updates a view if it's compatible with a given rendering,  otherwise replaces
it with one that is. Makes it much simpler to write custom containers.

### DIFF
--- a/kotlin/.idea/misc.xml
+++ b/kotlin/.idea/misc.xml
@@ -6,10 +6,11 @@
     </configurations>
   </component>
   <component name="EntryPointsManager">
-    <list size="3">
-      <item index="0" class="java.lang.String" itemvalue="com.squareup.moshi.ToJson" />
-      <item index="1" class="java.lang.String" itemvalue="org.openjdk.jmh.annotations.Benchmark" />
-      <item index="2" class="java.lang.String" itemvalue="org.openjdk.jmh.annotations.BenchmarkMode" />
+    <list size="4">
+      <item index="0" class="java.lang.String" itemvalue="androidx.lifecycle.OnLifecycleEvent" />
+      <item index="1" class="java.lang.String" itemvalue="com.squareup.moshi.ToJson" />
+      <item index="2" class="java.lang.String" itemvalue="org.openjdk.jmh.annotations.Benchmark" />
+      <item index="3" class="java.lang.String" itemvalue="org.openjdk.jmh.annotations.BenchmarkMode" />
     </list>
   </component>
   <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" project-jdk-name="1.8" project-jdk-type="JavaSDK">
--- a/kotlin/samples/dungeon/app/src/main/java/com/squareup/sample/dungeon/GameLayoutRunner.kt
+++ b/kotlin/samples/dungeon/app/src/main/java/com/squareup/sample/dungeon/GameLayoutRunner.kt
@@ -30,15 +30,13 @@ import com.squareup.workflow.ui.LayoutRunner
 import com.squareup.workflow.ui.LayoutRunner.Companion.bind
 import com.squareup.workflow.ui.ViewBinding
 import com.squareup.workflow.ui.ViewRegistry
-import com.squareup.workflow.ui.showRendering
 
 class GameLayoutRunner(
   view: View,
   private val viewRegistry: ViewRegistry
 ) : LayoutRunner<GameRendering> {
 
-  private val boardContainer: ViewGroup = view.findViewById(R.id.board_container)
-  private lateinit var boardView: View
+  private var boardView: View = view.findViewById(R.id.board_stub)
   private val moveLeft: View = view.findViewById(R.id.move_left)
   private val moveRight: View = view.findViewById(R.id.move_right)
   private val moveUp: View = view.findViewById(R.id.move_up)
@@ -54,13 +52,7 @@ class GameLayoutRunner(
   }
 
   override fun showRendering(rendering: GameRendering) {
-    if (!::boardView.isInitialized) {
-      boardView = viewRegistry.buildView(rendering.board, boardContainer)
-      boardContainer.addView(boardView)
-    } else {
-      boardView.showRendering(rendering.board)
-    }
-
+    boardView = viewRegistry.updateView(boardView, rendering.board)
     this.rendering = rendering
 
     // Disable the views if we don't have an event handler, e.g. when the game has finished.
--- a/kotlin/samples/dungeon/app/src/main/java/com/squareup/sample/dungeon/GameLayoutRunner.kt
+++ b/kotlin/samples/dungeon/app/src/main/java/com/squareup/sample/dungeon/GameLayoutRunner.kt
@@ -19,7 +19,6 @@ import android.view.MotionEvent.ACTION_DOWN
 import android.view.MotionEvent.ACTION_MASK
 import android.view.MotionEvent.ACTION_UP
 import android.view.View
-import android.view.ViewGroup
 import com.squareup.sample.dungeon.Direction.DOWN
 import com.squareup.sample.dungeon.Direction.LEFT
 import com.squareup.sample.dungeon.Direction.RIGHT
@@ -30,13 +29,14 @@ import com.squareup.workflow.ui.LayoutRunner
 import com.squareup.workflow.ui.LayoutRunner.Companion.bind
 import com.squareup.workflow.ui.ViewBinding
 import com.squareup.workflow.ui.ViewRegistry
+import com.squareup.workflow.ui.WorkflowViewStub
 
 class GameLayoutRunner(
   view: View,
   private val viewRegistry: ViewRegistry
 ) : LayoutRunner<GameRendering> {
 
-  private var boardView: View = view.findViewById(R.id.board_stub)
+  private val boardView: WorkflowViewStub = view.findViewById(R.id.board_stub)
   private val moveLeft: View = view.findViewById(R.id.move_left)
   private val moveRight: View = view.findViewById(R.id.move_right)
   private val moveUp: View = view.findViewById(R.id.move_up)
@@ -52,7 +52,7 @@ class GameLayoutRunner(
   }
 
   override fun showRendering(rendering: GameRendering) {
-    boardView = viewRegistry.updateView(boardView, rendering.board)
+    boardView.update(rendering.board, viewRegistry)
     this.rendering = rendering
 
     // Disable the views if we don't have an event handler, e.g. when the game has finished.
--- a/kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/AiWorkflow.kt
+++ b/kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/AiWorkflow.kt
@@ -84,7 +84,7 @@ class AiWorkflow(
       DOWN -> LEFT
       LEFT -> UP
     }
-//    println("AI changing direction from ${state.direction} to $newDirection")
+
     state = state.copy(direction = newDirection)
 
     null
--- a/kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/AiWorkflow.kt
+++ b/kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/AiWorkflow.kt
@@ -84,7 +84,7 @@ class AiWorkflow(
       DOWN -> LEFT
       LEFT -> UP
     }
-    println("AI changing direction from ${state.direction} to $newDirection")
+//    println("AI changing direction from ${state.direction} to $newDirection")
     state = state.copy(direction = newDirection)
 
     null
--- a/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ModalContainer.kt
+++ b/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ModalContainer.kt
@@ -49,21 +49,19 @@ abstract class ModalContainer<ModalRenderingT : Any>
   context: Context,
   attributeSet: AttributeSet? = null
 ) : FrameLayout(context, attributeSet) {
+  init {
+    // Stub base view will be replaced by registry.
+    this.addView(View(context))
+  }
 
-  private val base: View? get() = getChildAt(0)
+  private val baseView: View get() = getChildAt(0)
 
   private var dialogs: List<DialogRef<ModalRenderingT>> = emptyList()
 
   protected lateinit var registry: ViewRegistry
 
   protected fun update(newScreen: HasModals<*, ModalRenderingT>) {
-    base?.takeIf { it.canShowRendering(newScreen.baseScreen) }
-        ?.showRendering(newScreen.baseScreen)
-        ?: run {
-          removeAllViews()
-          val newBase = registry.buildView(newScreen.baseScreen, this)
-          addView(newBase)
-        }
+    registry.updateView(baseView, newScreen.baseScreen)
 
     val newDialogs = mutableListOf<DialogRef<ModalRenderingT>>()
     for ((i, modal) in newScreen.modals.withIndex()) {
--- a/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ModalContainer.kt
+++ b/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ModalContainer.kt
@@ -24,6 +24,8 @@ import android.os.Parcelable
 import android.os.Parcelable.Creator
 import android.util.AttributeSet
 import android.view.View
+import android.view.ViewGroup
+import android.view.ViewGroup.LayoutParams
 import android.widget.FrameLayout
 import androidx.annotation.IdRes
 import androidx.annotation.StyleRes
@@ -49,19 +51,21 @@ abstract class ModalContainer<ModalRenderingT : Any>
   context: Context,
   attributeSet: AttributeSet? = null
 ) : FrameLayout(context, attributeSet) {
-  init {
-    // Stub base view will be replaced by registry.
-    this.addView(View(context))
-  }
+  private val baseView: WorkflowViewStub = WorkflowViewStub(context).also {
+    it.layoutParams = (ViewGroup.LayoutParams(
+        ViewGroup.LayoutParams.MATCH_PARENT,
+        ViewGroup.LayoutParams.MATCH_PARENT
+    ))
 
-  private val baseView: View get() = getChildAt(0)
+    addView(it)
+  }
 
   private var dialogs: List<DialogRef<ModalRenderingT>> = emptyList()
 
   protected lateinit var registry: ViewRegistry
 
   protected fun update(newScreen: HasModals<*, ModalRenderingT>) {
-    registry.updateView(baseView, newScreen.baseScreen)
+    baseView.update(newScreen.baseScreen, registry)
 
     val newDialogs = mutableListOf<DialogRef<ModalRenderingT>>()
     for ((i, modal) in newScreen.modals.withIndex()) {
--- a/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewRegistry.kt
+++ b/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewRegistry.kt
@@ -84,36 +84,6 @@ class ViewRegistry private constructor(
   )
 
   /**
-   * Updates the given view if it [can show][View.canShowRendering] the given [rendering].
-   * If not, replaces it with one that can.
-   *
-   * Returns the view that is now showing the rendering. If the original view was in a parent
-   * and a new view was required for [rendering], the new view replaces the original in
-   * its parent.
-   */
-  fun <RenderingT : Any> updateView(
-    view: View,
-    rendering: RenderingT
-  ): View {
-    if (view.canShowRendering(rendering)) {
-      view.showRendering(rendering)
-      return view
-    }
-
-    return when (val parent = view.parent) {
-      is ViewGroup -> buildView(rendering, parent).also { newView ->
-        val index = parent.indexOfChild(view)
-        parent.removeView(view)
-
-        view.layoutParams
-            ?.let { parent.addView(newView, index, it) }
-            ?: run { parent.addView(newView, index) }
-      }
-      else -> buildView(rendering, view.context)
-    }
-  }
-
-  /**
    * Creates a [View] to display [initialRendering], which can be updated via calls
    * to [View.showRendering].
    */
@@ -127,7 +97,8 @@ class ViewRegistry private constructor(
         ?.buildView(this, initialRendering, contextForNewView, container)
         ?.apply {
           checkNotNull(showRenderingTag?.initialRendering) {
-            "View.bindShowRendering must be called for $this."
+            "View.bindShowRendering should have been called for $this, typically when its " +
+                "${ViewBinding::class.java.name} is invoked."
           }
         }
         ?: throw IllegalArgumentException(
--- a/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewRegistry.kt
+++ b/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewRegistry.kt
@@ -84,6 +84,36 @@ class ViewRegistry private constructor(
   )
 
   /**
+   * Updates the given view if it [can show][View.canShowRendering] the given [rendering].
+   * If not, replaces it with one that can.
+   *
+   * Returns the view that is now showing the rendering. If the original view was in a parent
+   * and a new view was required for [rendering], the new view replaces the original in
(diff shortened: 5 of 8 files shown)


kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/AiWorkflow.kt
```diff
@@ -84,7 +84,7 @@ class AiWorkflow(
       DOWN -> LEFT
       LEFT -> UP
     }
-    println("AI changing direction from ${state.direction} to $newDirection")
+//    println("AI changing direction from ${state.direction} to $newDirection")
```
SalvatoreT: `//`?

rjrjr: It's SO NOISY, and since it's not Timber I can't make it `.v` or whatever.

@zach-klippenstein, your call on what I should really do with it. Happy to add Timber.

kotlin/samples/dungeon/app/src/main/java/com/squareup/sample/dungeon/GameLayoutRunner.kt
```diff
@@ -54,13 +52,7 @@ class GameLayoutRunner(
...
 
   override fun showRendering(rendering: GameRendering) {
-    if (!::boardView.isInitialized) {
-      boardView = viewRegistry.buildView(rendering.board, boardContainer)
-      boardContainer.addView(boardView)
-    } else {
-      boardView.showRendering(rendering.board)
-    }
-
+    boardView = viewRegistry.updateView(boardView, rendering.board)
```
zach-klippenstein: For this case, we could make a property delegate like:

```kotlin
private val boardView by viewRegistry.bindStub<GameBoard>(R.id.board_stub)
// Or
private val boardView = viewRegistry.bindStub<GameBoard>(R.id board_stub)

…

boardView.update(rendering.board)
// Or even
boardView = rendering.board
```

kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/AiWorkflow.kt
zach-klippenstein: Just delete this logging, it is redundant with the diagnostic listener stuff anyway.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ModalContainer.kt
```diff
@@ -49,21 +49,19 @@ abstract class ModalContainer<ModalRenderingT : Any>
   context: Context,
   attributeSet: AttributeSet? = null
 ) : FrameLayout(context, attributeSet) {
+  init {
+    // Stub base view will be replaced by registry.
+    this.addView(View(context))
+  }
 
-  private val base: View? get() = getChildAt(0)
+  private val baseView: View get() = getChildAt(0)
```
zach-klippenstein: This now seems over complicated, can we just do

```kotlin
private var baseView = View(context)
  .also { addView(it) }
```

And then the same assign-from-update below as for the `boardView` above?

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewRegistry.kt
```diff
@@ -83,6 +83,36 @@ class ViewRegistry private constructor(
...
+    }
+
+    return when (val parent = view.parent) {
+      is ViewGroup -> buildView(rendering, parent).also { newView ->
+        val index = parent.indexOfChild(view)
+        parent.removeView(view)
+
+        view.layoutParams
+            ?.let { parent.addView(newView, index, it) }
+            ?: run { parent.addView(newView, index) }
```
zach-klippenstein: This is cool, but almost feels too magic… don't have any better ideas atm though.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ModalContainer.kt
```diff
@@ -49,21 +49,19 @@ abstract class ModalContainer<ModalRenderingT : Any>
   context: Context,
   attributeSet: AttributeSet? = null
 ) : FrameLayout(context, attributeSet) {
+  init {
+    // Stub base view will be replaced by registry.
+    this.addView(View(context))
```
zach-klippenstein: Use `ViewStub` to help make the intent more clear?

kotlin/samples/dungeon/app/src/main/java/com/squareup/sample/dungeon/GameLayoutRunner.kt
zach-klippenstein: But we still need an update method for the modal case below. Seems easy to forget to reassign from the return value - instead the update can just take a `KMutableProperty` and then you just pass in the property with `::baseView`.

zach-klippenstein: But we still need an update method for the modal case below. Seems easy to forget to reassign from the return value - instead the update can just take a `KMutableProperty` and then you just pass in the property with `::baseView`.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewRegistry.kt
zach-klippenstein: Yep, just looked at the impl in `ViewStubCompat`, looks very similar.

rjrjr: Yes, I took the code from ViewStub. But no, we can't just use ViewStub, if that's where you're going. It's single shot, and it's strictly related to layouts.

rjrjr:
I get the direction of your BindingStub thing, but it feels a bit complicated to me. It's simpler if we always pass the ViewRegistry via the update method. Self contained, no need for ViewRegistry to be aware of any of this.

```kotlin
class WorkflowViewStub(...) : View(...) {
  private var _actual : View = this

  val actual: View? get() = when(this) {
    is WorkflowViewStub -> null
    else -> _actual
  }

  fun update(
    rendering: Any, 
    registry: ViewRegistry
  ) : View {
    // same update / replacement logic
    return actual
  }

  // Should also proxy setVisibility like ViewStub does.
  // I don't see any reason to use a WeakRef though.
}
```

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewRegistry.kt
zach-klippenstein: Didn't mean use it, just use it as the placeholder since it's semantically very similar (and already zero-sized). But don't feel too strongly.

zach-klippenstein:
I wasn't thinking a custom view, just a wrapper type so `showRendering` doesn't need to ever touch the view directly. But yea that might be too magic. How do you feel about the property reference?

rjrjr:
Not sure we need something like the property reference once the stub is working, since it can maintain the reference to the swizzled in views even after the stub is out of the view hierarchy. 🤞

zach-klippenstein:
What's the memory cost of keeping a detached `View` around just for this? E.g. if we have a list of these, or a deep tree of nested container views?

kotlin/samples/dungeon/app/src/main/java/com/squareup/sample/dungeon/GameLayoutRunner.kt
rjrjr: Finally starting to grok this, maybe I do like it better than `WorkflowViewStub`. Also, holy shit, `KMutableProperty` is our ticket to Swift inout parameters, no?

rjrjr: I'll finish `WorkflowViewStub` before exploring this.

rjrjr:
_What's the memory cost of keeping a detached View around just for this?_ I'll dig to verify this, but my impression is that an undrawn view is pretty trivial, very little actual state.

kotlin/samples/dungeon/app/src/main/java/com/squareup/sample/dungeon/GameLayoutRunner.kt
zach-klippenstein: Only a tiny aspect of them, and the least interesting part. An early implementation of what is now called `Mutator` used them, but that wasn't very interesting because it wasn't public-facing code. 

The real win with `inout` is that you can pass structs that have `var` properties, assign to those properties like they were vars on a mutable reference type, and automatically see those changes in the caller, all without losing any purity.

rjrjr:
Updated (unsquashed), `WorkflowViewStub` replaces `ViewRegistry.updateView`. 

I'm liking this, it's really simple and a lot harder to misuse. `android.view.View` looks a bit more heavy weight than I had hoped, with a _lot_ of instance fields, but they're all value types. I think it's a reasonable tradeoff. WDYT?

zach-klippenstein:
So unfortunate that `View` isn't just an interface or something. It's probably fine, I don't suspect there will be a ton of these things allocated at once (maybe on the order of about 10 for most UIs?), and they are long-lived allocations relative to frame rate so it's not gonna cause too much churn. The API usability improvement seems like a huge win and probably worth the trade-off.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowViewStub.kt
```diff
@@ -0,0 +1,58 @@
+package com.squareup.workflow.ui
+
+import android.content.Context
+import android.util.AttributeSet
+import android.view.View
+import android.view.ViewGroup
+
+class WorkflowViewStub(
```
rjrjr: doc

kotlin/.idea/misc.xml
```diff
@@ -6,10 +6,11 @@
...
   <component name="EntryPointsManager">
-    <list size="3">
-      <item index="0" class="java.lang.String" itemvalue="com.squareup.moshi.ToJson" />
-      <item index="1" class="java.lang.String" itemvalue="org.openjdk.jmh.annotations.Benchmark" />
-      <item index="2" class="java.lang.String" itemvalue="org.openjdk.jmh.annotations.BenchmarkMode" />
+    <list size="4">
+      <item index="0" class="java.lang.String" itemvalue="androidx.lifecycle.OnLifecycleEvent" />
+      <item index="1" class="java.lang.String" itemvalue="com.squareup.moshi.ToJson" />
+      <item index="2" class="java.lang.String" itemvalue="org.openjdk.jmh.annotations.Benchmark" />
+      <item index="3" class="java.lang.String" itemvalue="org.openjdk.jmh.annotations.BenchmarkMode" />
```
zach-klippenstein: Well at least they're sorted alphabetically…

kotlin/samples/dungeon/app/src/main/java/com/squareup/sample/dungeon/GameLayoutRunner.kt
```diff
@@ -54,13 +53,7 @@ class GameLayoutRunner(
...
 
   override fun showRendering(rendering: GameRendering) {
-    if (!::boardView.isInitialized) {
-      boardView = viewRegistry.buildView(rendering.board, boardContainer)
-      boardContainer.addView(boardView)
-    } else {
-      boardView.showRendering(rendering.board)
-    }
-
+    boardView.update(rendering.board, viewRegistry)
```
zach-klippenstein: 😍

kotlin/samples/dungeon/app/src/main/java/com/squareup/sample/dungeon/GameLayoutRunner.kt
```diff
@@ -30,15 +30,14 @@ import com.squareup.workflow.ui.LayoutRunner
...
+import com.squareup.workflow.ui.WorkflowViewStub
 
 class GameLayoutRunner(
   view: View,
   private val viewRegistry: ViewRegistry
 ) : LayoutRunner<GameRendering> {
 
-  private val boardContainer: ViewGroup = view.findViewById(R.id.board_container)
-  private lateinit var boardView: View
+  private val boardView: WorkflowViewStub = view.findViewById(R.id.board_stub)
```
zach-klippenstein: So much cleaner.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ModalContainer.kt
```diff
@@ -49,21 +51,21 @@ abstract class ModalContainer<ModalRenderingT : Any>
...
 
   protected fun update(newScreen: HasModals<*, ModalRenderingT>) {
-    base?.takeIf { it.canShowRendering(newScreen.baseScreen) }
-        ?.showRendering(newScreen.baseScreen)
-        ?: run {
-          removeAllViews()
-          val newBase = registry.buildView(newScreen.baseScreen, this)
-          addView(newBase)
-        }
+    baseView.update(newScreen.baseScreen, registry)
```
zach-klippenstein: Omg yes.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowViewStub.kt
```diff
@@ -0,0 +1,62 @@
+package com.squareup.workflow.ui
+
+import android.content.Context
+import android.util.AttributeSet
+import android.view.View
+import android.view.ViewGroup
+
+/**
+ * A placeholder [View] that can replace itself with ones driven by workflow renderings,
+ * similar to [android.view.ViewStub].
```
zach-klippenstein:
```suggestion
 * similar to [android.view.ViewStub].
 *
 * ## Usage
 *
 * In the XML layout for your container view, place a [WorkflowViewStub] where you want child renderings
 * to be displayed. E.g.:
 *
 * ```
 * <LinearLayout
 *     android:…
 *     >
 *         < com.squareup.workflow.ui.WorkflowViewStub
 *             android:id="@+id/child_stub"
 *             />
 *    …
 * ```
 *
 * Then in your [LayoutRunner], get the `ViewRegistry` in your constructor:
 *
 * ```
 * class YourLayoutRunner(view: View, private val viewRegistry: ViewRegistry) {
 * ```
 *
 * pull the view out with `findViewById` like any other view:
 *
 * ```
 * private val child = view.findViewById<WorkflowViewStub>(R.id.child_stub)
 * ```
 *
 * and update it in your `showRendering` method:
 *
 * ```
 * override fun showRendering(rendering: YourRendering) {
 *   child.update(rendering.childRendering, viewRegistry)
 * }
 * ```
```

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowViewStub.kt
```diff
@@ -0,0 +1,62 @@
+package com.squareup.workflow.ui
+
+import android.content.Context
+import android.util.AttributeSet
+import android.view.View
+import android.view.ViewGroup
+
+/**
+ * A placeholder [View] that can replace itself with ones driven by workflow renderings,
```
zach-klippenstein:
```suggestion
 * A placeholder [View] that immediately replaces itself with ones driven by workflow renderings,
```

zach-klippenstein:
```suggestion
 * A placeholder [View] that can replace itself with ones driven by workflow renderings and a [ViewRegistry],
```

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowViewStub.kt
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   * By starting as our own delegate, we make the parent checking code in [update] simpler.
+   */
+  private var _actual: View = this
+
+  /**
+   * On-demand access to the delegate established by the last call to [update],
+   * or null if none has been set yet.
+   */
+  val actual: View? get() = _actual.takeIf { it !is WorkflowViewStub }
```
zach-klippenstein: Nice

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowViewStub.kt
```diff
@@ -0,0 +1,62 @@
...
+
+  /**
+   * Replaces this view with one that can display [rendering]. If it
+   * has already been replaced, updates that replacement if it [canShowRendering].
+   * If it can't, a new replacement is put in place.
```
zach-klippenstein: "Place" is starting to sound funny after reading this 🤪

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowViewStub.kt
```diff
@@ -0,0 +1,62 @@
...
+   * @return the view that showed [rendering]
+   * @throws RuntimeException if [rendering] cannot be displayed, typically indicating
+   * a misconfigured [registry]
+   */
+  fun update(
+    rendering: Any,
+    registry: ViewRegistry
+  ): View {
+    _actual.takeIf { it.canShowRendering(rendering) }
+        ?.let {
```
zach-klippenstein: I liked this better as an vanilla `if` statement, I don't think the fancy scoping functions add any value.

I'm curious if the compiler or R8 or something can optimize this down to a single `if`, instead of a local nullable variable with a null check. But I think the `if` is just more readable regardless.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowViewStub.kt
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+    return when (val parent = _actual.parent) {
+      is ViewGroup -> registry.buildView(rendering, parent).also { newView ->
+        val index = parent.indexOfChild(_actual)
+        parent.removeView(_actual)
+
+        _actual.layoutParams
+            ?.let { parent.addView(newView, index, it) }
+            ?: run { parent.addView(newView, index) }
```
zach-klippenstein: 😬 …ok, I guess this is pretty readable.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowViewStub.kt
```diff
@@ -0,0 +1,62 @@
...
+      }
```
zach-klippenstein: Nit: Maybe factor the body of this case out into a separate function `buildAndReplaceView` so this `when` is nice and scannable, especially since the `.also` at the end kind of throws you for a loop because you have to go back up to the top of the `when` to figure out what it actually refers to.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowViewStub.kt
```diff
@@ -0,0 +1,62 @@
...
+   */
+  val actual: View? get() = _actual.takeIf { it !is WorkflowViewStub }
+
+  /**
+   * Replaces this view with one that can display [rendering]. If it
+   * has already been replaced, updates that replacement if it [canShowRendering].
+   * If it can't, a new replacement is put in place.
+   *
+   * @return the view that showed [rendering]
+   * @throws RuntimeException if [rendering] cannot be displayed, typically indicating
```
zach-klippenstein: Where does this exception actually get thrown? I don't see it. If it actually is thrown, could we throw something more specific, like `IllegalStateException` or `AssertionError`?

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowViewStub.kt
rjrjr: Thanks!

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowViewStub.kt
rjrjr: I strongly prefer the mapping functions when dealing with mutable fields, even when smart casting isn't particularly needed. And I find this very readable.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowViewStub.kt
rjrjr: It's thrown by ViewRegistry, and could be either `IllegalStateException` or `IllegalArgumentException`. See [ViewRegistry.buildView], line 90. I thought about changing the `check` call there to `require`, but it just seemed wrong. And wrapping and re-throwing here to clean up the kdoc seems silly.

rjrjr:
> So unfortunate that View isn't just an interface or something. It's probably fine, I don't suspect there will be a ton of these things allocated at once (maybe on the order of about 10 for most UIs?), and they are long-lived allocations relative to frame rate so it's not gonna cause too much churn. The API usability improvement seems like a huge win and probably worth the trade-off.

The more I think about this, the more I'm at piece with it. Compare it with a `ViewGroup`: there would be even more state, and an unnecessary addition to the depth of the view stack.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowViewStub.kt
rjrjr:
```kotlin
/**
 * A placeholder [View] that can replace itself with ones driven by workflow renderings,
 * similar to [android.view.ViewStub].
 *
 * ## Usage
 *
 * In the XML layout for your container view, place a [WorkflowViewStub] where
 * you want child renderings to be displayed. E.g.:
 *
 *    <LinearLayout
 *        android:…
 *        >
 *            <com.squareup.workflow.ui.WorkflowViewStub
 *                android:id="@+id/child_stub"
 *                />
 *       …
 *
 * Then in your [LayoutRunner],
 *   - get the `ViewRegistry` in your constructor
 *   - pull the view out with `findViewById` like any other view
 *   - and update it in your `showRendering` method:
 *
 * ```
 *     class YourLayoutRunner(
 *       view: View, private
 *       val viewRegistry: ViewRegistry
 *     ) {
 *       private val child = view.findViewById<WorkflowViewStub>(R.id.child_stub)
 *
 *       override fun showRendering(rendering: YourRendering) {
 *         child.update(rendering.childRendering, viewRegistry)
 *       }
 *     }
 * ```
 */
```

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowViewStub.kt
rjrjr:
```
   * Replaces this view with one that can display [rendering]. If the receiver
   * has already been replaced, updates the replacement if it [canShowRendering].
   * If the current replacement can't handle [rendering], a new view is put in place.
```

rjrjr:
Cool, this PR breaks `TracingDiagnosticListenerTest.golden value`. 

```
org.junit.ComparisonFailure: expected:<...filing"}},
{"name":"[GC detected","cat":"system","ph":"i","ts":0,"pid":0,"tid":0,"s":"g","args":{"freeMemory":42,"totalMemory":43}},
{"name":"used/free memory","ph":"C","ts":0,"pid":0,"tid":0,"args":{"usedMemory":1,"freeMemory":42}},
{"name":"]TestWorkflow (0)","c...> but was:<...filing"}},
{"name":"[]TestWorkflow (0)","c...>
```

rjrjr:
Actually, that test is broken on master.

rjrjr:
Update ignores flaky test (#664), improves docs, factors out build-and-replace method.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowViewStub.kt
zach-klippenstein: I really hate this, but it's correct so I can't stop you.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowViewStub.kt
zach-klippenstein: Document those two exceptions directly then? Don't feel strongly though.

rjrjr: Saw this after merging. I'll figure out something, agree this is weird.

PR #663, opened 2019-10-11
Update tutorials for Swift 5/ReactiveSwift 6.0

### LINKED ISSUE #655: Issues with Swift samples
There are a few issues I ran into when checking out the Swift samples (/swift/Samples)

### 1. Build issues
[The `Tutorial` app readme](https://github.com/square/workflow/tree/master/swift/Samples/Tutorial) states you need to run the following:

1. `bundle install`
2. `bundle exec pod install`
3. Open `Tutorial.xcworkspace`

Unfortunately the Tutorial scheme does not build in Xcode 11.1 (11A1027), due to errors such as:
`No such module 'Result'` and `Use of undeclared type 'NoError'`

I get the same errors in the TicTacToe app. The SampleApp has no build issues.

### 2. Tutorial Tooling issues
The [first chapter of tutorial](https://github.com/square/workflow/blob/master/swift/Samples/Tutorial/Tutorial1.md) also mentions creating a new file using Xcode templates, unfortunately it doesn't mention that these first need to be installed nor how this is done. This is covered in the [Tooling readme](https://github.com/square/workflow/blob/96ce04cf56caba25d024a26c0da94ddcf7ca1355/docs/tutorial/tooling.md), so a link to it should do the trick.

AquaGeek:
Fixes #655 

* Remove all the `import Result`s from the tutorials and tutorial code
* Use `Never` instead of `NoError`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -182,3 +182,18 @@ matrix:
         - swift build --verbose
       script:
         - swift test
+
+    - language: swift
+      name: "Swift - Tutorial - Cocoapods (Xcode 10.2)"
+      os: osx
+      osx_image: xcode10.2
+      xcode_workspace: Tutorial.xcworkspace
+      xcode_scheme: Tutorial
+      xcode_destination: platform=iOS Simulator,OS=12.2,name=iPad Pro (9.7-inch)
+      before_install:
+        - gem update --system
+        - gem install bundler
+        - bundle install
+        - cd swift/Samples/Tutorial
+        - bundle exec pod repo update
+        - bundle exec pod install
--- a/swift/Samples/Tutorial/Frameworks/Tutorial1Complete/Sources/Welcome/WelcomeWorkflow.swift
+++ b/swift/Samples/Tutorial/Frameworks/Tutorial1Complete/Sources/Welcome/WelcomeWorkflow.swift
@@ -16,7 +16,6 @@
 import Workflow
 import WorkflowUI
 import ReactiveSwift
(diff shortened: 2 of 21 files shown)


AquaGeek:
We should probably figure out a CI story for the tutorials so this doesn't break again…

zach-klippenstein:
Can we build the tutorials in CI to prevent this in the future?

zach-klippenstein:
Yea, can we just build the samples in CI? Don't have to run them or anything.

AquaGeek:
> Yea, can we just build the samples in CI? Don't have to run them or anything.

Seems reasonable? 🤷‍♀

zach-klippenstein:
> Seems reasonable? 🤷‍♀

Do you want to take a stab at adding this? Should just be a copy-paste of one of the existing iOS jobs in `.travis.yml`.

.travis.yml
```diff
@@ -182,3 +182,18 @@ matrix:
...
+      xcode_workspace: Tutorial.xcworkspace
+      xcode_scheme: Tutorial
+      xcode_destination: platform=iOS Simulator,OS=12.2,name=iPad Pro (9.7-inch)
+      before_install:
+        - gem update --system
+        - gem install bundler
+        - bundle install
+        - cd swift/Samples/Tutorial
+        - bundle exec pod repo update
+        - bundle exec pod install
```
zach-klippenstein: Sweet! Thanks for doing this.

PR #666, opened 2019-10-11
Adds getRendering, getShowRendering.

rjrjr:
The rendering in a view's `ViewShowRendering` tag is now updated on each call
to `View.showRendering`. Access to that value and the `showRendering` function
itself is improved with new `View.getRendering` and `View.getShowRendering`
methods. With their introduction, we're able to make `View.showRenderingTag`
private.

Also updates `WorkflowLayout` to use `WorkflowStubView`. It's pretty thrilling
that this didn't break the view persistence code. Wasn't able to pull off the
same trick for BackStackContainer or ModalContainer, so the more fundamental
`ViewRegistry` and `ViewShowRendering.kt` methods are staying public. Punched
up their documentation. (Had to go with `fun` rather than `val` on the new
code due to all the param types.)

Another point for discussion: thought about changing `GameLayoutRunner` to
take advantage of the new methods instead of storing its own reference to the
latest rendering, but backed off because it just seemed sketchy, too magical
to encourage as a general practice.

So given that, why open up the API at all? We need access to both of  these
things when implementing the various containers, and expect that other
container authors will too.

Well, except for the change to keep the View's reference to the rendering  up
to date. So far we've been just fine having access to nothing but the original
rendering, but letting it stay stale like that just seems like a landmine.

### DIFF
--- a/kotlin/samples/tictactoe/app/src/androidTest/java/com/squareup/sample/TicTacToeEspressoTest.kt
+++ b/kotlin/samples/tictactoe/app/src/androidTest/java/com/squareup/sample/TicTacToeEspressoTest.kt
@@ -2,6 +2,7 @@ package com.squareup.sample
 
 import android.content.pm.ActivityInfo.SCREEN_ORIENTATION_LANDSCAPE
 import android.content.pm.ActivityInfo.SCREEN_ORIENTATION_PORTRAIT
+import android.view.View
 import androidx.test.core.app.ActivityScenario
 import androidx.test.core.app.ActivityScenario.launch
 import androidx.test.espresso.Espresso.onView
@@ -12,12 +13,18 @@ import androidx.test.espresso.action.ViewActions.click
 import androidx.test.espresso.action.ViewActions.closeSoftKeyboard
 import androidx.test.espresso.action.ViewActions.typeText
 import androidx.test.espresso.assertion.ViewAssertions.matches
+import androidx.test.espresso.matcher.ViewMatchers.hasDescendant
 import androidx.test.espresso.matcher.ViewMatchers.isDisplayed
 import androidx.test.espresso.matcher.ViewMatchers.withId
 import androidx.test.espresso.matcher.ViewMatchers.withText
 import androidx.test.ext.junit.runners.AndroidJUnit4
+import com.google.common.truth.Truth.assertThat
+import com.squareup.sample.gameworkflow.GamePlayScreen
+import com.squareup.sample.gameworkflow.Player
+import com.squareup.sample.gameworkflow.symbol
 import com.squareup.sample.mainactivity.MainActivity
 import com.squareup.sample.tictactoe.R
+import com.squareup.workflow.ui.getRendering
 import org.junit.After
 import org.junit.Before
 import org.junit.Test
@@ -32,19 +39,57 @@ class TicTacToeEspressoTest {
   fun setUp() {
     scenario = launch(MainActivity::class.java)
         .apply {
-          onActivity {
+          onActivity { activity ->
             IdlingRegistry.getInstance()
-                .register(it.idlingResource)
-            it.requestedOrientation = SCREEN_ORIENTATION_PORTRAIT
+                .register(activity.idlingResource)
+            activity.requestedOrientation = SCREEN_ORIENTATION_PORTRAIT
           }
         }
   }
 
   @After
   fun tearDown() {
-    scenario.onActivity {
+    scenario.onActivity { activity ->
       IdlingRegistry.getInstance()
-          .unregister(it.idlingResource)
+          .unregister(activity.idlingResource)
+    }
+  }
+
+  @Test fun showRenderingTagStaysFresh() {
+    // Start a game so that there's something interesting in the Activity window.
+    // (Prior screens are all in a dialog window.)
+
+    onView(withId(R.id.login_email)).type("foo@bar")
+    onView(withId(R.id.login_password)).type("password")
+    onView(withId(R.id.login_button)).perform(click())
+
+    onView(withId(R.id.start_game)).perform(click())
+
+    // Why should I learn how to write a matcher when I can just grab the activity
+    // and work with it directly?
+    scenario.onActivity { activity ->
+      val button = activity.findViewById<View>(R.id.game_play_board)
+      val rendering = (button.parent as View).getRendering<GamePlayScreen>()!!
+      assertThat(rendering.gameState.playing).isSameInstanceAs(Player.X)
+
+      // Make a move.
+      rendering.onClick(0, 0)
+    }
+
+    // I'm not an animal, though. Pop back out to the test to check that the update
+    // has happened, to make sure the idle check is allowed to do its thing. (Didn't
+    // actually seem to be necessary, originally did everything synchronously in the
+    // lambda above and it all worked just fine. But that seems like a land mine.)
+
+    onView(withId(R.id.game_play_toolbar))
+        .check(matches(hasDescendant(withText("O, place your ${Player.O.symbol}"))))
+
+    // Now that we're confident the views have updated, back to the activity
+    // to mess with what should be the updated rendering.
+    scenario.onActivity { activity ->
+      val button = activity.findViewById<View>(R.id.game_play_board)
+      val rendering = (button.parent as View).getRendering<GamePlayScreen>()!!
+      assertThat(rendering.gameState.playing).isSameInstanceAs(Player.O)
     }
   }
 
--- a/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/LayoutRunner.kt
+++ b/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/LayoutRunner.kt
@@ -24,10 +24,10 @@ import com.squareup.workflow.ui.LayoutRunner.Companion.bind
 import kotlin.reflect.KClass
 
 /**
- * An object that handles [View.showRendering] calls for a view inflated
- * from a layout resource in response to [ViewRegistry.buildView].
- * (Use [BuilderBinding] if you want to build views from code rather than
- * layouts.)
+ * A delegate that implements a [showRendering] method to be called when a workflow rendering
+ * of type [RenderingT] is ready to be displayed in a view inflated from a layout resource
+ * by a [ViewRegistry]. (Use [BuilderBinding] if you want to build views from code rather
+ * than layouts.)
  *
  * Typical usage is to have a [LayoutRunner]'s `companion object` implement
  * [ViewBinding] by delegating to [LayoutRunner.bind], specifying the layout resource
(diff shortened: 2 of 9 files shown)


zach-klippenstein:
> this didn't break the view persistence code

Wow, I am actually surprised it just worked.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/LayoutRunner.kt
```diff
@@ -24,10 +24,9 @@ import com.squareup.workflow.ui.LayoutRunner.Companion.bind
 import kotlin.reflect.KClass
 
 /**
- * An object that handles [View.showRendering] calls for a view inflated
- * from a layout resource in response to [ViewRegistry.buildView].
- * (Use [BuilderBinding] if you want to build views from code rather than
- * layouts.)
+ * The delegate for [View.showRendering] calls made on a view inflated
+ * from a layout resource in response by a [ViewRegistry]. (Use [BuilderBinding]
```
rjrjr: gibberish

kotlin/samples/tictactoe/app/src/androidTest/java/com/squareup/sample/TicTacToeEspressoTest.kt
```diff
@@ -32,19 +39,57 @@ class TicTacToeEspressoTest {
...
+    onView(withId(R.id.login_password)).type("password")
+    onView(withId(R.id.login_button)).perform(click())
+
+    onView(withId(R.id.start_game)).perform(click())
+
+    // Why should I learn how to write a matcher when I can just grab the activity
+    // and work with it directly?
+    scenario.onActivity { activity ->
+      val button = activity.findViewById<View>(R.id.game_play_board)
+      val rendering = (button.parent as View).getRendering<GamePlayScreen>()!!
```
zach-klippenstein: Neat!

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/LayoutRunner.kt
```diff
@@ -53,7 +52,7 @@ import kotlin.reflect.KClass
  *        NewGameLayoutRunner, GamePlayLayoutRunner, GameOverLayoutRunner
  *    )
  *
- * Also note that two flavors of [contructor][LayoutRunner.Binding.runnerConstructor]
+ * Also note that two flavors of [constructor][LayoutRunner.Binding.runnerConstructor]
  * are accepted by [bind]. Every [LayoutRunner] constructor must accept an [View].
```
rjrjr: "an View" 🤢

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewRegistry.kt
```diff
@@ -86,8 +87,14 @@ class ViewRegistry private constructor(
   /**
    * Creates a [View] to display [initialRendering], which can be updated via calls
    * to [View.showRendering].
+   *
+   * @throws IllegalArgumentException if no binding can be find for type [RenderingT]
```
rjrjr: can be find

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewShowRendering.kt
```diff
@@ -58,30 +58,60 @@ fun <RenderingT : Any> View.bindShowRendering(
...
+ *
+ * @throws IllegalStateException if [bindShowRendering] has not been called.
  */
 fun <RenderingT : Any> View.showRendering(rendering: RenderingT) {
   showRenderingTag
       ?.let { tag ->
-        check(tag.initialRendering.matches(rendering)) {
-          "Expected $this to be able show $rendering, should have matched ${tag.initialRendering}."
+        check(tag.showing.matches(rendering)) {
+          "Expected $this to be able show $rendering, should have matched ${tag.showing}. " +
```
rjrjr: to be able show

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowLayout.kt
```diff
@@ -35,8 +36,16 @@ class WorkflowLayout(
   context: Context,
   attributeSet: AttributeSet? = null
 ) : FrameLayout(context, attributeSet) {
+  private val showing: WorkflowViewStub = WorkflowViewStub(context).also {
+    it.layoutParams = (ViewGroup.LayoutParams(
+        ViewGroup.LayoutParams.MATCH_PARENT,
+        ViewGroup.LayoutParams.MATCH_PARENT
+    ))
```
zach-klippenstein: `addView` takes a `LayoutParams` parameter, why do you have to set it explicitly here?

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowLayout.kt
```diff
@@ -57,25 +66,17 @@ class WorkflowLayout(
...
-          return
-        }
-
-    removeAllViews()
-    val newView = registry.buildView(newRendering, this)
+    showing.update(newRendering, registry)
     restoredChildState?.let { restoredState ->
       restoredChildState = null
-      newView.restoreHierarchyState(restoredState)
+      showing.actual!!.restoreHierarchyState(restoredState)
```
zach-klippenstein: Could we override `restoreHierarchyState` on `WorkflowViewStub` to automatically delegate to `_actual` if it's not self-referential? Seems like this would be easy to get wrong.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowLayout.kt
```diff
@@ -57,25 +66,17 @@ class WorkflowLayout(
...
     }
-    addView(newView)
   }
 
   override fun onSaveInstanceState(): Parcelable? {
     return SavedState(
         super.onSaveInstanceState()!!,
-        SparseArray<Parcelable>().also { array -> showing?.saveHierarchyState(array) }
+        SparseArray<Parcelable>().also { array -> showing.actual?.saveHierarchyState(array) }
```
zach-klippenstein: Ditto for saving.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewRegistry.kt
```diff
@@ -86,8 +87,14 @@ class ViewRegistry private constructor(
...
+   *
+   * @throws IllegalStateException if the matching [ViewBinding] fails to call
+   * [View.bindShowRendering] when constructing the view
    */
-  fun <RenderingT : Any> buildView(
+  @PublishedApi
+  internal inline fun <reified RenderingT : Any> buildView(
```
zach-klippenstein: This method body is non-trivial, could we just pass a `KClass` to it and make a reified inline wrapper?

rjrjr:
> Also updates WorkflowLayout to use WorkflowStubView. It's pretty thrilling that this didn't break the view persistence code. Given that, this PR makes ViewRegistry.buildView internal — we can always open it up later if a use case appears.

Missed a couple of `buildView` spots, `ModalViewContainer` in particular. I imagine that I _could_ make that work with `WorkflowViewStub`, but it would just make things harder to understand. So, I think the `ViewRegistry` methods should stay public after all.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewShowRendering.kt
```diff
@@ -58,30 +58,60 @@ fun <RenderingT : Any> View.bindShowRendering(
...
+/**
+ * Returns the most recent rendering shown by this view, or null if [bindShowRendering]
+ * has never been called.
+ *
+ * @throws ClassCastException if a rendering is found but is not of type [RenderingT]
+ */
+inline fun <reified RenderingT : Any> View.getRendering(): RenderingT? {
+  return when (val showing = showRenderingTag?.showing) {
+    null -> null
+    else -> showing as RenderingT
```
zach-klippenstein: Is this function inline just for this cast? What happens if you just cast without it being inline?

rjrjr:
Update fixes a lot of docs and error messages.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewShowRendering.kt
rjrjr: I _think_ it won't compile. I'll check again. If it's just a warning, I'll try to undo some of the inlining.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowLayout.kt
rjrjr: It's not necessary in 95% of containers, e.g. note that none of the modal container do. Where it is necessary, like here or in `BackStackContainer`, I don't think there's a general solution. The two use cases are way different.

rjrjr: Oh, right, and I didn't even use the stub in BackStackContainer, so I'm down to a single use case. Definitely not ready to generalize, and pretty confident that day to day feature developers don't care.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowLayout.kt
rjrjr: Oh yeah! Because I didn't notice that until today. Will update everywhere.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewShowRendering.kt
rjrjr: Okay, it's just a warning. Sanity check please: that should in theory be okay since a real cast will happen as soon as the return value is assigned, right?

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewRegistry.kt
rjrjr: I suppressed a couple of warnings and was able to get rid of all the inlining.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewShowRendering.kt
zach-klippenstein: Yes, I believe the cast error will come from the assignment.

rjrjr:
Update backs out all the new inlining.

cgruber:
😈

PR #673, opened 2019-10-14
Make onEvent, makeActionSink, and makeEventSink include what action was lost to when they receive more than one event.

### LINKED ISSUE #539: Report who won the race when an extra event is received
Don't know why I didn't do this initially, but I think it would be helpful for debugging to include the deferred itself in this message, to see what we lost to.

_Originally posted by @zach-klippenstein in https://github.com/square/workflow/pull/537_

zach-klippenstein:
Fixes #539.

### DIFF
--- a/kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/RealRenderContext.kt
+++ b/kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/RealRenderContext.kt
@@ -26,6 +26,7 @@ import com.squareup.workflow.WorkflowAction
 import com.squareup.workflow.internal.Behavior.WorkerCase
 import com.squareup.workflow.internal.Behavior.WorkflowOutputCase
 import kotlinx.coroutines.CompletableDeferred
+import kotlinx.coroutines.ExperimentalCoroutinesApi
 
 /**
  * An implementation of [RenderContext] that builds a [Behavior] via [buildBehavior].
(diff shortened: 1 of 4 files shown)


kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/RealRenderContext.kt
```diff
@@ -52,31 +53,37 @@ class RealRenderContext<StateT, OutputT : Any>(
...
+  @UseExperimental(ExperimentalCoroutinesApi::class)
   override fun <A : WorkflowAction<StateT, OutputT>> makeActionSink(): Sink<A> {
     checkNotFrozen()
 
     return object : Sink<A> {
       override fun send(value: A) {
         // If this returns false, we lost the race with another event being sent.
         check(nextUpdateFromEvent.complete(value)) {
-          "Expected to successfully deliver $value. Are you using an old rendering?"
+          "Expected to successfully deliver action. Are you using an old rendering?\n" +
```
rjrjr: Is this any clearer?

```
"A rendering can only accept a single action.\n" +
    "\tprocessed action=${...}\n"
    "\tlate action=$value" +
```

zach-klippenstein: I like the processed/late terminology more.

PR #674, opened 2019-10-15
Adds dynamically configured master / detail behavior to To Do sample

### LINKED ISSUE #61: Demonstrate Master / Detail navigation
AC

 * Master detail navigation via split view container
 * Animated transitions on both left and right
 * Don't worry about config changes yet, that's covered by #29

### LINKED ISSUE #660: Make Todo sample master/detail on tablets
Requested: https://kotlinlang.slack.com/archives/C5HT9AL7Q/p1570716327013500?thread_ts=1570487716.042800&cid=C5HT9AL7Q

rjrjr:
Introduces "Aware" interface pattern for renderings.

 * `BackStackAware` allows a rendering to be customized based on
   its presence in a `BackStackScreen`, and whether it's the
   first frame or not.

 * `MasterDetailAware` allows a rendering to be customized based
   on its presence in a `MasterDetailScreen`, and whether that
   parent is single pane or split view.

Notice how a workflow emitting `MasterDetailScreen` renderings never knows
whether it's being shown in two panes or one -- those concerns are entirely
managed by view code, and may change at any time. E.g., on phones this sample
swaps between single pane in portrait and master detail in landscape.

`MasterDetailContainer` chooses the split pane or single pane behavior based
on the views made available in its layout file, and repackages its children
into a `BackStackScreen` in single pane mode. The split pane layouts are also
RTL / LTR aware, all through conventional Android resource loading.

More notes:

 * With the introduction of `View.backPressedHandler`, the `onGoBack`
   callback in `BackStackScreen` had become vestigial -- it was never hooked up,
   and was clearly unnecessary. It's been deleted.

 * The choice of push or pop effect in `BackStackContainer` is no longer based
   on the fragile `ViewStateCache`. Instead we choose by comparing the previous
   rendering with the new one. This ensures the right choice across configuration
   changes.

 * Constructors of the container views are tweaked to make them friendly to
   the Android Studio layout editor.

Closes #660, closes #61.

md

### DIFF
--- a/kotlin/samples/todo-android/app/src/main/java/com/squareup/sample/mainactivity/MasterDetailContainer.kt
+++ b/kotlin/samples/todo-android/app/src/main/java/com/squareup/sample/mainactivity/MasterDetailContainer.kt
@@ -1,0 +1,91 @@
+/*
+ * Copyright 2019 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.sample.mainactivity
+
+import android.view.View
+import com.squareup.sample.todo.MasterDetailAware.Companion.makeAware
+import com.squareup.sample.todo.MasterDetailConfig.Detail
+import com.squareup.sample.todo.MasterDetailConfig.Master
+import com.squareup.sample.todo.MasterDetailConfig.Only
+import com.squareup.sample.todo.MasterDetailScreen
+import com.squareup.sample.todo.R
+import com.squareup.workflow.ui.BackStackScreen
+import com.squareup.workflow.ui.LayoutRunner
+import com.squareup.workflow.ui.ViewBinding
+import com.squareup.workflow.ui.ViewRegistry
+import com.squareup.workflow.ui.WorkflowViewStub
+
+/**
+ * Displays [MasterDetailScreen] renderings in either split pane or single pane
+ * treatment, depending on the setup of the given [View]. The view must provide
+ * either a single [WorkflowViewStub] with id [R.id.master_detail_single_stub],
+ * or else two with ids [R.id.master_stub] and [R.id.detail_stub].
+ *
+ * For single pane layouts, [MasterDetailScreen] is repackaged as a [BackStackScreen]
+ * with [MasterDetailScreen.masterRendering] as the base of the stack.
+ */
+class MasterDetailContainer(
+  view: View,
+  private val registry: ViewRegistry
+) : LayoutRunner<MasterDetailScreen> {
+
+  private val masterStub: WorkflowViewStub? = view.findViewById(R.id.master_stub)
+  private val detailStub: WorkflowViewStub? = view.findViewById(R.id.detail_stub)
+  private val singleStub: WorkflowViewStub? = view.findViewById(R.id.master_detail_single_stub)
+
+  init {
+    check((singleStub == null) xor (masterStub == null && detailStub == null)) {
+      "Layout must define only R.id.master_detail_single_stub, " +
+          "or else both R.id.master_stub and R.id.detail_stub"
+    }
+  }
+
+  override fun showRendering(rendering: MasterDetailScreen) {
+    if (singleStub == null) renderSplitView(rendering)
+    else renderSingleView(rendering, singleStub)
+  }
+
+  private fun renderSplitView(rendering: MasterDetailScreen) {
+    if (rendering.detailRendering == null && rendering.selectDefault != null) {
+      rendering.selectDefault!!.invoke()
+    } else {
+      val aware = rendering.copy(
+          masterRendering = makeAware(rendering.masterRendering, Master),
+          detailRendering = makeAware(rendering.detailRendering, Detail)
+      )
+
+      masterStub!!.update(aware.masterRendering, registry)
+      aware.detailRendering?.let { detailStub!!.update(it, registry) }
+    }
+  }
+
+  private fun renderSingleView(
+    rendering: MasterDetailScreen,
+    stub: WorkflowViewStub
+  ) {
+    val asBackStack: BackStackScreen<Any> = rendering.detailRendering
+        ?.let { BackStackScreen(stack = listOf(rendering.masterRendering, makeAware(it, Only))) }
+        ?: run { BackStackScreen(only = makeAware(rendering.masterRendering, Only)) }
+
+    stub.update(asBackStack, registry)
+  }
+
+  companion object : ViewBinding<MasterDetailScreen> by LayoutRunner.Binding(
+      type = MasterDetailScreen::class,
+      layoutId = R.layout.master_detail,
+      runnerConstructor = ::MasterDetailContainer
+  )
+}
--- a/kotlin/samples/todo-android/app/src/main/java/com/squareup/sample/mainactivity/TodoListsLayoutRunner.kt
+++ b/kotlin/samples/todo-android/app/src/main/java/com/squareup/sample/mainactivity/TodoListsLayoutRunner.kt
@@ -19,20 +19,26 @@ import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import android.widget.TextView
+import com.squareup.sample.todo.MasterDetailConfig.Master
 import com.squareup.sample.todo.R
 import com.squareup.sample.todo.TodoList
-import com.squareup.sample.todo.TodoListsRendering
+import com.squareup.sample.todo.TodoListsScreen
 import com.squareup.workflow.ui.LayoutRunner
 import com.squareup.workflow.ui.LayoutRunner.Companion.bind
 import com.squareup.workflow.ui.ViewBinding
 
-internal class TodoListsLayoutRunner(view: View) : LayoutRunner<TodoListsRendering> {
+internal class TodoListsLayoutRunner(view: View) : LayoutRunner<TodoListsScreen> {
   private val inflater = LayoutInflater.from(view.context)
   private val listsContainer = view.findViewById<ViewGroup>(R.id.todo_lists_container)
 
-  override fun showRendering(rendering: TodoListsRendering) {
+  override fun showRendering(rendering: TodoListsScreen) {
     for ((index, list) in rendering.lists.withIndex()) {
-      addRow(index, list) { rendering.onRowClicked(index) }
+      addRow(
+          index,
+          list,
+          selectable = rendering.masterDetailConfig == Master,
+          selected = index == rendering.selection && rendering.masterDetailConfig == Master
+      ) { rendering.onRowClicked(index) }
     }
     pruneDeadRowsFrom(rendering.lists.size)
   }
@@ -40,15 +46,22 @@ internal class TodoListsLayoutRunner(view: View) : LayoutRunner<TodoListsRenderi
   private fun addRow(
     index: Int,
     list: TodoList,
+    selectable: Boolean,
+    selected: Boolean,
     onClick: () -> Unit
   ) {
     val row: TextView = if (index < listsContainer.childCount) {
       listsContainer.getChildAt(index)
     } else {
-      inflater.inflate(R.layout.todo_lists_row_layout, listsContainer, false)
+      val layout = when {
+        selectable -> R.layout.todo_lists_selectable_row_layout
+        else -> R.layout.todo_lists_unselectable_row_layout
+      }
+      inflater.inflate(layout, listsContainer, false)
           .also { listsContainer.addView(it) }
     } as TextView
 
+    row.isActivated = selected
     row.text = list.title
     row.setOnClickListener { onClick() }
   }
@@ -57,7 +70,7 @@ internal class TodoListsLayoutRunner(view: View) : LayoutRunner<TodoListsRenderi
     while (listsContainer.childCount > index) listsContainer.removeViewAt(index)
   }
 
-  companion object : ViewBinding<TodoListsRendering> by bind(
+  companion object : ViewBinding<TodoListsScreen> by bind(
       R.layout.todo_lists_layout, ::TodoListsLayoutRunner
   )
 }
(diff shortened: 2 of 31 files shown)


rjrjr:
After sleeping on this I still like it, a lot. Punching up the documentation now.

I’m also thinking about adding coalescing to BackStackScreen. That is, if you add a backStackScreen child to the stack of a BackStackScreen parent, it gets inlined. Will try that in a separate PR.

Interesting thing to note: with the introduction of `View.backPressedHandler` the other week, the need to carry event handlers in `BackStackScreen` and now `MasterDetailScreen` has evaporated — turns out the ones in `BackStackScreen` had become no-ops, and I never even noticed.

bencochran:
I’ve been toying with the idea of making BackStackScreen always a tree on iOS (essentially, always coalesced). On my phone right now, but will type up my thoughts when I have a chance at a computer.

rjrjr:
Okay, ready for review.

kotlin/samples/todo-android/app/src/main/java/com/squareup/sample/mainactivity/MasterDetailContainer.kt
```diff
@@ -0,0 +1,67 @@
+package com.squareup.sample.mainactivity
```
rjrjr: Copyright missing throughout.

kotlin/samples/todo-android/app/src/main/java/com/squareup/sample/mainactivity/MasterDetailContainer.kt
```diff
@@ -0,0 +1,67 @@
...
+import com.squareup.sample.todo.MasterDetailConfig.Only
+import com.squareup.sample.todo.MasterDetailScreen
+import com.squareup.sample.todo.R
+import com.squareup.workflow.ui.BackStackScreen
+import com.squareup.workflow.ui.LayoutRunner
+import com.squareup.workflow.ui.ViewBinding
+import com.squareup.workflow.ui.ViewRegistry
+import com.squareup.workflow.ui.WorkflowViewStub
+
+class MasterDetailContainer(
```
rjrjr: doc

kotlin/samples/todo-android/app/src/main/java/com/squareup/sample/mainactivity/TodoListsLayoutRunner.kt
```diff
@@ -19,27 +19,33 @@ import android.view.LayoutInflater
...
   private val listsContainer = view.findViewById<ViewGroup>(R.id.todo_lists_container)
 
-  override fun showRendering(rendering: TodoListsRendering) {
+  override fun showRendering(rendering: TodoListsScreen) {
     for ((index, list) in rendering.lists.withIndex()) {
-      addRow(index, list) { rendering.onRowClicked(index) }
+      addRow(
+          index,
+          list,
+          selected = index == rendering.selection && rendering.masterDetailConfig == Master
```
rjrjr: Oh crud, I forgot to fix actual display of the selection. It's being set just fine, but nothing visible happens.

kotlin/samples/todo-android/app/src/main/res/layout-ldrtl/master_detail_split.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
rjrjr: Add a note to the commit message about the LTR / RTL awareness of the layouts.

kotlin/samples/todo-android/app/src/main/res/layout/master_detail_single.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
rjrjr: Out of scope, but something to think about -- can we find a general way to get rid of this `FrameLayout` and just use the `WorkflowViewStub` directly?

kotlin/samples/todo-android/common/src/main/java/com/squareup/sample/todo/MasterDetailAware.kt
```diff
@@ -0,0 +1,62 @@
...
+  Master,
+
+  /**
+   * Rendering is in [MasterDetailScreen.detailRendering], running in split screen
+   * configuration.
+   */
+  Detail,
+
+  /**
+   * Rendering is the only visible child in a [MasterDetailScreen] running in
```
rjrjr: Seems odd that single pane children can't know if they're master or detail. OTOH I haven't actually needed that info.

kotlin/samples/todo-android/common/src/main/java/com/squareup/sample/todo/MasterDetailAware.kt
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Implemented by rendering types to allow them to customize appearance / behavior
+ * based on a [MasterDetailScreen] parent's [configuration][MasterDetailConfig].
+ */
+interface MasterDetailAware<T : MasterDetailAware<T>> {
+  /**
+   * The configuration of a possible [MasterDetailScreen] parent rendering.
+   * Should default to [MasterDetailConfig.None].
+   */
+  val masterDetailConfig: MasterDetailConfig
```
rjrjr: Thought about making this `MasterDetailConfig?` and getting rid of the `None` value (and same in back stack), but didn't seem like it would make anything more readable.

kotlin/samples/todo-android/common/src/main/java/com/squareup/sample/todo/TodoListsAppWorkflow.kt
```diff
@@ -76,18 +75,26 @@ class TodoListsAppWorkflow :
...
+      is ShowingLists -> MasterDetailScreen(
+          masterRendering = listOfLists,
+          selectDefault = { sink.send(onListSelected(0)) }
+      )
       is EditingList -> context.renderChild(
           editorWorkflow, state.lists[state.editingIndex], handler = this::onEditOutput
       ).let { editScreen ->
-        BackStackScreen(listOfLists, editScreen)
+        MasterDetailScreen(
+            listOfLists.copy(selection = state.editingIndex),
```
rjrjr: add field names

kotlin/samples/todo-android/common/src/main/java/com/squareup/sample/todo/TodoListsWorkflow.kt
```diff
@@ -20,19 +20,14 @@ import com.squareup.workflow.Sink
 import com.squareup.workflow.StatelessWorkflow
 import com.squareup.workflow.makeEventSink
 
-data class TodoListsRendering(
-  val lists: List<TodoList>,
-  val onRowClicked: (Int) -> Unit
-)
-
-class TodoListsWorkflow : StatelessWorkflow<List<TodoList>, Int, TodoListsRendering>() {
+class TodoListsWorkflow : StatelessWorkflow<List<TodoList>, Int, TodoListsScreen>() {
```
rjrjr: Note that this workflow is able to remain unaware of the current selection. The parent workflow adds that information to this child's rendering while assembling the composite rendering.

rjrjr: Probably worth adding a comment to that effect in the sample source code.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ModalContainer.kt
```diff
@@ -46,11 +46,13 @@ import com.squareup.workflow.ui.ModalContainer.Companion.forContainerScreen
  *
  * @param ModalRenderingT the type of the nested renderings to be shown in a dialog window.
  */
-abstract class ModalContainer<ModalRenderingT : Any>
-@JvmOverloads constructor(
+abstract class ModalContainer<ModalRenderingT : Any> @JvmOverloads constructor(
```
rjrjr: Update commit message to note these changes to make the container views friendly to the layout editor.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/backstack/BackStackContainer.kt
```diff
@@ -72,9 +83,12 @@ open class BackStackContainer(
         }
 
     val newView = registry.buildView(named.top, this)
-    val popped = viewStateCache.update(named.backStack, oldViewMaybe, newView)
+    viewStateCache.update(named.backStack, oldViewMaybe, newView)
+
+    val popped = currentRendering?.backStack?.any { compatible(it, named.top) } == true
```
rjrjr: Note in commit message that `BackStackContainer` no longer relies on its `ViewStateCache` to determine whether to run a push or pop effect. That choice is now completely deterministic, chosen by comparing the previous rendering with the new one.

rjrjr:
Update adds missing docs and copyright notices, and punches up the commit message.

bencochran:
Put my BackStackScreen thoughts into an issue to keep that conversation from consuming your PR: #676.

kotlin/samples/todo-android/app/src/main/java/com/squareup/sample/mainactivity/TodoListsLayoutRunner.kt
rjrjr: @zach-klippenstein Changed `selected` to `activated` which seems more correct, but I'm still clueless how to get the style system to acknowledge it. Do you know what the right thing is to do before I just hack something?

zach-klippenstein: I think you need to define something in your style to say how to render things in those states?  E.g. a state color drawable for the background of your row widget.

kotlin/samples/todo-android/common/src/main/java/com/squareup/sample/todo/MasterDetailAware.kt
```diff
@@ -0,0 +1,77 @@
...
+package com.squareup.sample.todo
+
+/**
+ * Informs [MasterDetailAware] renderings whether they're children of a [MasterDetailScreen],
+ * and if so in what configuration.
+ */
+enum class MasterDetailConfig {
+  /**
+   * Parent is not a [MasterDetailScreen]. Expected to be used as the default value
+   * of [MasterDetailAware.masterDetailConfig].
```
zach-klippenstein: I don't know what this value actually implies if this is the config that is being rendered. Is this a valid configuration to render as? If not, I like the nullable route better because it lets the type system enforce whether this default value is allowed (for example, shouldn't be allowed as an argument to `makeAware`).

kotlin/samples/todo-android/common/src/main/java/com/squareup/sample/todo/MasterDetailAware.kt
zach-klippenstein: What happens if you default to another value?

kotlin/samples/todo-android/common/src/main/java/com/squareup/sample/todo/MasterDetailAware.kt
```diff
@@ -0,0 +1,77 @@
...
+   * Creates a copy of the receiver with the given [config].
+   */
+  fun withMasterDetailConfig(config: MasterDetailConfig): T
+
+  companion object {
+    /**
+     * Convenience allowing [MasterDetailScreen] containers to call
+     * [withMasterDetailConfig] on conforming children before displaying them.
+     */
+    inline fun <reified T> makeAware(
```
zach-klippenstein: Shouldn't need to be inline/reified, you're just casting.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/backstack/BackStackContainer.kt
```diff
@@ -36,31 +39,39 @@ import com.squareup.workflow.ui.ViewBinding
...
+  private var currentRendering: BackStackScreen<Named<*>>? = null
+
   private fun update(newRendering: BackStackScreen<*>) {
-    // ViewStateCache requires that everything be Named, for ease of comparison and
-    // serialization (that Named.key string is very handy). It's fine if client code is
-    // already using Named for its own purposes, recursion works.
-    val named: BackStackScreen<Named<*>> =
-      BackStackScreen(newRendering.stack.map { Named(it, "backstack") }, newRendering.onGoBack)
+    val named: BackStackScreen<Named<*>> = newRendering.stack.asSequence()
+        // Let interested children know where that they're in a stack.
```
zach-klippenstein: Extra word?

kotlin/samples/todo-android/common/src/main/java/com/squareup/sample/todo/MasterDetailAware.kt
rjrjr: Then you're lying to yourself. If you get rendered in a container other than `MasterDetailScreen` there is no promise that the value will be updated.

kotlin/samples/todo-android/common/src/main/java/com/squareup/sample/todo/MasterDetailAware.kt
rjrjr: It is valid to be rendered with `None`, that's what happens if your parent isn't `MasterDetailScreen`. But you're right, `makeAware` should never be called with `None`. That sounds compelling to me.

rjrjr:
Update eliminates `None` values and fixes selection, selectors.

rjrjr:
@zach-klippenstein

Another alternative I considered was to pass the container hints via view tags (e.g. BackStackContainer could put an `R.id.back_stack_config` tag on each child view it creates, and define `val View.backStackConfig : BackStackConfig`), so all that stuff would truly live only in view land. 

Passing it through the renderings seemed less magical so I went that way, but you could also argue that it’s muddying the boundaries. Starting to think I made the wrong call. WDYT?

rjrjr:
As I think through the details I'm going pretty cold on using view tags. Too complex, too brittle, too racy. The recursive search notion is particularly fraught: it would require child views to be added to their parents before their `showRendering()` method is called, which will be easy for custom containers to break.

We could still introduce our own view-side-only ambients system (useful term!), and probably will have to sooner or later, but I'm not confident we've experienced enough use cases yet to do that justice. 

So far I'm inclined to stick with this, and would be willing to ship 1.0 this way. Even if it's a bit leaky, it gets the job done and requires almost new machinery in the library code (except for that change to BackStackContainer, sigh). I think we should live without ambients until we have a clearer idea of what we want from them. Much better to err on the side of minimalism than complexity.

PR #677, opened 2019-10-16
Remove robolectric dependency definition, we're never actually using it.

zach-klippenstein:
Makes #675 obsolete (closes #675).


zach-klippenstein:
@rjrjr was the one who pointed this out in https://github.com/square/workflow/pull/675#issuecomment-542422452, so I'm going to assume he approves.

rjrjr:
LGTM

PR #678, opened 2019-10-16
Make testing infra run double render passes to suss out side effect code in render methods.

### LINKED ISSUE #501: Make workflow test helpers check render idempotency
The contract of `render` methods is that they should be idempotent – calling them multiple times with the same values should be equivalent to calling them once. The testing infrastructure can help validate this by always invoking render twice to help catch invalid implementations that try doing side effects.

cc @rjrjr @grantamos

zach-klippenstein:
Closes #501.

### DIFF
--- a/kotlin/workflow-runtime/src/main/java/com/squareup/workflow/testing/LaunchWorkflow.kt
+++ b/kotlin/workflow-runtime/src/main/java/com/squareup/workflow/testing/LaunchWorkflow.kt
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2019 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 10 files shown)


kotlin/workflow-runtime/src/main/java/com/squareup/workflow/testing/LaunchWorkflow.kt
```diff
@@ -0,0 +1,59 @@
...
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.flow.Flow
+import org.jetbrains.annotations.TestOnly
+
+/**
+ * Launches the [workflow] in a new coroutine in [scope]. The workflow tree is seeded with
+ * [WorkflowTestParams.StartMode] and the first value emitted by [props].  Subsequent values
+ * emitted from [props] will be used to re-render the workflow.
+ *
+ * See [launchWorkflowIn] for documentation about most of the parameters and behavior.
```
rjrjr: I don't see where you've documented how to opt in / out of double rendering.

Also wondering if this method and `WorkflowTestParams` really need to be public.

zach-klippenstein: They do because this method is called from the `workflow-testing` API, and there's no other way to make that possible, and `WorkflowTestParams` is a first-class API currency – all the test functions take one.

zach-klippenstein: I will add more documentation about opting out.

zach-klippenstein: It's actually documented on `WorkflowTestParams`.

PR #679, opened 2019-10-16
Eliminates WorkflowRunner.onSaveInstanceState.

### LINKED ISSUE #431: Use AndroidX ViewModel serialization instead of activity bundle

rjrjr:
We now use `androidx.savedstate` for persistence purposes, so the
`WorkflowRunner` Android integration API is as simple as it  will ever be.

Closes #431.

### DIFF
--- a/kotlin/workflow-ui-android/src/test/java/com/squareup/workflow/ui/WorkflowRunnerViewModelTest.kt
+++ b/kotlin/workflow-ui-android/src/test/java/com/squareup/workflow/ui/WorkflowRunnerViewModelTest.kt
@@ -1,5 +1,10 @@
 package com.squareup.workflow.ui
 
+import androidx.lifecycle.Lifecycle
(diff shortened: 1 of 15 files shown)


kotlin/workflow-ui-android/src/test/java/com/squareup/workflow/ui/WorkflowRunnerViewModelTest.kt
```diff
@@ -132,9 +137,28 @@ class WorkflowRunnerViewModelTest {
   }
 
   private fun <O : Any, R : Any> Workflow<Unit, O, R>.run(): WorkflowRunnerViewModel<O> {
+    val registryOwner = object : SavedStateRegistryOwner {
```
rjrjr: Thanks for the ease of testing, Google. 🙄

rjrjr:
> Do we need any documentation updates to go along with this?

I don't see any.

PR #681, opened 2019-10-16
Make it possible to pass just the snapshot for the root workflow into the test methods.

### LINKED ISSUE #538: testFromStart with SnapShot fails tests
I’m upgrading our OnlineBookingWorkflow to v2. In our tests we’re using 
```
workflow.testFromStart(snapshot)

{ ... }
```
It looks like we’re the first people to be using this method with a snapshot and we’re having some issues. We’re getting EOFExceptions (end of file) when we try to read from our snapshot.

It looks like when we call `testFromStart` we create a `WorkflowNode`, which has a `state` as an instance variable. If the snapshot is supplied then it will try to restore the state during initialization. When it restores the state it appears that it expects the buffer to be formatted as so
`[length of parent snapshot][parent snapshot][length of child snapshot][child snapshot]`
But the snapshot that we pass in is only the `[parent snapshot]`. So when we try to collect the [parent snapshot] to parse into a state it crashes since the first 4 bytes in the buffer are `com.`, which is a huge int and tries to read too much. lmk if that makes sense.

The consumers of the method probably shouldn’t be adding those bytes. It seems like `testFromStart` should prepend the length of the snapshot.

zach-klippenstein:
This PR gives you the option of starting a workflow test session with either just the snapshot for the root workflow itself, or the complete snapshot of the entire tree from a previous workflow run.

Fixes #538.

### DIFF
--- a/kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/TreeSnapshots.kt
+++ b/kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/TreeSnapshots.kt
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2019 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.workflow.internal
+
+import com.squareup.workflow.Snapshot
+import com.squareup.workflow.parse
+import com.squareup.workflow.readByteStringWithLength
+import com.squareup.workflow.readList
+import com.squareup.workflow.writeByteStringWithLength
+import com.squareup.workflow.writeList
+import okio.ByteString
+
+/**
+ * Returns a [Snapshot] that lazily aggregates a root snapshot and the snapshots of all its children
+ * into a single [ByteString].
+ *
+ * The component snapshots can be restored with [parseTreeSnapshot].
+ *
+ * @see parseTreeSnapshot
+ */
+internal fun createTreeSnapshot(
+  rootSnapshot: Snapshot,
+  childSnapshots: List<Pair<AnyId, Snapshot>>
+): Snapshot = Snapshot.write { sink ->
+  sink.writeByteStringWithLength(rootSnapshot.bytes)
+  sink.writeList(childSnapshots) { (childId, childSnapshot) ->
+    writeByteStringWithLength(childId.toByteString())
+    writeByteStringWithLength(childSnapshot.bytes)
+  }
+}
+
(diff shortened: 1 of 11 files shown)


kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/TreeSnapshots.kt
```diff
@@ -0,0 +1,65 @@
...
+    writeByteStringWithLength(childId.toByteString())
+    writeByteStringWithLength(childSnapshot.bytes)
+  }
+}
+
+/**
+ * Parses a "root" snapshot and the list of child snapshots with associated [WorkflowId]s from a
+ * [ByteString] returned by [createTreeSnapshot].
+ *
+ * Never returns an empty root snapshot: if the root snapshot is empty it will be null.
```
rjrjr: What about child snapshots? Might they be empty?

rjrjr: Looks like no. Why the difference?

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/TreeSnapshots.kt
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * @see createTreeSnapshot
+ */
+internal fun parseTreeSnapshot(
+  treeSnapshot: ByteString
```
rjrjr: Nit, your call: be a bit less confusing in here if the `ByteString`s had some other name. e.g. `rootBytes` or `rootSnapshotBytes`.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/testing/WorkflowTestParams.kt
```diff
@@ -59,9 +58,30 @@ data class WorkflowTestParams<out StateT>(
...
-     * snapshot.
+     * snapshot.  Only applies to [StatefulWorkflow][com.squareup.workflow.StatelessWorkflow]s.
+     *
+     * This differs from [StartFromCompleteSnapshot] because it represents only the snapshot for
+     * the root workflow, without any of the snapshots of its children or other bookkeeping data
+     * added by the workflow runtime.
+     *
+     * @param snapshot A [Snapshot] that can be directly parsed by a workflow's `initialState`
+     * method. For workflow trees, this is only the snapshot of the _root_ workflow. Testing with
+     * a complete snapshot of a workflow tree is not currently supported.
```
rjrjr: Maybe say explicitly that this is the snapshot returned by the root's `StatefulWorkflow.snapshotState`, for easier contrast with the `WorkflowTester.awaitNextSnapshot` bit below.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/TreeSnapshots.kt
zach-klippenstein: Because the current node doesn't actually care if its children have snapshots or not.

zach-klippenstein: Added some documentation about this.

zach-klippenstein:
Pardon the noise there, merge-when-green bot did a bunch of weird stuff that I had to clean up. This should be good to merge on green now.

PR #682, opened 2019-10-17
Add configuration for merge-when-green app.

zach-klippenstein:


.github/merge-when-green.yml
```diff
@@ -0,0 +1,13 @@
+# See https://github.com/phstc/probot-merge-when-green for documentation of configuration options.
+
+# requiredChecks:
+#   - circleci
+#   - travis-ci
+# requiredStatuses:
+#   - jenkins
+
+requireApprovalFromRequestedReviewers: true
```
rjrjr: Are you sure this will do what you want w/o travis-ci configured? Looks like it might just merge as soon as we approve.

rjrjr: Oh, I see: 

> By default, Merge when green will wait for branch protection to pass and also require circleci and travis-ci before attempting to merge.

Now I'm wondering if it will hang out forever waiting for the `circleci` checks we don't run. Let's find out!

zach-klippenstein: I would have assumed it just waits for all checks to pass, but if you have to opt-in… seems to be doing the right thing:

image

Issue #635, opened 2019-10-02
Throwing exception on render throws it multiple times when using awaitFailure() in test


kirillzh:
Consider this simple stateful workflow that throws `TODO()` on render.
```kotlin
class FooWorkflow : StatefulWorkflow<Unit, State, Unit, Unit>() {

  override fun initialState(
    props: Unit,
    snapshot: Snapshot?
  ): State = State

  override fun render(
    props: Unit,
    state: State,
    context: RenderContext<State, Unit>
  ) {
    TODO("implement me")
  }

  override fun snapshotState(state: State): Snapshot = Snapshot.EMPTY

  object State
}
```

This is what my unit test looks like to test that the workflow indeed throws (realistically I wanted to throw in a very certain, not yet implemented case and I added a unit test for that):
```kotlin
class FooWorkflowTest {
  private val fooWorkflow = FooWorkflow()

  @Test
  fun verifyThrowsException() {
    fooWorkflow.testFromStart {
      assertThat(awaitFailure()).isInstanceOf<NotImplementedError>()
    }
  }
}
```

While test passes it throws right after that:
```
Exception in thread "main @coroutine#3" kotlin.NotImplementedError: An operation is not implemented: implement
	at com.squareup.scales.FooWorkflow.render(FooWorkflow.kt:20)
	at com.squareup.scales.FooWorkflow.render(FooWorkflow.kt:8)
	at com.squareup.workflow.internal.WorkflowNode.renderWithStateType(WorkflowNode.kt:178)
	at com.squareup.workflow.internal.WorkflowNode.render(WorkflowNode.kt:95)
	at com.squareup.workflow.internal.RealWorkflowLoop$runWorkflowLoop$2.invokeSuspend(WorkflowLoop.kt:79)
	(Coroutine boundary)
	at com.squareup.workflow.LaunchWorkflowKt$launchWorkflowImpl$workflowJob$1.invokeSuspend(LaunchWorkflow.kt:174)
Caused by: kotlin.NotImplementedError: An operation is not implemented: implement
	at com.squareup.scales.FooWorkflow.render(FooWorkflow.kt:20)
	at com.squareup.scales.FooWorkflow.render(FooWorkflow.kt:8)
	at com.squareup.workflow.internal.WorkflowNode.renderWithStateType(WorkflowNode.kt:178)
	at com.squareup.workflow.internal.WorkflowNode.render(WorkflowNode.kt:95)
[...]

PR #686, opened 2019-10-18
Throw workflow exceptions from test/testFromStart instead of leaking to uncaught exception handler.

zach-klippenstein:
Simplifies testing API by removing `awaitFailure()` – this method is now redundant since any
failures are thrown from the `test`/`testFromStart` function directly. This is simpler to use since
any exceptions from the workflow or runtime are now just thrown normally and will always cause
tests to fail, and it doesn't require any additional API to test expected failures.

Fixe #635.


kotlin/internal-testing-utils/src/main/java/com/squareup/workflow/internal/util/UncaughtExceptionGuard.kt
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * Helper for collecting uncaught exceptions and reporting them during the execution of some
+ * function.
+ *
+ * ## Usage
+ *
+ * 1. Register an uncaught exception handler to report uncaught exceptions with [reportUncaught].
+ * 2. Run your code with [runRethrowingUncaught]. Any exceptions passed to [reportUncaught] will be
+ * thrown after the block returns. If the block itself throws, that exception will be rethrown
+ * directly with uncaught exceptions marked as suppressed.
```
rjrjr: _marked as suppressed_ is a bit vague. Perhaps:

```
 * thrown after the block returns. If the block itself throws, that exception will be rethrown
 * directly with uncaught exceptions added to its [suppressed list][Throwable.addSuppressed].
```

kotlin/internal-testing-utils/src/main/java/com/squareup/workflow/internal/util/UncaughtExceptionGuard.kt
```diff
@@ -0,0 +1,87 @@
...
+    uncaughtException.get()
+        ?.let { throw it }
+
+    return result
+  }
+
+  /**
+   * Adds `e` as a suppressed exception to this one, unless `e` is null or the same object as this.
+   */
+  private fun Throwable.addSuppressedSafely(e: Throwable?) {
```
rjrjr: Doesn't look like you ever call this with null.

kotlin/internal-testing-utils/src/main/java/com/squareup/workflow/internal/util/UncaughtExceptionGuard.kt
```diff
@@ -0,0 +1,87 @@
...
+import java.util.concurrent.atomic.AtomicReference
+
+/**
+ * Helper for collecting uncaught exceptions and reporting them during the execution of some
+ * function.
+ *
+ * ## Usage
+ *
+ * 1. Register an uncaught exception handler to report uncaught exceptions with [reportUncaught].
+ * 2. Run your code with [runRethrowingUncaught]. Any exceptions passed to [reportUncaught] will be
```
rjrjr: No discussion of `rethrowingUncaughtExceptions()`, so I come away unclear when I should use that and when I should use these methods directly.

kotlin/internal-testing-utils/src/test/java/com/squareup/workflow/internal/util/UncaughtExceptionGuardTest.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 Square Inc.
+ * Copyright 2019 Square Inc.
```
rjrjr: Updating copyright dates is weird. There is still a lot of code that was written two years ago, a newer date is a weaker claim. (IANAL.)

kotlin/workflow-testing/build.gradle
```diff
@@ -29,7 +29,9 @@ dependencies {
   api project(':workflow-core')
   api project(':workflow-runtime')
   api deps.kotlin.coroutines.test
-  api deps.kotlin.stdLib.jdk6
+  api deps.kotlin.stdLib.jdk7
```
rjrjr: We don't use 8?

kotlin/internal-testing-utils/src/main/java/com/squareup/workflow/internal/util/UncaughtExceptionGuard.kt
zach-klippenstein: Added a note saying when to use it.

kotlin/workflow-testing/build.gradle
zach-klippenstein: Updated.

kotlin/internal-testing-utils/src/main/java/com/squareup/workflow/internal/util/UncaughtExceptionGuard.kt
zach-klippenstein: It is above, in `runRethrowingUncaught`. `uncaughtException.get()` can return null.

PR #687, opened 2019-10-18
Proposal: New render testing API for kotlin.

### LINKED ISSUE #489: Need better testing story for anonymous workers
It is very convenient to create anonymous workers inside a workflow. However, since they're not injected, they can't be tested with `testRender`. They _can_ be tested with `testFromState`, but only starting state can be specified – there's no way to get the transitioned-to state. 

Should we make `WorkflowTester` expose state transitions?

cc @rjrjr @grantamos

### LINKED ISSUE #499: testRender testing helper should allow overriding the rendering returned by renderChild
This is useful when the workflow-under-test is doing something internal to transform an injected workflow, but the test knows the type+key and just wants to force a specific render result.

cc @grantamos

zach-klippenstein:
Fixes #489 and fixes #499.

### DIFF
--- a/kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RealRenderTester.kt
+++ b/kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RealRenderTester.kt
@@ -1,0 +1,227 @@
+/*
+ * Copyright 2019 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.workflow.testing
+
+import com.squareup.workflow.RenderContext
+import com.squareup.workflow.Sink
+import com.squareup.workflow.StatefulWorkflow
+import com.squareup.workflow.Worker
+import com.squareup.workflow.Workflow
+import com.squareup.workflow.WorkflowAction
+import com.squareup.workflow.applyTo
+import com.squareup.workflow.testing.RealRenderTester.Expectation.ExpectedWorker
+import com.squareup.workflow.testing.RealRenderTester.Expectation.ExpectedWorkflow
+import kotlin.reflect.KClass
+
+internal class RealRenderTester<PropsT, StateT, OutputT : Any, RenderingT>(
+  private val workflow: StatefulWorkflow<PropsT, StateT, OutputT, RenderingT>,
+  private val props: PropsT,
+  private val state: StateT,
+  private val expectations: MutableList<Expectation<*>> = mutableListOf(),
+  private val consumedExpectations: MutableList<Expectation<*>> = mutableListOf(),
+  private var childWillEmitOutput: Boolean = false,
+  private var processedAction: WorkflowAction<StateT, OutputT>? = null
+) : RenderTester<PropsT, StateT, OutputT, RenderingT>,
+    RenderContext<StateT, OutputT>,
+    RenderTestResult<StateT, OutputT> {
+
+  internal sealed class Expectation<out OutputT> {
+    abstract val output: EmittedOutput<OutputT>?
+
+    data class ExpectedWorkflow<OutputT : Any, RenderingT>(
+      val workflowType: KClass<out Workflow<*, OutputT, RenderingT>>,
+      val key: String,
+      val assertProps: (props: Any?) -> Unit,
+      val rendering: RenderingT,
+      override val output: EmittedOutput<OutputT>?
+    ) : Expectation<OutputT>()
+
+    data class ExpectedWorker<out OutputT>(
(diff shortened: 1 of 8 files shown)


zach-klippenstein:
Usage:

```kotlin
 workflow.renderTester()
   // Test rendering from initial state…
   .render(
      props = MyProps(…),
      expectations = RenderExpectations(
        expectedWorkers = listOf(
          ExpectedWorker(
            matchesWhen = { it is SubmitLoginWorker },
            key = "signin",
            output = EmittedOutput(LoginResponse(success = true))
          )
        ),
        expectedWorkflows = listOf(
          ExpectedWorkflow(
            workflowType = ChildWorkflow::class,
            key = "child",
            assertProps = { assertThat(it.email).isEqualTo("test@foo.com") },
            rendering = ChildRendering("message")
          )
        )
        assertState = { assertTrue(it is FooState) },
        assertOutput = { assertTrue(it is LoginSuccessful) },
        assertRendering = { assertTrue(it is LoadingScreen) }
      )
   )
   // State has now transitioned, test rendering with the new state…
   .render(
     …
   )
```

kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RenderTester.kt
```diff
@@ -0,0 +1,107 @@
...
+  val key: String = "",
+  val output: EmittedOutput<OutputT>? = null
+)
+
+/**
+ * @param assertProps Assertions on the props passed to `renderChild`.
+ * @param rendering The rendering to return from `renderChild` when this workflow is rendered.
+ * @param output If non-null, [EmittedOutput.output] will be emitted when this worker is ran.
+ */
+data class ExpectedWorkflow<PropsT, OutputT : Any, RenderingT>(
```
kirillzh: Would it be possible to pass workflow type here? `workflowType` would become unnecessary.

zach-klippenstein: You can't only specify a subset of type parameters, so you'd have to also explicitly pass every other type. It would be gross.

zach-klippenstein:
Talked to @bencochran offline, he's not happy with this API in Swift either, so closing until we can chat and figure out a path forward.

zach-klippenstein:
This PR is now ready for review.

kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RenderTester.kt
```diff
@@ -0,0 +1,159 @@
...
+      props: PropsT,
+      initialState: StateT
+    ): RenderTester<PropsT, StateT, OutputT, RenderingT> =
+// @formatter:on
+  RealRenderTester(this, props, initialState)
+
+/**
+ * The props must be specified, the initial state may be specified, and then all child workflows
+ * and workers that are expected to run must be specified with [expectWorkflow] and [expectWorker]
+ * calls. Then call [render] and perform any assertions on the rendering. An event may also be sent
```
rjrjr: You're implying here that, in addition to being able to send an event to a rendering, I can also simulate a child worker or workflow emitting output. Seems like that should be said out loud. Bonus points for providing an example.

kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RenderTester.kt
```diff
@@ -0,0 +1,159 @@
...
+ *     output = EmittedOutput(LoginResponse(success = true))
+ *   )
+ *   .expectWorkflow(
+ *     workflowType = ChildWorkflow::class,
+ *     key = "child",
+ *     assertProps = { assertThat(it.email).isEqualTo("test@foo.com") },
+ *     rendering = ChildRendering("message")
+ *   )
+ *   .render { rendering ->
+ *     assertThat(rendering.text).isEqualTo("foo")
```
rjrjr: Perhaps here could add a couple of (commented out?) lines demonstrating sending an event or having a child emit.

kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RealRenderTester.kt
```diff
@@ -0,0 +1,227 @@
...
+      val (newState, output) = action.applyTo(state)
+      block(newState, output)
+    }
+  }
+
+  private inline fun <reified T : Expectation<*>> consumeExpectation(
+    predicate: (T) -> Boolean,
+    description: () -> String
+  ): T {
+    val matchedExpectations = expectations.filterIsInstance<T>()
```
rjrjr: Kind of sad you didn't find a way to name something `greatExpectations`.

kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RenderTester.kt
zach-klippenstein: A child does emit, on line 79.

kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RenderTester.kt
zach-klippenstein: Added a bunch more examples, and updated this paragraph to mention that `expect*` can be used to configure outputs.

kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RenderTester.kt
zach-klippenstein: Added another example demonstrating this.

zach-klippenstein:
Pushed additional documentation and examples.

Issue #676, opened 2019-10-15
Proposal: Tree-shaped BackStackScreens


bencochran:
Right now BackStackScreen is essentially:

```swift
struct BackStackScreen {
     var items: [Item]
}
```
(where `Item` is basically a screen)

This work well for flat stacks, but means that if I want to push the details of a sub-flow down into a child workflow, I have to manage the concatenation of those screens myself. It also means that protecting from an empty list of items happens as a runtime assertion rather than being modeled in the type itself.

Back in the original iOS architecture discussions, the following basic shape was proposed:

```swift
struct BackStackScreen {
     var first: Item
     var subsequent: [BackStackScreen]
}
```

(ignore those particular spellings probably?)

For one, this lets me guarantee at compile time that someone rendering a BackStackScreen has at least one screen for me. It also allows me to nest child flows without reshaping the stack myself.

When it comes time to install this stack into a view, the flattened stack would be a depth-first traversal of this tree:

```swift
extension BackStackScreen {
    var flattented: [Item] {
        return [item] + subsequent.flatMap({ $0.flattened })
    }
}
```

------

**Alternative:** We could also simplify the composing of today’s `BackStackScreen`s by adding a convenience `func appending(backStackScreen: BackStackScreen)` and doing the concatenation in there. But it leaves the fact that as far as my code is concerned, I need to deal with the fact that the overall list of items _could_ be empty. (Also, on iOS inside Square today this is tricky because we still have support for implicit BackStackScreens, oops).

rjrjr:
WRT implicit, @AquaGeek is in the process of ripping that out, I think we can ignore for this conversation.

That's pretty similar to today's Kotlin `BackStackScreen`, except that the top / rest are just projections from the underlying list, basically the following (elides some convenience constructors):

```
data class BackStackScreen<StackedT : Any>(
    val stack: List<StackedT>
) {
  init {
    require(stack.isNotEmpty()) { "Empty stacks are not allowed." }
  }

  val top: StackedT = stack.last()
  val backStack: List<StackedT> = stack.subList(0, stack.size - 1)
}
```

That seems simpler to me — I treat the thing as a list a lot more often than I give special treatment to `0`, and I'm still protected from empty stacks.

[...]

PR #688, opened 2019-10-19
Compile time assurance that BackStackScreen is not empty.

rjrjr:
Previously we only enforced this with a runtime check.
Inspired by #676.

### DIFF
--- a/kotlin/samples/tictactoe/common/src/main/java/com/squareup/sample/mainworkflow/MainWorkflow.kt
+++ b/kotlin/samples/tictactoe/common/src/main/java/com/squareup/sample/mainworkflow/MainWorkflow.kt
@@ -19,7 +19,6 @@ import com.squareup.sample.authworkflow.AuthResult
 import com.squareup.sample.authworkflow.AuthResult.Authorized
 import com.squareup.sample.authworkflow.AuthResult.Canceled
 import com.squareup.sample.authworkflow.AuthWorkflow
-import com.squareup.sample.authworkflow.LoginScreen
 import com.squareup.sample.gameworkflow.GamePlayScreen
 import com.squareup.sample.gameworkflow.RealRunGameWorkflow
 import com.squareup.sample.gameworkflow.RunGameScreen
@@ -32,6 +31,7 @@ import com.squareup.workflow.Snapshot
 import com.squareup.workflow.StatefulWorkflow
 import com.squareup.workflow.Workflow
 import com.squareup.workflow.WorkflowAction
+import com.squareup.workflow.WorkflowAction.Companion.noAction
 import com.squareup.workflow.renderChild
 import com.squareup.workflow.ui.AlertContainerScreen
 
@@ -71,6 +71,9 @@ class MainWorkflow(
       val authScreen = context.renderChild(authWorkflow) { handleAuthResult(it) }
       val emptyGameScreen = GamePlayScreen()
 
+      // IDE is wrong, removing them breaks the compile.
+      // Probably due to https://youtrack.jetbrains.com/issue/KT-32869
+      @Suppress("RemoveExplicitTypeArguments")
       AlertContainerScreen(authScreen.inPanelOver<Any, Any>(emptyGameScreen))
     }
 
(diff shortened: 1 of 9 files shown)


kotlin/workflow-ui-core/src/main/java/com/squareup/workflow/ui/BackStackScreen.kt
```diff
@@ -16,17 +16,59 @@
...
+ * screens to which we may return ([backStack]). By rendering the entire
+ * history we allow the UI to do things like maintain cached view state,
+ * implement drag-back gestures without waiting for the workflow, etc.
+ *
+ * Effectively a list that can never be empty.
  */
-data class BackStackScreen<StackedT : Any>(val stack: List<StackedT>) {
-  constructor(only: StackedT) : this(listOf(only))
-  constructor(vararg stack: StackedT) : this(stack.toList())
+class BackStackScreen<StackedT : Any>(
```
rjrjr: This is the interesting part

kotlin/samples/tictactoe/common/src/main/java/com/squareup/sample/mainworkflow/MainWorkflow.kt
```diff
@@ -71,6 +71,8 @@ class MainWorkflow(
       val authScreen = context.renderChild(authWorkflow) { handleAuthResult(it) }
       val emptyGameScreen = GamePlayScreen()
 
+      // IDE is wrong, removing them breaks the compile.
```
kirillzh: This looks like a bug in the new type inference in Kotlin plugin. Looks similar to [KT-32869](https://youtrack.jetbrains.com/issue/KT-32869).

kotlin/workflow-ui-core/src/main/java/com/squareup/workflow/ui/BackStackScreen.kt
```diff
@@ -16,17 +16,59 @@
...
+
+  /**
+   * The active screen.
+   */
+  val top: StackedT = frames.last()
+
+  /**
+   * Screens to which we may return.
+   */
+  val backStack: List<StackedT> = frames.subList(0, frames.size - 1)
```
kirillzh: You can also do `frames.dropLast(1)`.

kotlin/workflow-ui-core/src/main/java/com/squareup/workflow/ui/BackStackScreen.kt
```diff
@@ -16,17 +16,59 @@
...
+  override fun equals(other: Any?): Boolean {
+    return (other as? BackStackScreen<*>)?.frames == frames
+  }
+
+  override fun hashCode(): Int {
+    return frames.hashCode()
+  }
 }
+
+private fun <T> List<T>.rest() = subList(1, size)
```
kirillzh: Similarly you can drop first element and get the rest of the list: `drop(1)`.

kotlin/workflow-ui-core/src/main/java/com/squareup/workflow/ui/BackStackScreen.kt
rjrjr: Nice, thanks.

rjrjr: Oh, wait a minute: those drop methods make new lists rather than just views on the original. Wasteful.

kirillzh: Good point!

rjrjr:
PTAL. Added another constructor:

```
  constructor(
    backStack: List<StackedT>,
    top: StackedT
  ) : this(...)
```

I just found myself implementing that in a one-off, and I think it could be common.

rjrjr:
After offline conversations gonna merge this.

kotlin/workflow-ui-core/src/main/java/com/squareup/workflow/ui/BackStackScreen.kt
```diff
@@ -16,17 +16,80 @@
...
+    backStack: List<StackedT>,
+    top: StackedT
+  ) : this(
+      bottom = backStack.firstOrNull() ?: top,
+      rest = backStack.takeIf { it.isNotEmpty() }
+          ?.let { it.subList(1, it.size) + top }
+          ?: emptyList()
+  )
+
+  val frames: List<StackedT> = listOf(bottom) + rest
```
zach-klippenstein: These are doing a bunch of unnecessary list copies, might be better to use a mutable list to massage the lists.

kotlin/workflow-ui-core/src/main/java/com/squareup/workflow/ui/BackStackScreen.kt
zach-klippenstein: This would be a great place to use the persistent collections library if it were stable.

kotlin/workflow-ui-core/src/main/java/com/squareup/workflow/ui/BackStackScreen.kt
rjrjr: We talked, don't actually see any real improvements to be made.

PR #690, opened 2019-10-19
Add WorkflowUI to Package.swift

timdonnelly:
Now that Xcode 11 supports Swift Package Manager directly, tests can be executed from the command line with `xcodebuild`. This allows us to specify an SDK (including iOS) and simulator device, which unlocks testing of code with system (UIKit/SwiftUI/AppKit) dependencies directly from the Swift package – no CocoaPods required.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -169,19 +169,11 @@ matrix:
         - bundle exec pod install
 
     - language: swift
-      name: "Swift - SPM (Xcode 10.2)"
+      name: "Swift - SPM (Xcode 11.1)"
       os: osx
-      osx_image: xcode10.2
-      cache:
-        directories:
-          - .build
-      before_install:
-        - gem update --system
-        - gem install bundler
-        - swift --version
-        - swift build --verbose
-      script:
-        - swift test
+      osx_image: xcode11.1
+      xcode_scheme: Workflow
+      xcode_destination: platform=iOS Simulator,OS=13.1,name=iPhone 11
 
     - language: swift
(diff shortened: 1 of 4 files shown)


swift/Workflow/Sources/SubtreeManager.swift
```diff
@@ -338,7 +338,7 @@ extension WorkflowNode.SubtreeManager {
         }
 
         func handle(event: Output) {
-            if #available(iOS 10.0, OSX 10.12, *) {
```
timdonnelly: We don't support macOS, so I removed this extra availability check while auditing for the new iOS requirement.

.travis.yml
```diff
@@ -169,19 +169,11 @@ matrix:
...
-          - .build
-      before_install:
-        - gem update --system
-        - gem install bundler
-        - swift --version
-        - swift build --verbose
+      osx_image: xcode11.1
       script:
-        - swift test
+        - xcodebuild -scheme "Workflow-Package" test -destination "name=iPhone 11"
```
timdonnelly: SPM generates a scheme called "<PackageName>-Package" to disambiguate if the package name is the same as one of its products (as is the case here)

swift/Workflow/Sources/SubtreeManager.swift
bencochran: In theory we _could_ support macOS in `Workflow` itself. But probably best to keep that a separate concerted effort (with tests to prevent regressions as long as no one is using it that way).

zach-klippenstein:
Omg, this is so much nicer. Thanks @timdonnelly!

PR #691, opened 2019-10-21
Add ContainerView to enable SwiftUI integration

timdonnelly:
This change allows a workflow to be used directly within a SwiftUI hierarchy. Note that `ViewRegistry` is no longer in play as SwiftUI views are already simple value types.

There are two ways to use `ContainerView`:

You can provide a `content` closure to `init(workflow:onOutput:content:)` to map the workflow's rendering into a SwiftUI view hierarchy:

```swift
var body: some View {
    ContainerView(workflow: MyWorkflow(), onOutput: self.handleOutput) { rendering in
        VStack {

            Text("The value is \(rendering.value)")

            Button(action: rendering.onIncrement) {
                Text("+")
            }

            Button(action: rendering.onDecrement) {
                Text("-")
            }

        }
    }
}
```

Or, if the rendering type already conforms to `View`, you can omit the closure:

```swift
var body: some View {
    ContainerView(workflow: MyWorkflow(), onOutput: self.handleOutput)
}
```

### DIFF
--- a/swift/WorkflowUI/Sources/Container/ContainerView.swift
+++ b/swift/WorkflowUI/Sources/Container/ContainerView.swift
@@ -1,0 +1,202 @@
+#if canImport(SwiftUI) && canImport(Combine) && swift(>=5.1)
+
+import Workflow
+import ReactiveSwift
+import SwiftUI
+import Combine
+
+
+/// Hosts a Workflow-powered view hierarchy.
+///
+/// Example:
+///
+/// ```
+/// var body: some View {
+///     ContainerView(workflow: MyWorkflow(), onOutput: { self.handleOutput($0) }) { rendering in
+///         VStack {
+///
+///             Text("The value is \(rendering.value)")
+///
+///             Button(action: rendering.onIncrement) {
+///                 Text("+")
+///             }
+///
+///             Button(action: rendering.onDecrement) {
+///                 Text("-")
+///             }
+///
+///         }
+///     }
+/// }
+/// ```
+@available(iOS 13.0, *)
+public struct ContainerView<T: Workflow, Content: View>: View {
+    
+    // The workflow implementation to use
+    public var workflow: T
+    
+    // A handler for any output events emitted by the workflow
+    public var onOutput: (T.Output) -> Void
+    
+    // A closure that maps the workflow's rendering type into a view of type `Content`.
+    public var content: (T.Rendering) -> Content
+    
+    public init(workflow: T, onOutput: @escaping (T.Output) -> Void, content: @escaping (T.Rendering) -> Content) {
+        self.onOutput = onOutput
+        self.content = content
+        self.workflow = workflow
+    }
+    
+    public var body: some View {
+        IntermediateView(
+            workflow: workflow,
+            onOutput: onOutput,
+            content: content)
+    }
+    
+}
+
+@available(iOS 13.0, *)
+extension ContainerView where T.Output == Never {
+    
+    // Convenience initializer for workflows with no output.
+    public init(workflow: T, content: @escaping (T.Rendering) -> Content) {
+        self.init(workflow: workflow, onOutput: { _ in }, content: content)
+    }
+    
+}
+
+@available(iOS 13.0, *)
+extension ContainerView where T.Rendering == Content {
+    
+    // Convenience initializer for workflows whose rendering type conforms to `View`.
+    public init(workflow: T, onOutput: @escaping (T.Output) -> Void) {
+        self.init(workflow: workflow, onOutput: onOutput, content: { $0 })
+    }
+    
+}
+
+@available(iOS 13.0, *)
+extension ContainerView where T.Output == Never, T.Rendering == Content {
+    
+    // Convenience initializer for workflows with no output whose rendering type conforms to `View`.
+    public init(workflow: T) {
+        self.init(workflow: workflow, onOutput: { _ in }, content: { $0 })
+    }
+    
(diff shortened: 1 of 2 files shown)


timdonnelly:
cc @bencochran

kyleve:
Dope

zach-klippenstein:
Not reviewing code because I'm on phone and vacation, but `ContainerView` isn't really a view, right? Just a node that does some wiring?

timdonnelly:
> Not reviewing code because I'm on phone and vacation, but `ContainerView` isn't really a view, right? Just a node that does some wiring?

It has no visual content of its own... Is that what you mean by 'not a view'?

AquaGeek:
My only question is where we want to put this in the repo. Given we won't be using SwiftUI for a _long_ time, I don't necessarily want to take on the maintenance burden or signal that this is a fully supported part of the library. Maybe we put it in the samples? Do we split it into a separate repo?

timdonnelly:
@AquaGeek My hope is that this is a really fast way to get up and running with Workflow, either for new folks or just for prototyping. You can get a working project with UI up and running in < 30 seconds with SPM; and putting a simple UI around a workflow is incredibly fast.

If that is true, I see this as a valuable addition even if we know it won't ship in a Square app for some number of years.

swift/WorkflowUI/Sources/Container/ContainerView.swift
```diff
@@ -0,0 +1,203 @@
...
+///             Button(action: rendering.onDecrement) {
+///                 Text("-")
+///             }
+///
+///         }
+///     }
+/// }
+/// ```
+@available(iOS 13.0, *)
+public struct ContainerView<T: Workflow, Content: View>: View {
```
zach-klippenstein: This name seems pretty vague, can we call it `WorkflowContainerView` or something?

swift/WorkflowUI/Sources/Container/ContainerView.swift
```diff
@@ -0,0 +1,203 @@
...
+    
+    // The workflow implementation to use
```
zach-klippenstein: Shouldn't these be triple-slashes?

swift/WorkflowUI/Sources/Container/ContainerView.swift
```diff
@@ -0,0 +1,203 @@
...
+    public init(workflow: T, content: @escaping (T.Rendering) -> Content) {
+        self.init(workflow: workflow, onOutput: { _ in }, content: content)
+    }
+    
+}
+
+@available(iOS 13.0, *)
+extension ContainerView where T.Rendering == Content {
+    
+    // Convenience initializer for workflows whose rendering type conforms to `View`.
```
zach-klippenstein: Nice!

zach-klippenstein:
cc @JustinDSN

swift/WorkflowUI/Sources/Container/ContainerView.swift
timdonnelly: `WorkflowView`? I know we used the term `container` internally, but I agree that it isn't a very useful term here.

swift/WorkflowUI/Sources/Container/ContainerView.swift
timdonnelly: Yes, fixed.

swift/WorkflowUI/Sources/Container/WorkflowView.swift
```diff
@@ -0,0 +1,203 @@
...
+fileprivate final class WorkflowHostingViewController<T: Workflow, Content: View>: UIViewController {
+    
+    private let workflowHost: WorkflowHost<T>
+    private let hostingController: UIHostingController<RootView<Content>>
+    private let rootViewProvider: RootViewProvider<Content>
+    
+    var content: (T.Rendering) -> Content
+    var onOutput: (T.Output) -> Void
+    
+    private let (lifetime, token) = Lifetime.make()
```
zach-klippenstein: I don't see anywhere this lifetime is being disposed - am I just missing it, or is this a potential leak if this container is removed from the view hierarchy?

timdonnelly: Disposal happens when it is destroyed after its reference count reaches zero – which happens when its owner (the view controller) is removed from the hierarchy.

zach-klippenstein: Cool! Must be nice having deterministic memory management 😅

swift/WorkflowUI/Sources/Container/WorkflowView.swift
```diff
@@ -0,0 +1,203 @@
...
+    
+    private func didEmit(output: T.Output) {
+        onOutput(output)
+    }
+    
+    func update(to workflow: T) {
+        workflowHost.update(workflow: workflow)
+    }
+    
+}
```
bencochran: Should this plumb all the `childViewControllerFor…` methods through to `hostingController`? (Not exactly sure how status bar and whatnot interact with SwiftUI right now).

timdonnelly: They mostly... don't. But preferences go up the tree even if there are UIKit layers in the middle, so this should be ok.

zach-klippenstein:
> I could see this being in a separate module from WorkflowUI, probably not necessary to start.

We'll do this in kotlin because Compose is a separate dependency, but I don't think that's an issue here.

PR #694, opened 2019-10-24
DNR Raven sample app, richer master / detail sample

rjrjr:
This is very much work in progress


rjrjr:
I have to admit that adding `R.id.back_stack_body` id feels sketchy and weirdly specific to me. I'm inclined to think that I should either move it out to a subclass of `BackStackContainer` out in samples land; or perhaps make `BackStackContainer` itself a sample.

kotlin/samples/containers/poetry/src/main/java/com/squareup/sample/poetry/model/Poem.kt
```diff
@@ -0,0 +1,18 @@
+package com.squareup.sample.poetry.model
+
+class Poem(
+  val title: String,
+  val poet: Poet,
+  val stanzas: List<List<String>>
+) {
+  val initialStanzas = stanzas.map { lines -> lines[0].trim() }
```
zach-klippenstein: Nit
```suggestion
  val initialStanzas = stanzas.map { lines -> lines.first().trim() }
```

rjrjr:
@zach-klippenstein 

> I have to admit that adding `R.id.back_stack_body` feels sketchy and weirdly specific to me. I'm inclined to think that I should either move it to a subclass of `BackStackContainer` out in samples land; or perhaps make `BackStackContainer` itself a sample.

Thoughts?

rjrjr:
Updates:

 * Catch up with ContainerHints merge to master
 * Delete premature poetry app
 * Fix some theming errors
 * Delete some stale copies of resources that were moved to `samples/containers/android`
 * Add some missing copyright headers.

zach-klippenstein:
`back_stack_body` looks a bit odd, it's to support our desired actionbar animations right? It's zero-cost if you don't need it, you don't need to know about it, so i'm not too worried. Better to be overly-specific at first, we can generalize it later.

PR #695, opened 2019-10-25
Adds Poetry sample app, tweaks WorkflowStubView

rjrjr:
Demonstrates:

 * More complex master / detail behavior, with parallel backstacks.
 * Sharing workflows between apps.

Revealed weakness of `WorkflowStubView` WRT hiding. To make that practical,
`WorkflowStubView.actual` is never null. We expose the fact that `actual` is
`this` until a delegate arrives. Turns out that's just as convenient for
external callers as it is for the implementation.

poetry


kotlin/samples/containers/app-poetry/src/main/java/com/squareup/sample/poetryapp/PoemsBrowserWorkflow.kt
```diff
@@ -0,0 +1,47 @@
...
+import com.squareup.sample.poetry.model.Poem
+import com.squareup.workflow.RenderContext
+import com.squareup.workflow.Snapshot
+import com.squareup.workflow.StatefulWorkflow
+import com.squareup.workflow.WorkflowAction
+import com.squareup.workflow.parse
+import com.squareup.workflow.ui.BackStackScreen
+import com.squareup.workflow.workflowAction
+
+object PoemsBrowserWorkflow : StatefulWorkflow<List<Poem>, Int, Nothing, MasterDetailScreen>() {
```
zach-klippenstein: Nit: Typealias this state type.

kotlin/samples/containers/common/src/main/java/com/squareup/sample/container/masterdetail/MasterDetailScreen.kt
```diff
@@ -23,11 +23,62 @@ import com.squareup.workflow.ui.BackStackScreen
...
+   */
+  constructor(
+    masterRendering: BackStackScreen<Any>,
+    selectDefault: (() -> Unit)? = null
+  ) : this(masterRendering, null, selectDefault)
+
+  operator fun component1(): BackStackScreen<Any> = masterRendering
+  operator fun component2(): BackStackScreen<Any>? = detailRendering
+
+  operator fun plus(other: MasterDetailScreen): MasterDetailScreen {
```
zach-klippenstein: Needs kdoc.

kotlin/samples/containers/app-poetry/src/main/java/com/squareup/sample/poetryapp/PoemsBrowserWorkflow.kt
```diff
@@ -0,0 +1,47 @@
...
+    context: RenderContext<Int, Nothing>
+  ): MasterDetailScreen {
+    val poems = context.renderChild(PoemListWorkflow, props) { selected -> choosePoem(selected) }
+        .copy(selection = state)
+        .let { BackStackScreen<Any>(it) }
+
+    return if (state == -1) {
+      MasterDetailScreen(poems)
+    } else {
+      MasterDetailScreen(poems) + context.renderChild(PoemWorkflow, props[state]) { choosePoem(-1) }
```
zach-klippenstein: This `+` operator almost seems too magic to me. There's too much going on, since this type is a pair of lists. Does this operator concatenate to a new pair? Or is it a distributed concatenation of all the component lists? We can add a bunch of comments but that's already a smell. I'd rather have a method like `appendMasterDetail` or something, to make it clear that it's operating on both and what it's doing.

rjrjr:
Update catches up to #694 rebasing.

kotlin/samples/containers/common/src/main/java/com/squareup/sample/container/masterdetail/MasterDetailScreen.kt
rjrjr: 🐮 🔔

rjrjr:
Update adds missing copyright docs. Still need to address overly magic `plus` operator. _mañana_

kotlin/samples/containers/common/src/main/java/com/squareup/sample/container/masterdetail/MasterDetailScreen.kt
rjrjr: Needs unit testing too.

kotlin/samples/containers/app-poetry/src/main/java/com/squareup/sample/poetryapp/PoemsBrowserWorkflow.kt
kirillzh: Alternatively we can implement it as some infix function, though at this point you might as well just have a regular function.

zach-klippenstein: Actually I would also probably pull this render call out into a val so the screen composition is more clear.

kotlin/samples/containers/common/src/main/java/com/squareup/sample/container/masterdetail/MasterDetailScreen.kt
rjrjr:
```
  /**
   * Returns a new [MasterDetailScreen] appending the [masterRendering] and
   * [detailRendering] of [other] to those of the receiver. If the new screen's
   * [detailRendering] is `null`, it will have the [selectDefault] function of [other].
   */
  fun combine(other: MasterDetailScreen): MasterDetailScreen {
```

kotlin/samples/containers/app-poetry/src/main/java/com/squareup/sample/poetryapp/PoemsBrowserWorkflow.kt
zach-klippenstein: "combine" isn't really any clearer to me. But I don't have any better suggestions, might as well just use the `+` operator. Extracting the render call makes me a lot more comfortable in any case.

rjrjr: Yay! Restoring plus.

rjrjr:
```
    val poems: MasterDetailScreen =
      context.renderChild(PoemListWorkflow, props) { selected -> choosePoem(selected) }
          .copy(selection = state)
          .let { MasterDetailScreen(BackStackScreen(it)) }

    return if (state == -1) {
      poems
    } else {
      val poem: MasterDetailScreen =
        context.renderChild(PoemWorkflow, props[state]) { clearSelection }
      poems + poem
    }
```

kotlin/samples/containers/common/src/test/java/com/squareup/sample/container/masterdetail/MasterDetailScreenTest.kt
```diff
@@ -0,0 +1,180 @@
+package com.squareup.sample.container.masterdetail
```
rjrjr: Gah, missing frickin' copyright block.

rjrjr:
Update keeps up with master, fixes lint check (blank line).

PR #697, opened 2019-10-28
Drops BackStackScreen var-args constructor.

rjrjr:
Too easy to shoot yourself in the foot otherwise. I keep creating
`BackStackScreen<List<Foo>>` instead of `BackStackScreen<Foo>`. Yes, it's
annoying to have to call `listOf`, but it's not the end of the world.


rjrjr:
Argh, OTOH this brings back the empty list runtime error that we were trying to prevent. 

It's very hard to win here. Gonna close this until we have a chance to chat a bit.

PR #708, opened 2019-10-30
Publish internal testing artifact, required by other published artifacts.

zach-klippenstein:
Because the `internal-testing-utils` module wasn't configured to publish to maven, the modules that depend on it were using version `unspecified` in their POMs. This was invalid, but also the shared artifact just wasn't being published, so everything breaks consumers.


zach-klippenstein:
Manually verified that the generated POM includes the correct version now.

zach-klippenstein:
I'm happy with ½ checks passing, merging optimistically.

PR #709, opened 2019-10-30
Update changelog for v0.21.1.

zach-klippenstein:


zach-klippenstein:
This is a docs-only change, only waiting for one of the Travis builds to succeed.

PR #710, opened 2019-10-30
Add note in RELEASING about only using Xcode 10.

zach-klippenstein:
Sourcedocs doesn't seem to support Xcode 11, see https://github.com/eneko/SourceDocs/issues/30

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -141,6 +141,9 @@ sourcedocs generate -- -scheme WorkflowUI -workspace SampleApp.xcworkspace
 sourcedocs generate -- -scheme WorkflowTesting -workspace SampleApp.xcworkspace
 ```
 
+Note that currently sourcedocs only supports Xcode 10, if you run it with Xcode 11 you might see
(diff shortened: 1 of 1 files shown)


RELEASING.md
```diff
@@ -141,6 +141,9 @@ sourcedocs generate -- -scheme WorkflowUI -workspace SampleApp.xcworkspace
 sourcedocs generate -- -scheme WorkflowTesting -workspace SampleApp.xcworkspace
 ```
 
+Note that currently sourcedocs only supports Xcode 10, if you run it with Xcode 11 you might see
```
AquaGeek: Is this a `Sourcedocs` issue or is this a configuration issue on our side?

zach-klippenstein: Seems like a Sourcedocs issue. Filed a bug: https://github.com/eneko/SourceDocs/issues/30

zach-klippenstein:
Build failed due to cocoapods network timeout.

PR #721, opened 2019-11-05
Inline dependency version numbers so Dependabot can update them.

zach-klippenstein:
Discussion: https://workflow-community.slack.com/archives/CHTFPR277/p1572980514086900

### DIFF
--- a/kotlin/build.gradle
+++ b/kotlin/build.gradle
@@ -16,13 +16,6 @@
 buildscript {
   ext.versions = [
       'targetSdk': 29,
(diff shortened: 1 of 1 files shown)


zach-klippenstein:
After this merges, go to https://app.dependabot.com/accounts/square/projects/76682 and hit Bump Now to manually trigger dependabot.

zach-klippenstein:
Kotlin build went green, failure was only due to the Swift - SPM shard flaking (filed #722). Merging.

PR #742, opened 2019-11-12
Make event sinks queued and reusable instead of throw after going stale.

### LINKED ISSUE #741: Kotlin event sink should get more channel like
At the moment a Kotlin rendering accepts only one action, and throws if it receives a later one. This should be loosened up to instead accept actions for the entire lifetime of the WorkflowNode, to match the Swift semantics in #740.

zach-klippenstein:
Fixes #741, based on #743.

This change is actually very simple, it just replaces the per-render-pass `CompletableDeferred` with a per-`WorkflowNode`, unlimited-capacity `Channel`.


zach-klippenstein:
I will add that separately, it will probably be a bigger change.

PR #747, opened 2019-11-14
Make `lifecycleOrNull` public.

### LINKED ISSUE #734: Kotlin: Make Context.lifecycleOrNull public

zach-klippenstein:
Closes #734.

### DIFF
--- a/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/Lifecycles.kt
+++ b/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/Lifecycles.kt
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2019 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/Lifecycles.kt
```diff
@@ -0,0 +1,33 @@
...
+
+import android.content.Context
+import android.content.ContextWrapper
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.LifecycleOwner
+
+/**
+ * The [Lifecycle] for this context, or null if one can't be found.
+ *
+ * @implNote We keep all this very forgiving because we're just using it to keep some logging
```
rjrjr: Should drop the `implNote` entirely.

zach-klippenstein: It won't show up in rendered kdoc, if that's what you're concerned about.

rjrjr: Neat to know, but it's misleading. IIRC we use it for more than just that original reason these days.

zach-klippenstein: Removed.

zach-klippenstein:
Swift build errored, but this doesn't contain any swift changes and all the Android stuff is green, so I'm going to go ahead and merge.

PR #750, opened 2019-11-15
Add macOS support

timdonnelly:
For now this primarily marks `UIKit` code as unavailable when not compiling for iOS. The benefit today is that we can run unit tests directly on macOS.

As we add SwiftUI support in the future, a common API across Apple platforms becomes more interesting (which was the real motivation behind this PR).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -219,16 +219,16 @@ matrix:
         - bundle exec pod install
 
     - language: swift
-      name: "Swift - SPM - iOS (Xcode 11.2)"
+      name: "Swift - SPM - iOS (Xcode 11.1)"
       os: osx
-      osx_image: xcode11.2
+      osx_image: xcode11.1
       script:
         - xcodebuild -scheme "Workflow-Package" test -destination "name=iPhone 11"
 
     - language: swift
-      name: "Swift - SPM - macOS (Xcode 11.2)"
+      name: "Swift - SPM - macOS (Xcode 11.1)"
       os: osx
-      osx_image: xcode11.2
+      osx_image: xcode11.1
       script:
         - xcodebuild -scheme "Workflow-Package" test
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -219,11 +219,18 @@ matrix:
         - bundle exec pod install
 
     - language: swift
-      name: "Swift - SPM (Xcode 11.1)"
+      name: "Swift - SPM - iOS (Xcode 11.2)"
       os: osx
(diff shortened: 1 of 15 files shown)


AquaGeek:
Can we add macOS to the test matrix to ensure we don't break this going forward?

Also, I'm fine landing this, but it falls into the same bucket as the SwiftUI stuff of "we don't use it internally, so support will be limited."

swift/WorkflowUI/Sources/Container/ContainerViewController.swift
```diff
@@ -13,6 +13,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
+#if os(iOS)
```
bencochran: Would `#if canImport(UIKit)` be more appropriate? Not sure how the os check would behave in a Catalyst world.

timdonnelly: Oh right, it would work here.

timdonnelly:
> Can we add macOS to the test matrix to ensure we don't break this going forward?
> 
> Also, I'm fine landing this, but it falls into the same bucket as the SwiftUI stuff of "we don't use it internally, so support will be limited."

As far as I can tell the UIKit code has not materially changed since I initially committed it, so I'm not particularly worried about supporting this.

timdonnelly:
> Can we add macOS to the test matrix to ensure we don't break this going forward?

Added.

swift/WorkflowUI/Sources/Container/ContainerViewController.swift
timdonnelly: Updated.

.travis.yml
```diff
@@ -169,16 +169,16 @@ matrix:
         - bundle exec pod install
 
     - language: swift
-      name: "Swift - SPM - iOS (Xcode 11.2)"
+      name: "Swift - SPM - iOS (Xcode 11.1)"
```
zach-klippenstein: Why the downgrade?

timdonnelly: There is an issue with 11.2 that impacts SPM (it omits symbols in the compiled output when using generics in a few cases). It manifests as a crash. There are some compiler flags that I could add to work around it, but there's no real benefit to running CI on 11.2... so we can wait this one out until it's fixed in the next point release.

zach-klippenstein:
@timdonnelly Looks like your latest force-push clobbered some Travis changes?

PR #755, opened 2019-11-18
Fix RenderTester not checking worker keys.

### LINKED ISSUE #752: RenderTester doesn't assert worker keys match
If an expected worker matches but has the wrong key, it should fail the test. That does not happen.

zach-klippenstein:
Fixes #752.


zach-klippenstein:
Rebased on release branch after disabling instrumentation tests.

PR #756, opened 2019-11-18
Fix RenderTester not allowing output-less expectations to be specified after one with output.

### LINKED ISSUE #753: RenderTester will incorrectly fail on first worker expectation after a worker with an output, even if that Worker doesn't emit any outputs
Incorrectly reports that multiple workers emit.

E.g.
```kotlin
workflow.renderTester(…)
  .expectWorker(output = EmittedOutput(…))
  .expectWorker(output = null) // This one fails incorrectly.
```

As a workaround, just specify your output worker last.

zach-klippenstein:
Fixes #753.


zach-klippenstein:
Rebased on release branch after disabling instrumentation tests.

zach-klippenstein:
Both android shards passed, in the Pull Request build only a Swift shard errored, so I'm going to merge.

PR #757, opened 2019-11-18
DNM Increase timeout for emulator start, see if this fixes CI issues.

zach-klippenstein:


zach-klippenstein:
Still timed out, which makes sense given #758 showing that the emulator is failing to start. Closing.

Issue #759, opened 2019-11-18
Android CI is broken


zach-klippenstein:
The android shard for PRs over the last couple days have all been erroring out.

For now, **if you're merging a non-Kotlin/Android PR**, you should go ahead and merge if only the Android shard fails.

zach-klippenstein:
Cause is we time out waiting for the `android-wait-for-emulator` command.

Removing the run-in-background ampersand in the travis file (#758) shows the emulator command fails because it can't find a file:
```
78.43s$ emulator -verbose -avd test -no-accel -no-snapshot -no-window -no-audio -camera-back none -camera-front none -selinux permissive -qemu -m 2048
statvfs('/home/travis/.android/avd/test.avd/snapshots/default_boot/ram.img') failed: No such file or directory
emulator: ERROR: AdbHostServer.cpp:102: Unable to connect to adb daemon on port: 5037
/home/travis/.travis/functions: line 109:  4717 Segmentation fault      (core dumped) emulator -verbose -avd test -no-accel -no-snapshot -no-window -no-audio -camera-back none -camera-front none -selinux permissive -qemu -m 2048
The command "emulator -verbose -avd test -no-accel -no-snapshot -no-window -no-audio -camera-back none -camera-front none -selinux permissive -qemu -m 2048" failed and exited with 139 during .
```

zach-klippenstein:
@phoenixuprising suggested this can be caused by running out of memory.

Confirmed with @rjrjr offline, I'm going to disable instrumentation tests in CI for release branch so I can roll out some fixes, and then continue investigating on master.

zach-klippenstein:
Some ideas to try:

* [ ] Reduce memory config of emulator
  * Same error.
* [ ] Enable sudo to disable containerization (I think support for this was recently discontinued?)
  * Same error.
* [ ] Change emulator SDK version.
* [ ] Set `dist: trusty`, which is [apparently required](https://docs.travis-ci.com/user/languages/android/) even though we've (never?) done it.

zach-klippenstein:
Build is broken even on the first commit that introduced instrumentation tests: #767

zach-klippenstein:
Enabling sudo and explicitly setting the `dist` to `trusty` didn't work: #768

zach-klippenstein:
The first time we observed this issue was on [Nov 15th at 17:06, on one of @JustinDSN's builds](https://square.slack.com/archives/CE470RL5T/p1573866376162800).

[...]

PR #761, opened 2019-11-18
Split the Android instrumentation tests into their own Travis shard.

zach-klippenstein:
This allows the UI tests to fail/error independently, and hopefully allows the
tests to start without waiting for lint checks.

Hopefully makes issues like #759 easier to spot and workaround temporarily in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -98,9 +98,6 @@ matrix:
         # 82 is KEYCODE_MENU. Everyone seems to send this. Does anyone know why?
         - adb shell input keyevent 82
 
-      after_success:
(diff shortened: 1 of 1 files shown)


zach-klippenstein:
Branch build finished, Android & Kotlin shard is green, only instrumentation shard is red, as expected.

zach-klippenstein:
The builds on this PR should go red, since they still try to run instrumentation tests, but only that shard should fail. If that’s the case, I will merge as-is, and then continue trying to fix instrumentation tests in their own shard.

zach-klippenstein:
Second shard finished, everything's green but UI tests. Merging.

PR #762, opened 2019-11-18
Upgrade Gradle to 6.0.1.

zach-klippenstein:
Gradle 6.0 release notes: https://docs.gradle.org/6.0/release-notes.html
Gradle 6.0.1 release notes: https://docs.gradle.org/6.0.1/release-notes.html

This depends on upgrading ktlint first (#780), the version we're currently using doesn't work with Gradle 6.


zach-klippenstein:
Probably blocked by #759 for now.

PR #770, opened 2019-11-20
BREAKING CHANGE: ViewRegistry is now a ContainerHint.

### LINKED ISSUE #737: View Registry should be a container hint

rjrjr:
`LayoutRunner` constructors and `bindView` factory methods no longer have the
option to include a `ViewRegistry` parameter. Instead, every `ContainerHint`
map includes a hint pointing to the related `ViewRegistry`.

This further simplifies the chore of creating custom container views.
They no longer have to build a mechanism for getting their hands
on a `ViewRegistry` reference. In this PR you'll see various
`registry = viewRegistry` lines deleted from various factory functions
in various `BuilderBinding` calls, e.g. in `AlertContainer`,
`ModalViewContainer` and `BackStackContainer`.

Closes #737.

### DIFF
--- a/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ContainerHints.kt
+++ b/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ContainerHints.kt
@@ -19,13 +19,20 @@ import kotlin.reflect.KClass
 
 /**
  * Immutable, append-only map of values that a parent view can pass down to
- * its children via [android.view.View.showRendering] et al. Allows container views
- * to give descendants information about the context in which they're drawing.
+ * its children via [View.showRendering][android.view.View.showRendering] et al.
+ * Allows container views to give descendants information about the context in which
+ * they're drawing.
+ *
+ * Every [View.showRendering][android.view.View.showRendering] call initiated via
+ * [ViewRegistry.buildView] or [WorkflowLayout.start] includes an appropriate [ViewRegistry]
+ * hint. This allows container views to make recursive [ViewRegistry.buildView]
+ * calls to build child views to show nested renderings.
  */
 class ContainerHints private constructor(
   private val map: Map<ContainerHintKey<*>, Any>
 ) {
-  constructor() : this(emptyMap())
+  constructor(registry: ViewRegistry) :
+      this(mapOf<ContainerHintKey<*>, Any>(ViewRegistry to registry))
 
   @Suppress("UNCHECKED_CAST")
   operator fun <T : Any> get(key: ContainerHintKey<T>): T = map[key] as? T ?: key.default
--- a/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewRegistry.kt
+++ b/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewRegistry.kt
@@ -102,7 +102,12 @@ class ViewRegistry private constructor(
   ): View {
     @Suppress("UNCHECKED_CAST")
     return (allBindings[initialRendering::class] as? ViewBinding<RenderingT>)
-        ?.buildView(this, initialRendering, initialContainerHints, contextForNewView, container)
+        ?.buildView(
+            initialRendering,
+            initialContainerHints,
+            contextForNewView,
+            container
+        )
         ?.apply {
           checkNotNull(getRendering<RenderingT>()) {
             "View.bindShowRendering should have been called for $this, typically by the " +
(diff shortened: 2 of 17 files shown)


rjrjr:
`./gradlew clean build connectedCheck` passes locally.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewRegistry.kt
```diff
@@ -102,7 +102,12 @@ class ViewRegistry private constructor(
   ): View {
     @Suppress("UNCHECKED_CAST")
     return (allBindings[initialRendering::class] as? ViewBinding<RenderingT>)
-        ?.buildView(this, initialRendering, initialContainerHints, contextForNewView, container)
+        ?.buildView(
+            initialRendering,
+            initialContainerHints + (ViewRegistry to this),
```
zach-klippenstein: Isn't this always a no-op? When would `this` be different than the registry already stored in the hints?

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowViewStub.kt
```diff
@@ -79,9 +78,9 @@ class WorkflowViewStub @JvmOverloads constructor(
         }
 
     return when (val parent = actual.parent) {
-      is ViewGroup -> registry.buildView(rendering, containerHints, parent)
+      is ViewGroup -> containerHints[ViewRegistry].buildView(rendering, containerHints, parent)
```
zach-klippenstein: Could we provide a top-level `buildView` function that would have the same arguments, and just pull the `ViewRegistry` out of the hints itself? Would clean up all these call sites a little.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ContainerHints.kt
```diff
@@ -19,13 +19,20 @@ import kotlin.reflect.KClass
...
+ * [ViewRegistry.buildView] or [WorkflowLayout.start] includes an appropriate [ViewRegistry]
+ * hint. This allows container views to make recursive [ViewRegistry.buildView]
+ * calls to build child views to show nested renderings.
  */
 class ContainerHints private constructor(
   private val map: Map<ContainerHintKey<*>, Any>
 ) {
-  constructor() : this(emptyMap())
+  constructor(registry: ViewRegistry) :
+      this(mapOf<ContainerHintKey<*>, Any>((ViewRegistry to registry)))
```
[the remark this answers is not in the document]
rjrjr: You and me both, but the compiler insisted. I've seen that before, when the inferred type is more explicit than the wildcards on the LHS. Will fix the parens.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewRegistry.kt
rjrjr: Good catch, will nuke.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ContainerHints.kt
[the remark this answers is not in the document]
zach-klippenstein: I wonder if the new type inferencer is smart enough to figure this out…

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowViewStub.kt
rjrjr: Yeah, I like that.

rjrjr: Meh. Interferes with the convention of having companion objects implementing `ViewBinding` — the `buildView` function there takes precedent. And it's not all that much of a win, because most custom containers can use `WorkflowViewStub` instead of calling `buildView` themselves.

tehras:
Super small, but I noticed the `usage` example in `WorkflowViewStub` still has `viewRegistry` in the constructor.

rjrjr:
> Super small, but I noticed the usage example in WorkflowViewStub still has viewRegistry in the constructor.

Thanks!

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowViewStub.kt
zach-klippenstein: The other option is `ContainerHints.buildView` but that feels very wrong.

rjrjr: Yeah, I went through the same thought process.

It's fine, I don't think a lot of people are going to write this fragment. The main win from this PR is the ability to use `WorkflowViewStub` without having to figure out how to get your hands on a `ViewRegistry`. I'm content.

PR #772, opened 2019-11-21
[DNR] Try ARM emulator.

zach-klippenstein:


zach-klippenstein:
I copied LeakCanary's config line-for-line, and I'm still getting errors about `ANDROID_SDK_ROOT` not being defined. I checked LeakCanary's travis config and it doesn't have this envvar set there either.

@pyricau Are you setting `ANDROID_SDK_ROOT` for leakcanary builds some other way?

pyricau:
nope

PR #773, opened 2019-11-21
[DNR] Zachklipp/try sudo

zach-klippenstein:


zach-klippenstein:
Moving sudo to the right place didn't change anything.

PR #777, opened 2019-11-22
Fix some stuff in the RenderTester kdoc.

zach-klippenstein:


zach-klippenstein:
Everything other than UI tests passed, merging.

PR #780, opened 2019-11-22
Upgrade ktlint to latest version.

zach-klippenstein:


zach-klippenstein:
> What do you think about detekt ktlint extension, by the way? It can make things more centralized.

Sounds great!

PR #792, opened 2019-12-01
Add the ability for RenderTest to test onPropsChanged.

### LINKED ISSUE #791: [Kotlin] RenderTester should allow you to test `onPropsChanged`

zach-klippenstein:
Fixes #791.

### DIFF
--- a/kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RenderTester.kt
+++ b/kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RenderTester.kt
@@ -231,6 +231,14 @@ interface RenderTester<PropsT, StateT, OutputT : Any, RenderingT> {
   ): RenderTester<PropsT, StateT, OutputT, RenderingT>
 
   /**
+   * TODO kdoc
+   */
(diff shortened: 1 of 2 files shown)


zach-klippenstein:
Once this lands, rebase #788.

kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RenderTester.kt
```diff
@@ -230,6 +230,14 @@ interface RenderTester<PropsT, StateT, OutputT : Any, RenderingT> {
     output: EmittedOutput<Any?>? = null
   ): RenderTester<PropsT, StateT, OutputT, RenderingT>
 
+  /**
+   * TODO kdoc
```
JustinDSN: Should we do this TODO?

zach-klippenstein: Yep!

zach-klippenstein:
I'm pretty sure this is a super dumb PR – just call `onPropsChanged` directly if you want to test it.

PR #795, opened 2019-12-05
Write benchmarks specifically for render passes.

zach-klippenstein:
These new benchmarks only measure the render pass itself, by directly
interacting with `WorkflowNode`. The render pass is the most expensive
operation and the most interesting one. This is where all the work of starting
new nodes and workers and tearing them down happens, and is going to be a
target for optimizing soon (caching renderings, etc.).

A sample run of the benchmarks in this PR on my laptop:
```
Benchmark                                         (treeShape)  Mode  Cnt    Score    Error  Units
WorkflowNodeBenchmark.renderPassAlternateLeaves          DEEP  avgt    3  335.116 ± 38.784  us/op
WorkflowNodeBenchmark.renderPassAlternateLeaves         BUSHY  avgt    3  566.874 ± 88.549  us/op
WorkflowNodeBenchmark.renderPassAlternateLeaves        SQUARE  avgt    3  509.196 ± 16.121  us/op
WorkflowNodeBenchmark.renderPassAlternateWorkers         DEEP  avgt    3  314.860 ± 39.209  us/op
WorkflowNodeBenchmark.renderPassAlternateWorkers        BUSHY  avgt    3  553.105 ± 44.201  us/op
WorkflowNodeBenchmark.renderPassAlternateWorkers       SQUARE  avgt    3  492.601 ± 11.650  us/op
WorkflowNodeBenchmark.renderPassAlwaysLeaves             DEEP  avgt    3  155.881 ±  7.505  us/op
WorkflowNodeBenchmark.renderPassAlwaysLeaves            BUSHY  avgt    3  146.051 ±  7.314  us/op
WorkflowNodeBenchmark.renderPassAlwaysLeaves           SQUARE  avgt    3  134.365 ±  2.624  us/op
```


kotlin/workflow-runtime/src/jmh/java/com/squareup/workflow/WorkflowNodeBenchmark.kt
```diff
@@ -0,0 +1,168 @@
...
+   * @param childCount The number of children _each workflow in the tree_ should have.
+   * @param depth The depth of the workflow tree.
+   */
+  enum class TreeShape(
+    val childCount: Int,
+    val depth: Int
+  ) {
+    DEEP(childCount = 1, depth = 100),
+    BUSHY(childCount = 100, depth = 1),
+    SQUARE(childCount = 5, depth = 5);
```
rjrjr: Seems like it would be more consistent for SQUARE to be 10 x 10, so that you're always checking 100 nodes.

rjrjr: And if that's a bad idea for some reason, be good to document why.

kotlin/workflow-runtime/src/jmh/java/com/squareup/workflow/WorkflowNodeBenchmark.kt
```diff
@@ -0,0 +1,168 @@
...
+      initialProps = RENDER_LEAVES,
+      snapshot = null,
+      baseContext = context
+  )
+}
+
+private class FractalWorkflow(
+  private val childCount: Int,
+  private val depth: Int
+) : StatefulWorkflow<Props, Unit, Nothing, Unit>() {
```
rjrjr: Couldn't this be `StatelessWorkflow`?

kotlin/workflow-runtime/src/jmh/java/com/squareup/workflow/WorkflowNodeBenchmark.kt
zach-klippenstein: That's 10^10=10,000,000,000 nodes, which takes 5 minutes and then runs out of memory :P 

But yea we should keep it to the same overall node count for each.

kotlin/workflow-runtime/src/jmh/java/com/squareup/workflow/WorkflowNodeBenchmark.kt
zach-klippenstein: Yes, but I wanted to avoid the extra allocation.

zach-klippenstein: Will document.

kotlin/workflow-runtime/src/jmh/java/com/squareup/workflow/WorkflowNodeBenchmark.kt
zach-klippenstein: Done.

PR #802, opened 2019-12-07
Add a benchmark that just alternates a single leaf.

zach-klippenstein:


zach-klippenstein:
This measures the cost of starting/stopping a single node in the tree in the common case of an update occurring in that node, which is much more typical than either of the other cases, which are both worst cases.

Issue #319, opened 2019-04-25
Any plan to make workflow to support kotlin multiplatform?


fengdai:

zach-klippenstein:
Yes, we do eventually want to do that. @rjrjr actually started the work as a hackweek project a couple weeks ago, but I'm not sure where it ended up. We will probably do this after v1.0.0.

fengdai:
> Yes, we do eventually want to do that. @rjrjr actually started the work as a hackweek project a couple weeks ago, but I'm not sure where it ended up. We will probably do this after v1.0.0.

Fantastic!

rjrjr:
It didn't get too far since our public API is so dependent on Okio. But the good news is that they're hard at work on mp themselves. With that done, I don't see much blocking us.

akfpnd:
Any update so far? I willing to try how it works with KMP.

rjrjr:
@akfpnd there's some work in progress, as a matter of fact. https://github.com/square/workflow-kotlin/pull/746

PR #804, opened 2019-12-07
Migrate to multiplatform

ATizik:
#319 
Currently the only target platform is JVM

For native platforms there is a blocker:
Although state restoration isn't as much of a concern on iOS, it's still recommended in [the docs](https://developer.apple.com/documentation/uikit/view_controllers/preserving_your_app_s_ui_across_launches)
As far as I understand to support snapshots in Kotlin/Native targets, WorkflowId can't depend on KClass, because there are no Class.forName alternatives on Kotin/Native

Should WorkflowId be reworked, or maybe native platforms should just have no-op state restoration for now?


zach-klippenstein:
This is cool, thanks! Might be a while until we're able to land this unfortunately, we need to make sure it doesn't break anything internally and this month is pretty full already.

I believe `WorkflowId` only uses the workflow's `KClass` for comparison, and if that's true then we technically don't _need_ to rehydrate the actual class. I'm a little nervous about doing a bunch of long-ish string comparisons on every render, but we could mitigate that by changing the `==` to `===` and then canonicalizing the class names (e.g. using `intern()` on the JVM).

zach-klippenstein:
I'm going to close this PR, I'm starting to convert to multiplatform but doing so one module at a time. I'll link them to this issue as I go, first one is #1077.

PR #808, opened 2019-12-09
Activity can provide ContainerHints

### LINKED ISSUE #796: Add an optional ContainerHints argument to WorkflowLayout.start
And on `Activity.setContentWorkflow`. Should make it a lot easier for apps to pass in view-specific stuff. 

If this turns into a general cleanup, should also get `ViewRegistry` out of `WorkflowRunner.Config` and make it a `setContentWorkflow` argument as well. Much more natural set that on every `Activity.onCreate` call, `Config` shouldn't be about view concerns.

In the sample code, replace `Context.isTablet` with something that uses this.

rjrjr:
This allows activities / fragments to pass arbitrary information down the view system. (I had intended to update one of the samples to take advantage, but not of them really needed it.)

Closes #796

### DIFF
--- a/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowFragment.kt
+++ b/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowFragment.kt
@@ -17,7 +17,6 @@ package com.squareup.workflow.ui
 
 import android.os.Bundle
 import android.view.LayoutInflater
-import android.view.View
 import android.view.ViewGroup
 import androidx.fragment.app.Fragment
 import com.squareup.workflow.Workflow
@@ -34,7 +33,6 @@ import io.reactivex.Flowable
  *      override fun onCreateWorkflow(): Config<Unit, Unit> {
  *        return Config(
  *            workflow = HelloWorkflow,
- *            viewRegistry = ViewRegistry(HelloFragmentLayoutRunner),
  *            input = Unit
  *        )
  *      }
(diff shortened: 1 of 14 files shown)


kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowFragment.kt
```diff
@@ -61,22 +60,27 @@ abstract class WorkflowFragment<PropsT, OutputT : Any> : Fragment() {
...
    * Called from [onActivityCreated], so it should be safe for implementations
    * to call [getActivity].
    */
   protected abstract fun onCreateWorkflow(): Config<PropsT, OutputT>
 
+  /**
+   * Provides subclasses with access to the products of the running [Workflow].
+   * Safe to call after [onActivityCreated].
+   */
+  protected val runner: WorkflowRunner<OutputT> get() = _runner
```
zach-klippenstein: Is this property actually being used outside of this class? I think it can be private.

rjrjr: Intended to provide subclasses with access to the runner.

rjrjr:
Update cleans up some stale kdoc.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/WorkflowFragment.kt
zach-klippenstein: Do they need access to it?

Issue #815, opened 2019-12-11
[Swift] Proposal: Remove ViewRegistry


dhavalshreyas:
# [Swift] Proposal: Deprecate ViewRegistry

* Discussion: #390
* Authors: @bencochran, @dhavalshreyas
* Status: Proposal

## Overview
This issue proposes:
* Deprecating `ViewRegistry` in favor of expanding the `Screen` protocol to help build and update the `ViewController` to be rendered.
* Not requiring the rendered `ViewController` to inherit from `ScreenViewController`.

## Background & History
Currently, for a `Screen` type to be rendered, a view factory must first be registered using `viewRegistry.register(screenType:factory:)`.

This mapping can be easily missed during development and will throw a `fatalError` when a screen render is attempted.

`ViewRegistry` also requires the provided factory to return a `ScreenViewController` type.  `ScreenViewController` has a defined `initializer` and a `update(screen:)` method defined, which helps `ViewRegistry` build and update the `ViewController` as required. However, projects which already have a `BaseViewController` that all `ViewController`s are required to inherit from will have a tough time adopting `Workflow`.

## Goals
* Help make the `Workflow` APIs simpler and clearer.
* Provide a more flexible API to describe a `ViewController`.

## Non-Goals
> TODO

### 1. Introduce ViewControllerDescription

`ViewControllerDescription` is a struct which describes how to `build` and `update` a `ViewController`.

```swift
ViewControllerDescription(
    builder: { NewViewController() },
    updater: { viewController in
      viewController.update()
    }
)
```

### 2. Add `viewControllerDescription` property requirement to `Screen`
Introduce a property requirement on `Screen` protocol that returns `ViewControllerDescription`.

```swift
protocol Screen {
    var viewControllerDescription: ViewControllerDescription { get }
}
```

This lets us render `WelcomeScreen` using `WelcomeViewController` by adding:

```swift
extension WelcomeScreen: Screen {
    var viewControllerDescription: ViewControllerDescription {
        ViewControllerDescription(
            builder: { WelcomeViewController(screen: self) },
            updater: { $0.update(screen: self) })
    }
}
```

### 3. Remove `ViewRegistry`
Since each `Screen` can now describe how to both create and update a rendered `ViewController`, we will no longer need `ViewRegistry` and it can now be removed.

> **Related**: Requirement for a rendered `ViewController` to inherit form `ScreenViewController` will also be removed.

## Trade-off
[...]

PR #816, opened 2019-12-11
Soft-deprecate ViewRegistry

dhavalshreyas:
Provides a path for `ViewRegistry` migration without any breaking API changes.

Individual `Screen`s can move from using `register(screenViewControllerType:)` to providing a `viewControllerDescription`.

Initializers for both `ScreenViewController` and `ContainerViewController` have default values for `ViewRegistry`.

This is an awkward intermediate step since we're not gaining any compile-time safety here and are introducing some dangerous side-effects like:
* a `ViewControllerDescription` described `ViewController` cannot build a new `ViewController` using the `ViewRegistry`, since it will not have the correct `ViewRegistry` on init.
* `ViewControllerDescription` can only return implementations of `ScreenViewController`, if not, we'll see a run-time error.

This is the Migration strategy 1 as described in #815

### DIFF
--- a/swift/WorkflowUI/Sources/Screen/ScreenViewController.swift
+++ b/swift/WorkflowUI/Sources/Screen/ScreenViewController.swift
@@ -23,7 +23,7 @@ import UIKit
 /// given screen type.
 open class ScreenViewController<ScreenType: Screen>: UIViewController {
 
-    public final let viewRegistry: ViewRegistry
+    public final var viewRegistry: ViewRegistry
 
     private var currentScreen: ScreenType
 
@@ -35,7 +35,7 @@ open class ScreenViewController<ScreenType: Screen>: UIViewController {
         return ScreenType.self
     }
 
-    public required init(screen: ScreenType, viewRegistry: ViewRegistry) {
+    public required init(screen: ScreenType, viewRegistry: ViewRegistry = ViewRegistry()) {
         self.currentScreen = screen
         self.viewRegistry = viewRegistry
         super.init(nibName: nil, bundle: nil)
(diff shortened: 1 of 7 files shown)


dhavalshreyas:
Would love some input on this, however I'm out of town until Dec 19th, I will be responding to feedback then.

swift/WorkflowUI/Sources/ViewRegistry/ViewRegistry.swift
```diff
@@ -25,6 +25,8 @@ import ReactiveSwift
 /// In order for the registry to handle a given screen type, a view factory
 /// must first be registered using `register(screenType:factory:)`, where the
 /// factory is a simple closure that is responsible for instantiating a view.
+///
+/// Deprecated: Return the appropriate `ViewControllerDescription` from `Screen` instead
 public struct ViewRegistry {
```
AquaGeek: Should this be annotated as deprecated instead of commented as deprecated?

swift/WorkflowUI/Sources/Screen/ScreenViewController.swift
```diff
@@ -35,7 +35,7 @@ open class ScreenViewController<ScreenType: Screen>: UIViewController {
         return ScreenType.self
     }
 
-    public required init(screen: ScreenType, viewRegistry: ViewRegistry) {
+    public required init(screen: ScreenType, viewRegistry: ViewRegistry = ViewRegistry()) {
```
AquaGeek: Why the default arg?

swift/WorkflowUI/Sources/ViewRegistry/ViewRegistry.swift
dhavalshreyas: This triggers tons of warnings. Especially on all the `Screen` initializers. I decided to just use a comment here, and use the annotations for the `register` call.

JustinDSN:
Superseded by #974

dhavalshreyas:
This change landed here: https://github.com/square/workflow/pull/974

Issue #812, opened 2019-12-10
Simplify Kotlin WorkflowAction


rjrjr:
This is a simpler alternative to #644.

`WorkflowAction` is nice conceptually, but its Kotlin incarnation still requires too much ceremony. A handful of fixes should go a long way to streamlining its use:

- [ ] Eliminate `EventSink`, and instead make `RenderContext` itself the sink. (Or at least make `RenderContext` implement the interface itself, s.t. we can preserve `contramap` and such.)
- [x] Instead of returning the output value, give `Mutator` an output field that defaults to null. (Take care to disallow setting it to null.)
- [x] Give `StatelessWorkflow` the same workflowAction extension that `StatefulWorkflow` has
- [x] ~Provide a ktlint check that warns when event handler lambdas do anything other than call `context.update`~ (moved to #834)

zach-klippenstein:
Example:
```kotlin
class MyWorkflow : StatelessWorkflow<Foo, Bar, Baz>() {
  override fun render(
    props: Foo,
    context: RenderContext<Nothing, Bar>
  ): Baz {
    return Baz(
      onSelect = { selection -> context.update(myAction(selection)) }
    )
  }

  private fun myAction(selection: String) = workflowAction {
    analytics.log(MyEvent)
    output = Selected(selection)
  }
}
```

afollestad:
Excited for this to land in SPOS 👏

PR #817, opened 2019-12-11
BREAKING: WorkflowAction.Mutator deprecated by .Updater (#812, #813)

### LINKED ISSUE #813: Rename Mutator.state to something like newState so it doesn't conflict with state parameter in inline actions
When writing inline `WorkflowAction`s inside a `render` method of a `StatefulWorkflow`, an unqualified reference to `state` will resolve to the `render` method's `state` parameter, not the `state` property on the action's `Mutator` receiver. To get around this, you have to write `this.state` which feels clunky. 

Easiest way to get around this, and maybe also improve a little readability/self-documentation, is to just rename the `Mutator`'s `state` property to something else, like `newState` or `nextState`.

For migration purposes, we can leave the existing state property in there for now and deprecate it.

rjrjr:
First chunk of the work to reduce Kotlin's boilerplate and weirdness around
`WorkflowAction`, tracked by issue #812.

`Updater` has two fields for the action's `apply()` method to work with,
`var nextState`, and a write-only output field set via `setOutput()`.
This has two benefits:

 * You no longer use `return` to provide an output value.
 * Since `nextState` does not conflict with the `state`  argument of
   `render` methods, inline actions should be less painful, and people can
   stop telling us how terrible it was that we deprecated `emitState` and
   `emitOutput`.

Also adds `StatefulWorkflow.action` and `StatelessWorkflow.action` extension
functions. `StatefulWorkflow.workflowAction` is deprecated, as would
`StatelessWorkflow.workflowAction` have been if I had remembered to write it
in the first place.

This is a breaking change because the `OutputT` type on `WorkflowAction` and
`Workflow` now must be invariant instead of `out` (so that we can push output
values into `nextOutput`), and because the `WorkflowAction.companion.invoke`
factory functions couldn't be overloaded to work with both `Mutator` and `Updater`.

But it should not be necessary to rewrite every existing action implementation.
The `fun Mutator.apply(): OutputT?` function on the `WorkflowAction` interface
is still there for now, and the new `fun Updater.apply()` function has a default
implementation that calls the old one. (These will be deleted in a later
release, obv.)

Closes #813.

### DIFF
--- a/kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/GameWorkflow.kt
+++ b/kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/GameWorkflow.kt
@@ -34,8 +34,8 @@ import com.squareup.workflow.RenderContext
 import com.squareup.workflow.Snapshot
 import com.squareup.workflow.StatefulWorkflow
 import com.squareup.workflow.Worker
+import com.squareup.workflow.action
 import com.squareup.workflow.renderChild
-import com.squareup.workflow.workflowAction
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.flow
@@ -158,7 +158,7 @@ class GameWorkflow(
     playerRendering: Rendering,
     board: Board,
     aiRenderings: List<Pair<Location, ActorRendering>>
-  ) = workflowAction("updateGame") {
+  ) = action("updateGame") {
     // Calculate if this tick should result in movement based on the movement's speed.
     fun Movement.isTimeToMove(): Boolean {
       val ticksPerCell = (ticksPerSecond / cellsPerSecond).roundToLong()
@@ -191,12 +191,12 @@ class GameWorkflow(
     )
 
     // Check if AI captured player.
-    return@workflowAction if (newGame.isPlayerEaten) {
-      state = state.copy(game = newGame)
-      PlayerWasEaten
+    if (newGame.isPlayerEaten) {
+      nextState = nextState.copy(game = newGame)
+      setOutput(PlayerWasEaten)
     } else {
-      state = state.copy(game = newGame)
-      output
+      nextState = nextState.copy(game = newGame)
+      output?.let { setOutput(it) }
     }
   }
 }
--- a/kotlin/samples/hello-workflow-fragment/src/main/java/com/squareup/sample/helloworkflowfragment/HelloWorkflowFragment.kt
+++ b/kotlin/samples/hello-workflow-fragment/src/main/java/com/squareup/sample/helloworkflowfragment/HelloWorkflowFragment.kt
@@ -21,10 +21,10 @@ import com.squareup.workflow.ui.ViewRegistry
 import com.squareup.workflow.ui.WorkflowFragment
 import com.squareup.workflow.ui.WorkflowRunner
 
-class HelloWorkflowFragment : WorkflowFragment<Unit, Unit>() {
+class HelloWorkflowFragment : WorkflowFragment<Unit, Nothing>() {
   override val containerHints = ContainerHints(ViewRegistry(HelloFragmentLayoutRunner))
 
-  override fun onCreateWorkflow(): WorkflowRunner.Config<Unit, Unit> {
+  override fun onCreateWorkflow(): WorkflowRunner.Config<Unit, Nothing> {
     return WorkflowRunner.Config(
         HelloWorkflow, diagnosticListener = SimpleLoggingDiagnosticListener()
     )
--- a/kotlin/samples/todo-android/common/src/main/java/com/squareup/sample/todo/TodoListsAppWorkflow.kt
+++ b/kotlin/samples/todo-android/common/src/main/java/com/squareup/sample/todo/TodoListsAppWorkflow.kt
@@ -24,6 +24,7 @@ import com.squareup.workflow.RenderContext
 import com.squareup.workflow.Snapshot
 import com.squareup.workflow.StatefulWorkflow
 import com.squareup.workflow.WorkflowAction
+import com.squareup.workflow.action
 import com.squareup.workflow.ui.BackStackScreen
 
 private typealias TodoListsAction = WorkflowAction<TodoListsAppState, Nothing>
@@ -59,22 +60,20 @@ object TodoListsAppWorkflow :
   private val listsWorkflow = TodoListsWorkflow()
   private val editorWorkflow = TodoEditorWorkflow()
 
-  private fun onListSelected(index: Int): TodoListsAction = WorkflowAction {
-    state = EditingList(state.lists, index)
-    null
+  private fun onListSelected(index: Int): TodoListsAction = action {
+    nextState = EditingList(nextState.lists, index)
   }
 
   private fun onEditOutput(output: TodoEditorOutput): TodoListsAction = WorkflowAction {
-    state = when (output) {
+    nextState = when (output) {
       is ListUpdated -> {
-        val oldState = state as EditingList
+        val oldState = nextState as EditingList
         oldState.copy(
-            lists = state.lists.updateRow(oldState.editingIndex, output.newList)
+            lists = nextState.lists.updateRow(oldState.editingIndex, output.newList)
         )
       }
-      Done -> ShowingLists(state.lists)
+      Done -> ShowingLists(nextState.lists)
     }
-    null
   }
 
   override fun render(
--- a/kotlin/workflow-core/src/main/java/com/squareup/workflow/StatefulWorkflow.kt
+++ b/kotlin/workflow-core/src/main/java/com/squareup/workflow/StatefulWorkflow.kt
@@ -13,10 +13,13 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+@file:Suppress("DEPRECATION")
+
 package com.squareup.workflow
 
 import com.squareup.workflow.WorkflowAction.Companion.toString
 import com.squareup.workflow.WorkflowAction.Mutator
+import com.squareup.workflow.WorkflowAction.Updater
 
 /**
  * A composable, stateful object that can [handle events][RenderContext.makeActionSink],
@@ -66,7 +69,7 @@ import com.squareup.workflow.WorkflowAction.Mutator
 abstract class StatefulWorkflow<
     in PropsT,
     StateT,
-    out OutputT : Any,
+    OutputT : Any,
     out RenderingT
     > : Workflow<PropsT, OutputT, RenderingT> {
 
@@ -230,35 +233,62 @@ inline fun <StateT, OutputT : Any, RenderingT> Workflow.Companion.stateful(
 
 /**
  * Convenience to create a [WorkflowAction] with parameter types matching those
- * of the receiving [StatefulWorkflow]. The action will invoke the given [lambda][block]
+ * of the receiving [StatefulWorkflow]. The action will invoke the given [lambda][update]
  * when it is [applied][WorkflowAction.apply].
  *
- * The returned object will include the string returned from [name] in its [toString].
- *
- * @param name A string describing the update for debugging.
- * @param block Function that defines the workflow update.
+ * @param name A string describing the update for debugging, included in [toString].
+ * @param update Function that defines the workflow update.
  */
+fun <PropsT, StateT, OutputT : Any, RenderingT>
+    StatefulWorkflow<PropsT, StateT, OutputT, RenderingT>.action(
+      name: String = "",
+      update: Updater<StateT, OutputT>.() -> Unit
+    ) = action({ name }, update)
+
+/**
+ * Convenience to create a [WorkflowAction] with parameter types matching those
+ * of the receiving [StatefulWorkflow]. The action will invoke the given [lambda][update]
+ * when it is [applied][WorkflowAction.apply].
+ *
+ * @param name Function that returns a string describing the update for debugging, included
+ * in [toString].
+ * @param update Function that defines the workflow update.
+ */
+fun <PropsT, StateT, OutputT : Any, RenderingT>
+    StatefulWorkflow<PropsT, StateT, OutputT, RenderingT>.action(
+      name: () -> String,
+      update: Updater<StateT, OutputT>.() -> Unit
+    ): WorkflowAction<StateT, OutputT> = object : WorkflowAction<StateT, OutputT> {
+  override fun Updater<StateT, OutputT>.apply() = update.invoke(this)
+  override fun toString(): String = "action(${name()})-${this@action}"
+}
+
+@Deprecated(
+    message = "Use action",
+    replaceWith = ReplaceWith(
+        expression = "action(name) { nextState = state }",
+        imports = arrayOf("com.squareup.workflow.action")
+    )
+)
 fun <PropsT, StateT, OutputT : Any, RenderingT>
     StatefulWorkflow<PropsT, StateT, OutputT, RenderingT>.workflowAction(
       name: String = "",
       block: Mutator<StateT>.() -> OutputT?
     ) = workflowAction({ name }, block)
 
-/**
- * Convenience to create a [WorkflowAction] with parameter types matching those
- * of the receiving [StatefulWorkflow]. The action will invoke the given [lambda][block]
- * when it is [applied][WorkflowAction.apply].
- *
- * The returned object will include the string returned from [name] in its [toString].
- *
- * @param name Function that returns a string describing the update for debugging.
- * @param block Function that defines the workflow update.
- */
+@Suppress("OverridingDeprecatedMember")
+@Deprecated(
+    message = "Use action",
+    replaceWith = ReplaceWith(
+        expression = "action(name) { nextState = state }",
+        imports = arrayOf("com.squareup.workflow.action")
+    )
+)
 fun <PropsT, StateT, OutputT : Any, RenderingT>
     StatefulWorkflow<PropsT, StateT, OutputT, RenderingT>.workflowAction(
       name: () -> String,
       block: Mutator<StateT>.() -> OutputT?
     ): WorkflowAction<StateT, OutputT> = object : WorkflowAction<StateT, OutputT> {
   override fun Mutator<StateT>.apply() = block.invoke(this)
-  override fun toString(): String = "Action(${name()})-${this@workflowAction}"
+  override fun toString(): String = "workflowAction(${name()})-${this@workflowAction}"
 }
--- a/kotlin/workflow-core/src/main/java/com/squareup/workflow/StatelessWorkflow.kt
+++ b/kotlin/workflow-core/src/main/java/com/squareup/workflow/StatelessWorkflow.kt
@@ -15,6 +15,8 @@
  */
 package com.squareup.workflow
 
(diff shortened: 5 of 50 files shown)


kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowAction.kt
```diff
@@ -18,24 +18,41 @@
...
-   * @return an output value for the workflow to emit, or null to emit nothing.
-   */
-  fun Mutator<StateT>.apply(): OutputT?
+  class Updater<S, O : Any>(var nextState: S) {
+    private var _nextOutput: O? = null
 
-  companion object {
+    var nextOutput: O?
+      get() = _nextOutput
+      set(value) {
```
rjrjr: I'm not convinced that this is a good idea. It requires a lot of this kind of thing in order to avoid runtime exceptions:

```
      output?.let { nextOutput = it }
```

e.g. see line 199 of the Dungeon app's `GameWorkflow`, where I had to debug a crash.

kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/GameWorkflow.kt
```diff
@@ -191,12 +191,12 @@ class GameWorkflow(
...
-      state = state.copy(game = newGame)
-      PlayerWasEaten
+    if (newGame.isPlayerEaten) {
+      nextState = nextState.copy(game = newGame)
+      nextOutput = PlayerWasEaten
     } else {
-      state = state.copy(game = newGame)
-      output
+      nextState = nextState.copy(game = newGame)
+      output?.let { nextOutput = it }
```
rjrjr: This is nasty, and easy to get wrong.

kotlin/workflow-core/src/main/java/com/squareup/workflow/StatelessWorkflow.kt
```diff
@@ -108,6 +110,43 @@ fun <PropsT, OutputT : Any, FromRenderingT, ToRenderingT>
       transform: (FromRenderingT) -> ToRenderingT
     ): Workflow<PropsT, OutputT, ToRenderingT> = Workflow.stateless { props ->
   // @formatter:on
-  renderChild(this@mapRendering, props) { output -> WorkflowAction({ "mapRendering" }) { output } }
-      .let(transform)
+  renderChild(this@mapRendering, props) { output ->
+    @Suppress("DEPRECATION")
```
rjrjr: Unnecessary.

kotlin/workflow-core/src/main/java/com/squareup/workflow/StatelessWorkflow.kt
```diff
@@ -108,6 +110,43 @@ fun <PropsT, OutputT : Any, FromRenderingT, ToRenderingT>
...
+    WorkflowAction({ "mapRendering" }) { nextOutput = output }
+  }.let(transform)
+}
+
+/**
+ * Convenience to create a [WorkflowAction] with parameter types matching those
+ * of the receiving [StatefulWorkflow]. The action will invoke the given [lambda][block]
+ * when it is [applied][WorkflowAction.apply].
+ *
+ * The returned object will include the string returned from [name] in its [toString].
```
rjrjr: name isn't a lambda

kotlin/workflow-testing/src/test/java/com/squareup/workflow/TreeWorkflow.kt
```diff
@@ -76,8 +76,7 @@ internal class TreeWorkflow(
       it.writeUtf8WithLength(state)
     }
 
-  private fun onEvent(newState: String) = workflowAction {
-    state = newState
-    null
+  private fun onEvent(newState: String) = action {
+    nextState = newState
```
rjrjr: Wait, oops maybe?

kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowAction.kt
```diff
@@ -18,24 +18,41 @@
...
+  @Deprecated("Use Updater")
   class Mutator<S>(var state: S)
 
-  /**
-   * Called with the current state of a workflow, allowing the state to be updated.
-   *
-   * @return an output value for the workflow to emit, or null to emit nothing.
-   */
-  fun Mutator<StateT>.apply(): OutputT?
+  class Updater<S, O : Any>(var nextState: S) {
```
rjrjr: kdoc

kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowAction.kt
```diff
@@ -18,24 +18,41 @@
 package com.squareup.workflow
 
 import com.squareup.workflow.WorkflowAction.Companion.toString
-import com.squareup.workflow.WorkflowAction.Mutator
+import com.squareup.workflow.WorkflowAction.Updater
 
 /**
  * A function that can change the current state of a [Workflow] by returning a new one, and
```
rjrjr: Update

kotlin/workflow-testing/src/test/java/com/squareup/workflow/testing/RealRenderTesterTest.kt
```diff
@@ -41,11 +41,18 @@ class RealRenderTesterTest {
   @Test fun `expectWorkflow with output throws when already expecting workflow output`() {
     // Don't need an implementation, the test should fail before even calling render.
     val workflow = Workflow.stateless<Unit, Nothing, Unit> {}
+    @Suppress("UNCHECKED_CAST")
     val tester = workflow.renderTester(Unit)
-        .expectWorkflow(ChildWorkflow::class, rendering = Unit, output = EmittedOutput(Unit))
+        .expectWorkflow(
+            ChildWorkflow::class, rendering = Unit,
+            output = EmittedOutput(Unit) as EmittedOutput<Nothing>
```
kirillzh: Casting is strange, `EmittedOutput `with `Nothing` shouldn't be possible. Would it make sense to convert `EmittedOutput` to a sealed class - first implementation contains an actual output value, second implementation is just a singleton representing `Nothing` for emitted output?

kotlin/workflow-runtime/src/test/java/com/squareup/workflow/internal/RealRenderContextTest.kt
```diff
@@ -209,6 +214,7 @@ class RealRenderContextTest {
     val childCases = context.buildBehavior()
         .childCases
     assertEquals(1, childCases.size)
+    // IDE is lying, this is fine.
```
kirillzh: Was there a warning of some sort? I'm not seeing it :)

kotlin/workflow-runtime/src/test/java/com/squareup/workflow/internal/RealRenderContextTest.kt
```diff
@@ -172,8 +173,12 @@ class RealRenderContextTest {
   }
 
   @Test fun `makeEventSink works with OutputT of Nothing`() {
-    val context = RealRenderContext<String, Nothing>(PoisonRenderer(), eventActionsChannel)
-    val sink: Sink<String> = context.makeEventSink { null }
+    @Suppress("UNCHECKED_CAST")
+    val context = RealRenderContext(
+        PoisonRenderer(),
+        eventActionsChannel as SendChannel<WorkflowAction<String, Nothing>>
```
kirillzh: Why can't be infer the type here? So many new unchecked casts introduced...

rjrjr: Beats me, something to do w/the loss of `out` on `OutputT`, but the compiler wants what the compiler wants.

All the casts are in unit tests of the machinery itself, so I'm not concerned.

kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowAction.kt
```diff
@@ -83,70 +110,85 @@ interface WorkflowAction<StateT, out OutputT : Any> {
...
       }
 
     /**
      * Convenience function to implement [WorkflowAction] without changing the state.
      */
-    @Deprecated("Use `WorkflowAction { output }`")
+    @Deprecated(
+        "Implement WorkflowAction directly, " +
+            "or use StatefulWorkflow.action or StatelessWorkflow.action"
+    )
```
kirillzh: Adding replacement strategy here and for other deprecated APIs would make migration easier:
```kotlin
@Deprecated(
    message = "Implement WorkflowAction directly, " +
        "or use StatefulWorkflow.action or StatelessWorkflow.action",
    replaceWith = ReplaceWith(
        expression = "action { nextOutput = output }",
        imports = arrayOf("com.squareup.workflow.action")
    )
)
```

kotlin/samples/todo-android/common/src/main/java/com/squareup/sample/todo/TodoListsAppWorkflow.kt
```diff
@@ -59,22 +60,20 @@ object TodoListsAppWorkflow :
   private val listsWorkflow = TodoListsWorkflow()
   private val editorWorkflow = TodoEditorWorkflow()
 
-  private fun onListSelected(index: Int): TodoListsAction = WorkflowAction {
-    state = EditingList(state.lists, index)
-    null
```
kirillzh: I really like this change, no need to explicitly returning `null` anymore in cases like this (most of the cases from my experience).

kotlin/workflow-testing/src/test/java/com/squareup/workflow/testing/RealRenderTesterTest.kt
rjrjr: Like you said, this is not a normal thing to do. Not inclined to change `EmittedOutput` to be able to express nonsense. Frankly, I was just banging on these tests as little as possible to keep them working.

I could take a stab at changing `workflow`'s output type to `String` or `Unit`, but I don't want to sink a lot of time in here.

kotlin/workflow-runtime/src/test/java/com/squareup/workflow/internal/RealRenderContextTest.kt
rjrjr: Type inference failed error. 🤷‍♂

kotlin/samples/hello-workflow-fragment/src/main/java/com/squareup/sample/helloworkflowfragment/HelloWorkflowFragment.kt
```diff
@@ -21,10 +21,10 @@ import com.squareup.workflow.ui.ViewRegistry
 import com.squareup.workflow.ui.WorkflowFragment
 import com.squareup.workflow.ui.WorkflowRunner
 
-class HelloWorkflowFragment : WorkflowFragment<Unit, Unit>() {
+class HelloWorkflowFragment : WorkflowFragment<Unit, Nothing>() {
```
zach-klippenstein: Nice catch

kotlin/workflow-runtime/src/test/java/com/squareup/workflow/internal/RealRenderContextTest.kt
rjrjr: The casts were all due to laziness, re-using helper stubs whose declared types didn't match the requirements in the specific test case. All fixed.

kotlin/workflow-core/src/main/java/com/squareup/workflow/StatelessWorkflow.kt
```diff
@@ -108,6 +110,39 @@ fun <PropsT, OutputT : Any, FromRenderingT, ToRenderingT>
...
+ * of the receiving [StatefulWorkflow]. The action will invoke the given [lambda][block]
+ * when it is [applied][WorkflowAction.apply].
+ *
+ * @param name A string describing the update for debugging, included in [toString].
+ * @param block Function that defines the workflow update.
+ */
+fun <PropsT, OutputT : Any, RenderingT>
+    StatelessWorkflow<PropsT, OutputT, RenderingT>.action(
+      name: String = "",
+      block: Updater<Unit, OutputT>.() -> Unit
```
zach-klippenstein: Nit: Call these blocks `update`?

rjrjr:
Update fixes jmh, and lazy casting in some unit tests.

kotlin/workflow-core/src/main/java/com/squareup/workflow/StatelessWorkflow.kt
```diff
@@ -108,6 +110,39 @@ fun <PropsT, OutputT : Any, FromRenderingT, ToRenderingT>
...
+  }.let(transform)
+}
+
+/**
+ * Convenience to create a [WorkflowAction] with parameter types matching those
+ * of the receiving [StatefulWorkflow]. The action will invoke the given [lambda][block]
+ * when it is [applied][WorkflowAction.apply].
+ *
+ * @param name A string describing the update for debugging, included in [toString].
+ * @param block Function that defines the workflow update.
```
zach-klippenstein: Nit:
```suggestion
 * @param block Function that defines the workflow update by setting properties on the [Updater].
```

kotlin/workflow-core/src/main/java/com/squareup/workflow/StatefulWorkflow.kt
```diff
@@ -233,32 +236,47 @@ inline fun <StateT, OutputT : Any, RenderingT> Workflow.Companion.stateful(
...
+fun <PropsT, StateT, OutputT : Any, RenderingT>
+    StatefulWorkflow<PropsT, StateT, OutputT, RenderingT>.action(
+      name: () -> String,
+      block: Updater<StateT, OutputT>.() -> Unit
+    ): WorkflowAction<StateT, OutputT> = object : WorkflowAction<StateT, OutputT> {
+  override fun Updater<StateT, OutputT>.apply() = block.invoke(this)
+  override fun toString(): String = "action(${name()})-${this@action}"
+}
+
+@Deprecated("Use action")
```
zach-klippenstein: Nit: Can we provide a `replaceWith`?

kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowAction.kt
zach-klippenstein: I don't think it's a good idea either, I was thinking this parameter would be non-nullable (and it can throw if you try reading before writing). In that case, maybe this should actually be a write-only method instead of a var.

kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowAction.kt
```diff
@@ -18,24 +18,51 @@
...
+      get() = _nextOutput
+      set(value) {
+        _nextOutput = requireNotNull(value)
+      }
+  }
 
+  @Suppress("DEPRECATION")
+  @Deprecated("Implement Updater.apply")
+  fun Mutator<StateT>.apply(): OutputT? {
+    throw UnsupportedOperationException()
```
zach-klippenstein: If this is just going to throw anyway, should the deprecation level be `Error`? If it's always going to fail anyway, might as well fail at compile time. But in that case I'm not sure what the point of leaving this method around is, since it will still require one giant pass to update.

kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowAction.kt
zach-klippenstein: In particular, relying on nullability here is going to make it harder to remove the non-nullable bound on `OutputT` in the future.

kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowAction.kt
```diff
@@ -83,70 +110,85 @@ interface WorkflowAction<StateT, out OutputT : Any> {
...
   }
 }
 
 fun <StateT, OutputT : Any> WorkflowAction<StateT, OutputT>.applyTo(
   state: StateT
 ): Pair<StateT, OutputT?> {
-  val box = Mutator(state)
-  val output = box.apply()
-  return Pair(box.state, output)
+  val box = Updater<StateT, OutputT>(state)
```
zach-klippenstein: Nit: s/`box`/`updater`/g

kotlin/workflow-runtime/src/test/java/com/squareup/workflow/internal/RealRenderContextTest.kt
zach-klippenstein: Did you have new inference enabled?

kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowAction.kt
rjrjr: I like the setter-method-only notion. We don't owe anyone an `output` field.

kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/GameWorkflow.kt
rjrjr: You couldn't, I'm down.

kotlin/samples/hello-workflow-fragment/src/main/java/com/squareup/sample/helloworkflowfragment/HelloWorkflowFragment.kt
rjrjr: Thank the compiler. A bonus of our new invariance.

kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowAction.kt
rjrjr: It won't require a single pass. If you implement your own Updater.apply() then this method will never be called. Nothing calls it except the default Updater.apply(). Existing actions will work unchanged. (I'll double check that, leave one of the sample unconverted and just suppress its warnings.)

kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowAction.kt
rjrjr:
```kotlin
    /**
     * Sets the value workflow will emit when this action is applied. If this method
     * is not called, there will be no output.
     */
    fun setOutput(output: O)
```

kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowAction.kt
rjrjr: Confirmed.

kotlin/workflow-core/src/main/java/com/squareup/workflow/WorkflowAction.kt
rjrjr: I've added them, but as usual I'm finding that they don't actually work. 

I suppose they at least get the import in place, and provide a bit more documentation.

kirillzh: In my experience replacements do work in IDE but imports are not always added for whatever reason. 😕

PR #827, opened 2019-12-16
Echo y to the gem update command to accept bundle executable overwrite prompt.

zach-klippenstein:


zach-klippenstein:
> Seems fine 🤷‍♂

image

Computer says no.

JustinDSN:
I would try removing the `gem update --system` as we could probably use the Travis provided bundler that's installed.

zach-klippenstein:
> I would try removing the gem update --system as we could probably use the Travis provided bundler that's installed.

See the commits in this PR – I tried that and I tried adding `--force --quiet` flags, and neither of those worked. The last commit just echos `y` to the command, if that doesn't work I give up for now.

.travis.yml
```diff
@@ -25,7 +25,7 @@ matrix:
       os: osx
       osx_image: xcode10.2
       install:
-        - gem update --system
+        - echo y | gem update --system
```
AquaGeek: `yes | gem update --system`?

zach-klippenstein: 😲

zach-klippenstein: Done

zach-klippenstein:
Re-requested Tyler's review before seeing he'd already looked at this again, so I'm going to merge once it goes green.

PR #828, opened 2019-12-16
Improve renderTester API

### LINKED ISSUE #814: Make RenderTester.render lambda optional
Tests that only care about verifying workflow actions often pass an empty lambda, which is unnecessary boilerplate.

zach-klippenstein:
- Give RenderTester.render function argument a default no-op value. Fixes #814.
- Add overload to RenderTester.expectWorker for simple worker comparisons.
- Improve documentation on state-less renderTester.

### DIFF
--- a/kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RenderTester.kt
+++ b/kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RenderTester.kt
@@ -248,6 +248,33 @@ interface RenderTester<PropsT, StateT, OutputT : Any, RenderingT> {
 }
 
 /**
+ * Specifies that this render pass is expected to run a particular worker.
+ *
+ * @param doesSameWorkAs Worker passed to the actual worker's
+ * [doesSameWorkAs][Worker.doesSameWorkAs] method to identify the worker. Note that the actual
+ * method is called on the worker instance given by the workflow-under-test, and the value of this
+ * argument is passed to that method – if you need custom comparison logic for some reason, use
+ * the overload of this method that takes a `matchesWhen` parameter.
+ * @param key The key passed to [runningWorker][com.squareup.workflow.RenderContext.runningWorker]
+ * when rendering this workflow.
+ * @param output If non-null, [EmittedOutput.output] will be emitted when this worker is ran.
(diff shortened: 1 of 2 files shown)


kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RenderTester.kt
```diff
@@ -244,9 +245,36 @@ interface RenderTester<PropsT, StateT, OutputT : Any, RenderingT> {
    * @return A [RenderTestResult] that can be used to verify the [WorkflowAction] that was used to
    * handle a workflow or worker output or a rendering event.
    */
-  fun render(block: (RenderingT) -> Unit): RenderTestResult<StateT, OutputT>
+  fun render(block: (rendering: RenderingT) -> Unit = {}): RenderTestResult<StateT, OutputT>
```
rjrjr: I completely forgot that you could name function arguments. Should be doing this all over the place.

kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RenderTester.kt
```diff
@@ -244,9 +245,36 @@ interface RenderTester<PropsT, StateT, OutputT : Any, RenderingT> {
...
+ * argument is passed to that method – if you need custom comparison logic for some reason, use
+ * the overload of this method that takes a `matchesWhen` parameter.
+ * @param key The key passed to [runningWorker][com.squareup.workflow.RenderContext.runningWorker]
+ * when rendering this workflow.
+ * @param output If non-null, [EmittedOutput.output] will be emitted when this worker is ran.
+ * The [WorkflowAction] used to handle this output can be verified using methods on
+ * [RenderTestResult].
+ */
+/* ktlint-disable parameter-list-wrapping */
+fun <PropsT, StateT, OutputT : Any, RenderingT>
```
rjrjr: Nice, should be handy.

zach-klippenstein:
UI test shard failed, which doesn't make any sense, and it passes locally. Gonna try just restarting it.

zach-klippenstein:
All non-iOS shards passed, merging.

PR #829, opened 2019-12-16
Disable ktlint and let IntelliJ format our weird long function signatures.

zach-klippenstein:


zach-klippenstein:
All non-iOS shards have passed, merging.

PR #832, opened 2019-12-18
Convert ViewRegistry to an interface.

### LINKED ISSUE #831: ViewRegistry should be an interface
This allows us to migrate a bit more incrementally off our internal `ViewRegistry`-like concept internally.

zach-klippenstein:
Closes #831.

### DIFF
--- a/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/CompositeViewRegistry.kt
+++ b/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/CompositeViewRegistry.kt
@@ -1,0 +1,79 @@
+/*
+ * Copyright 2019 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.workflow.ui
+
+import android.content.Context
+import android.view.View
+import android.view.ViewGroup
+
+/**
+ * A [ViewRegistry] that contains only other registries and delegates to their [buildView] methods.
+ *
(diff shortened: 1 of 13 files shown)


kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ContainerHints.kt
```diff
@@ -30,7 +30,7 @@ class ContainerHints private constructor(
   private val map: Map<ContainerHintKey<*>, Any>
 ) {
   constructor(registry: ViewRegistry) :
-      this(mapOf<ContainerHintKey<*>, Any>(ViewRegistry to registry))
+      this(mapOf<ContainerHintKey<*>, Any>(ViewRegistry to (registry + defaultViewBindings)))
```
zach-klippenstein: @rjrjr I'm not a fan of doing this here, but I didn't see any other obvious place to put it. This is the one choke point where all ViewRegistries have to go through to get into the runtime.

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewRegistry.kt
```diff
@@ -13,13 +13,23 @@
...
-import kotlin.reflect.KClass
+
+/**
+ * [ViewBinding]s that are always available.
+ */
+internal val defaultViewBindings = ViewRegistry(
+    NamedBinding,
+    BackStackContainer,
+    ModalContainer.forAlertContainerScreen()
+)
```
zach-klippenstein: These are wired up in `WorkflowLayout` instead of being hard-coded into any particular `ViewRegistry` class.

zach-klippenstein:
This is ready for review!

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/CompositeViewRegistry.kt
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * A [ViewRegistry] that contains only other registries and delegates to their [buildView] methods.
+ *
+ * Whenever any registries are combined using the [ViewRegistry] factory functions or `plus`
+ * operators, an instance of this class is returned. All registries' keys are checked at
+ * construction to ensure that no duplicate keys exist.
+ *
+ * The implementation of [buildView] consists of a single layer of indirection – the responsible
+ * [ViewRegistry] is looked up in a map by key, and then that registry's [buildView] is called.i
```
rjrjr:
```suggestion
 * [ViewRegistry] is looked up in a map by key, and then that registry's [buildView] is called.
```

PR #835, opened 2019-12-18
Make RenderContext itself a Sink.

### LINKED ISSUE #812: Simplify Kotlin WorkflowAction

zach-klippenstein:
Closes #812.

### DIFF
--- a/kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/PlayerWorkflow.kt
+++ b/kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/PlayerWorkflow.kt
@@ -65,15 +65,11 @@ class PlayerWorkflow(
     props: ActorProps,
     state: Movement,
     context: RenderContext<Movement, Nothing>
-  ): Rendering {
-    val sink = context.makeActionSink<Action>()
-
-    return Rendering(
-        actorRendering = ActorRendering(avatar = avatar, movement = state),
-        onStartMoving = { sink.send(StartMoving(it)) },
-        onStopMoving = { sink.send(StopMoving(it)) }
-    )
-  }
+  ): Rendering = Rendering(
+      actorRendering = ActorRendering(avatar = avatar, movement = state),
+      onStartMoving = { context.send(StartMoving(it)) },
+      onStopMoving = { context.send(StopMoving(it)) }
+  )
 
   override fun snapshotState(state: Movement): Snapshot = Snapshot.EMPTY
 }
--- a/kotlin/samples/todo-android/common/src/main/java/com/squareup/sample/todo/TodoEditorWorkflow.kt
+++ b/kotlin/samples/todo-android/common/src/main/java/com/squareup/sample/todo/TodoEditorWorkflow.kt
@@ -97,15 +97,13 @@ class TodoEditorWorkflow : StatelessWorkflow<TodoList, TodoEditorOutput, TodoRen
     props: TodoList,
     context: RenderContext<Nothing, TodoEditorOutput>
   ): TodoRendering {
-    val rawSink = context.makeActionSink<TodoAction>()
-
     // Make event handling idempotent until https://github.com/square/workflow/issues/541 is fixed.
     var eventFired = false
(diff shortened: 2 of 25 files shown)


kotlin/workflow-core/src/main/java/com/squareup/workflow/RenderContext.kt
```diff
@@ -26,6 +26,7 @@ import com.squareup.workflow.WorkflowAction.Updater
  *
  * ## Handling Events
  *
+ * TODO update
```
zach-klippenstein: TODO @zach-klippenstein Update docs

kotlin/samples/todo-android/common/src/main/java/com/squareup/sample/todo/TodoEditorWorkflow.kt
```diff
@@ -97,15 +97,13 @@ class TodoEditorWorkflow : StatelessWorkflow<TodoList, TodoEditorOutput, TodoRen
...
-    val rawSink = context.makeActionSink<TodoAction>()
-
     // Make event handling idempotent until https://github.com/square/workflow/issues/541 is fixed.
     var eventFired = false
     val sink = object : Sink<TodoAction> {
       override fun send(value: TodoAction) {
         if (eventFired) return
         eventFired = true
-        rawSink.send(value)
+        context.send(value)
```
rjrjr: I'm increasingly filled with regret that we didn't go with `RenderContext.render(…)` as the method signature.

rjrjr: Just thinking out loud, don't want to distract us from this release, but I bet we could retrofit that pretty easily. Same trick we're doing with `WorkflowAction` — introduce the new method with a default impl that calls the old method, etc.

kotlin/samples/dungeon/common/src/main/java/com/squareup/sample/dungeon/PlayerWorkflow.kt
```diff
@@ -65,15 +65,11 @@ class PlayerWorkflow(
...
-    return Rendering(
-        actorRendering = ActorRendering(avatar = avatar, movement = state),
-        onStartMoving = { sink.send(StartMoving(it)) },
-        onStopMoving = { sink.send(StopMoving(it)) }
-    )
-  }
+  ): Rendering = Rendering(
+      actorRendering = ActorRendering(avatar = avatar, movement = state),
+      onStartMoving = { context.send(StartMoving(it)) },
+      onStopMoving = { context.send(StopMoving(it)) }
```
kirillzh: 😍

kotlin/samples/todo-android/common/src/main/java/com/squareup/sample/todo/TodoEditorWorkflow.kt
zach-klippenstein: For sure. See #836.

kotlin/workflow-core/src/main/java/com/squareup/workflow/RenderContext.kt
zach-klippenstein: Done.

zach-klippenstein:
Addressed @rjrjr's comments, and also cleaned up `RealRenderContextTest`'s actions that were using the deprecated `apply` method.

PR #837, opened 2019-12-19
Make WorkflowAction covariant on OutputT again.

zach-klippenstein:
`WorkflowAction`'s output type is, as the name suggests, an output value,
and so it is natural that it have `out` variance (i.e. be covariant). This was changed because
the new `Updater` API stores the pending output value in a var, which is readable
as well as writeable and therefore an invariant (`in`) position, and therefore `Updater`'s `O`
must be neither `in` nor `out` (i.e. be invariant), and `WorkflowAction`'s `OutputT` must, in turn, also be invariant.
However, the only _API_ in which `Updater` exposes this type is as a method parameter,
which is an `in` position. It is only _read_ in the same context as the `WorkflowAction` for which
it was created, which means its type will never actually be out of bounds. Therefore, we can
just mark the output use of `Updater.O` as `@UnsafeVariance`, mark the type parameter itself
as contravariant (`in`), and revert `WorkflowAction`'s `OutputT` back to `out`.

### DIFF
--- a/kotlin/workflow-core/src/main/java/com/squareup/workflow/RenderContext.kt
+++ b/kotlin/workflow-core/src/main/java/com/squareup/workflow/RenderContext.kt
@@ -50,7 +50,7 @@ import com.squareup.workflow.WorkflowAction.Updater
  *
  * See [renderChild].
  */
-interface RenderContext<StateT, OutputT : Any> : Sink<WorkflowAction<StateT, OutputT>> {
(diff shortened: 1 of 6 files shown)


rjrjr:
Don't you want to unwind the parallel changes made to `Workflow` and `RenderContext` in the same PR?

zach-klippenstein:
Holy crap, how did I miss those in code review the first time. Yes definitely.

zach-klippenstein:
I think that's all of them…

zach-klippenstein:
All non-swift shards passed, merging.

PR #839, opened 2019-12-19
Undo making RenderContext a sink, give it a sink property instead.

zach-klippenstein:
It looks weird to say `context.send`, and weirder to say `context.contraMap` –
what does it mean to "contra map" a `RenderContext`? It makes more sense to say
a context _has_ a sink than it _is_ a sink. Only slightly more typing but much
more clear.

### DIFF
--- a/kotlin/workflow-core/src/main/java/com/squareup/workflow/RenderContext.kt
+++ b/kotlin/workflow-core/src/main/java/com/squareup/workflow/RenderContext.kt
@@ -50,21 +50,21 @@ import com.squareup.workflow.WorkflowAction.Updater
  *
  * See [renderChild].
  */
-interface RenderContext<StateT, in OutputT : Any> : Sink<WorkflowAction<StateT, OutputT>> {
-
-  @Deprecated("Use RenderContext.send.")
-  fun <EventT : Any> onEvent(
-    handler: (EventT) -> WorkflowAction<StateT, OutputT>
-  ): (EventT) -> Unit
+interface RenderContext<StateT, in OutputT : Any> {
 
   /**
-   * Accepts a single [WorkflowAction], and invokes that action by calling [WorkflowAction.apply]
+   * Accepts a single [WorkflowAction], invokes that action by calling [WorkflowAction.apply]
    * to update the current state, and optionally emits the returned output value if it is non-null.
    *
    * This method is defined by the [Sink] interface. Since [RenderContext] implements [Sink],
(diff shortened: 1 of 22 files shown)


kotlin/workflow-core/src/main/java/com/squareup/workflow/RenderContext.kt
```diff
@@ -178,7 +178,7 @@ fun <StateT, OutputT : Any> RenderContext<StateT, OutputT>.runningWorker(
  */
 fun <EventT, StateT, OutputT : Any> RenderContext<StateT, OutputT>.makeEventSink(
```
rjrjr: Could you add `Sink.asEventSink` and make this wrap that? And deprecate this.

Too much?

zach-klippenstein: Oo, I like.

zach-klippenstein: Actually, not sure. It would have to be an extension on `Sink<WorkflowAction<S, O>>`, and I'm not sure what the use case for having one of those that is not `actionSink` would be. I'm gonna leave it out for now.

kotlin/workflow-core/src/main/java/com/squareup/workflow/RenderContext.kt
```diff
@@ -50,21 +50,21 @@ import com.squareup.workflow.WorkflowAction.Updater
...
+   * Accepts a single [WorkflowAction], invokes that action by calling [WorkflowAction.apply]
    * to update the current state, and optionally emits the returned output value if it is non-null.
    *
    * This method is defined by the [Sink] interface. Since [RenderContext] implements [Sink],
    * operations like [contraMap] are available.
    */
-  override fun send(value: WorkflowAction<StateT, OutputT>)
+  val actionSink: Sink<WorkflowAction<StateT, OutputT>>
+
+  @Deprecated("Use RenderContext.send.")
```
kirillzh: Should this be `RenderContext.actionSink.send`? Also, possible to add a replacement strategy?

zach-klippenstein: Oops

zach-klippenstein: #841

PR #844, opened 2019-12-20
#842: Detect possible Worker mock setup issues.

### LINKED ISSUE #842: Detect null values returned from Worker.run and throw more helpful exception
In unit tests, if you use Mockito to create a Worker, the `run` method will return `null` even though the return type is non-nullable in Kotlin. Kotlin helps out with this by throwing an NPE before control is returned to any kotlin code, but the NPE that it throws includes an almost completely useless stacktrace and no other details.

We can do better: When calling `worker.run()`, we can catch the NPE and throw a wrapper exception that asks if you're using mocks and reminds you to do the right thing.

cc @afollestad

afollestad:
Per #842:
> In unit tests, if you use Mockito to create a Worker, the `run` method will return `null` even though the return type is non-nullable in Kotlin. Kotlin helps out with this by throwing an NPE before control is returned to any kotlin code, but the NPE that it throws includes an almost completely useless stacktrace and no other details.
>
> We can do better: When calling `worker.run()`, we can catch the NPE and throw a wrapper exception that asks if you're using mocks and reminds you to do the right thing.

Closes #842

### DIFF
--- a/kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/Workers.kt
+++ b/kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/Workers.kt
@@ -41,7 +41,7 @@ internal fun <T> CoroutineScope.launchWorker(
   workerDiagnosticId: Long,
   workflowDiagnosticId: Long,
   diagnosticListener: WorkflowDiagnosticListener?
-): ReceiveChannel<ValueOrDone<T>> = worker.run()
+): ReceiveChannel<ValueOrDone<T>> = worker.runWithNullCheck()
     .wireUpDebugger(workerDiagnosticId, workflowDiagnosticId, diagnosticListener)
     .transformToValueOrDone()
     .catch { e ->
@@ -55,6 +55,23 @@ internal fun <T> CoroutineScope.launchWorker(
     // operator is required to override the default buffer size.
     .buffer(RENDEZVOUS)
     .produceIn(this)
+
+/**
+ * In unit tests, if you use a mocking library to create a Worker, the run method will return null
+ * even though the return type is non-nullable in Kotlin. Kotlin helps out with this by throwing an
+ * NPE before before any kotlin code gets the null, but the NPE that it throws includes an almost
+ * completely useless stacktrace and no other details.
+ *
+ * This method does an explicit null check and throws an exception with a more helpful message.
+ *
+ * See [#842](https://github.com/square/workflow/issues/842).
+ */
+@Suppress("USELESS_ELVIS")
+private fun <T> Worker<T>.runWithNullCheck(): Flow<T> =
+  run() ?: throw NullPointerException(
+      "Worker $this returned a null Flow. " +
+          "If this is a test mock, make sure you mock the run() method!"
+  )
 
 @UseExperimental(VeryExperimentalWorkflow::class)
 private fun <T> Flow<T>.wireUpDebugger(
--- a/kotlin/workflow-runtime/src/test/java/com/squareup/workflow/internal/NullFlowWorker.java
+++ b/kotlin/workflow-runtime/src/test/java/com/squareup/workflow/internal/NullFlowWorker.java
@@ -1,0 +1,30 @@
+package com.squareup.workflow.internal;
+
+import com.squareup.workflow.Worker;
+import kotlinx.coroutines.flow.Flow;
+import org.jetbrains.annotations.NotNull;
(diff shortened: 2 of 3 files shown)


kotlin/.idea/misc.xml
```diff
@@ -13,10 +13,11 @@
       <item index="3" class="java.lang.String" itemvalue="org.openjdk.jmh.annotations.BenchmarkMode" />
     </list>
   </component>
+  <component name="ExternalStorageConfigurationManager" enabled="true" />
```
zach-klippenstein: What's this for?

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/Workers.kt
```diff
@@ -46,16 +47,35 @@ internal fun <T> CoroutineScope.launchWorker(
     .transformToValueOrDone()
     .catch { e ->
       // Workers that failed (as opposed to just cancelled) should have their failure reason
-      // re-thrown from the workflow runtime. If we don't unwrap the cause here, they'll just
-      // cause the runtime to cancel.
-      val cancellationCause = e.unwrapCancellationCause()
-      throw cancellationCause ?: e
+      // re-thrown from the workflow runtime.
+      throw e.wrapWorkerLaunchException(worker::class)
```
zach-klippenstein: The `worker.run()` call on line 44 is the one that we should wrap. Doing it here is confusing and doesn't communicate clearly the actual issue we're trying to handle.

Perhaps define a private extension like `worker.runWithNullCheck()` that can do the try/catch.

kotlin/.idea/misc.xml
afollestad: Damn, I thought I had removed this. Looks like I forgot to push.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/Workers.kt
afollestad: Looks like the crash actually happens within `transformToValueOrDone()`, where we `collect` from the Flow

zach-klippenstein: I thought the crash was caused by `run` returning null though - is that not true?

afollestad: Indeed - but it seems as though the crash isn't actually thrown until `collect` is used. The extension function on Flow must be causing the crash in the generated static method.

zach-klippenstein: It doesn't really matter where the crash currently happens, the point here is to try to catch the invalid behavior as soon as possible. The cancellation unwrap code is weird and already confusing, and adding more logic there, especially logic that is actually about a different part of the code altogether, just compounds the confusion. 

Checking for the null where it actually happens makes the intent of this code much clearer.

kotlin/.idea/misc.xml
zach-klippenstein: Fixed.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/Workers.kt
zach-klippenstein: I've implemented what I suggested and pushed this branch. Let me know what you think.

kotlin/workflow-runtime/src/test/java/com/squareup/workflow/internal/NullFlowWorker.java
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Worker that incorrectly returns null from {@link #run}, to simulate the default behavior of some
+ * mocking libraries.
+ *
+ * See <a href="https://github.com/square/workflow/issues/842">#842</a>.
+ */
+class NullFlowWorker implements Worker {
+  @NotNull @Override public Flow run() {
+    //noinspection ConstantConditions
+    return null;
```
afollestad: Oh nice - this setup is smart!

zach-klippenstein:
@afollestad I can't tag you as a review on this PR, if this looks good just let me know via comment and I'll merge.

afollestad:
@zach-klippenstein I like it, lgtm 👍 you're clearly the expert here though lol

PR #847, opened 2019-12-21
Update readme(square workflow droidcon talk video)

Johnett:
iframe not showing video thumbnail(github markdown doesn't support video embed videos). I have applied a closest solution.


zach-klippenstein:
It is unfortunate that GitHub markdown does not support it, however this code is intended to render the video inline on the docs site, square.github.io/workflow.

PR #850, opened 2019-12-24
Give worker coroutines descriptive names for debugging.

### LINKED ISSUE #848: Worker coroutines should be named
To help with debugging, coroutines launched to collect flows should include a `CoroutineName` context element with a name containing the worker's toString and its key.

This will incur some performance overhead when starting a worker, so we could put this behavior behind the same debug flags that the coroutine library uses, but I think it should be fine to just do it unconditionally for now and worry about optimizing later.

zach-klippenstein:
Also fills out the worker docs some more.

Closes #848.

### DIFF
--- a/kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/Workers.kt
+++ b/kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/Workers.kt
@@ -18,6 +18,7 @@ package com.squareup.workflow.internal
 import com.squareup.workflow.VeryExperimentalWorkflow
 import com.squareup.workflow.Worker
 import com.squareup.workflow.diagnostic.WorkflowDiagnosticListener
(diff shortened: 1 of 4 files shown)


kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/Workers.kt
```diff
@@ -124,3 +127,11 @@ private fun <T> Flow<T>.transformToValueOrDone(): Flow<ValueOrDone<T>> = flow {
   }
   emit(ValueOrDone.done())
 }
+
+private fun CoroutineScope.createWorkerScope(
+  worker: Worker<*>,
+  key: String
+): CoroutineScope = this + CoroutineName(worker.debugName(key))
```
afollestad: Whoa, TIL

PR #851, opened 2019-12-24
Make all workers run on the Unconfined dispatcher.

zach-klippenstein:
From the kdoc:

> The dispatcher is always set to `Unconfined` to
>    *   minimize overhead for workers that don't care which thread they're executed on (e.g. logging
>    *   side effects, workers that wrap third-party reactive libraries, etc.). If your work cares
>    *   which thread it runs on, use `withContext` or
>    *   `flowOn` to specify a dispatcher.

Previously, the workers would run on the same dispatcher as the workflow runtime (`Main.immediate`), which meant that if the worker was wrapping, e.g., an `Observable` that was emitting on a background thread, the coroutine would dispatch back onto the main thread before sending to the channel, which is completely unnecessary, since channels are already thread-safe.

This doesn't change the threading behavior for simple side-effect-only workers – since they're started from the workflow runtime, they will be dispatched onto the nested event loop that `Main.immediate` creates for "immediate" dispatches.

### DIFF
--- a/kotlin/workflow-core/src/main/java/com/squareup/workflow/Worker.kt
+++ b/kotlin/workflow-core/src/main/java/com/squareup/workflow/Worker.kt
@@ -123,11 +123,17 @@ interface Worker<out OutputT> {
    *
    * ## Coroutine Context
    *
-   * When a worker is started, a coroutine is launched to [collect][Flow.collect] the flow. This
-   * coroutine is launched in the same scope as the workflow runtime, with the addition a
-   * [CoroutineName][kotlinx.coroutines.CoroutineName] that describes the `Worker` instance
-   * (via `toString`) and the key specified by the workflow running the worker. When the worker is
-   * torn down, the coroutine is cancelled.
+   * When a worker is started, a coroutine is launched to [collect][Flow.collect] the flow.
+   * When the worker is torn down, the coroutine is cancelled.
+   * This coroutine is launched in the same scope as the workflow runtime, with a few changes:
+   *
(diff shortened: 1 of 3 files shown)


kotlin/workflow-core/src/main/java/com/squareup/workflow/Worker.kt
```diff
@@ -124,10 +124,16 @@ interface Worker<out OutputT> {
...
-   * torn down, the coroutine is cancelled.
+   * coroutine is launched in the same scope as the workflow runtime, with a few changes:
+   *
+   * - The dispatcher is always set to [Unconfined][kotlinx.coroutines.Dispatchers.Unconfined] to
+   *   minimize overhead for workers that don't care which thread they're executed on (e.g. logging
+   *   side effects, workers that wrap third-party reactive libraries, etc.). If your work cares
+   *   which thread it runs on, use [withContext][kotlinx.coroutines.withContext] or
+   *   [flowOn][kotlinx.coroutines.flow.flowOn] to specify a dispatcher.
+   * - A [CoroutineName][kotlinx.coroutines.CoroutineName] that describes the `Worker` instance
+   *   (via `toString`) and the key specified by the workflow running the worker. When the worker is
```
rjrjr: It's weird that the last sentence about teardown is in the bullet about the `CoroutineName.`

zach-klippenstein: oops

Issue #854, opened 2020-01-02
Gradle 6.0.1 and doesn't work with Java 8.0.232


SalvatoreT:
When trying to build from the command line (or Android Studio), I run into the following issue.

# Gradle 6.0.1
This is using Java 8.0.232.
 
```
10:48 $ ./gradlew :samples:todo-android:app:assembleDebug

FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring project ':workflow-ui-android'.
> Failed to notify project evaluation listener.
   > wrong number of arguments
   > KotlinJvmAndroidCompilation with name 'debugAndroidTest' not found.
   > Cannot access last() element from an empty List

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

Deprecated Gradle features were used in this build, making it incompatible with Gradle 7.0.
Use '--warning-mode all' to show the individual deprecation warnings.
See https://docs.gradle.org/6.0.1/userguide/command_line_interface.html#sec:command_line_warnings

BUILD FAILED in 891ms
```

# Gradle 5.6.3
I don't have the same problem when I revert commit c5ad4ed9abd5461e0fa3769f0a086249a8e606ec.

```
10:49 $ ./gradlew :samples:todo-android:app:assembleDebug
Starting a Gradle Daemon, 1 busy and 6 stopped Daemons could not be reused, use --status for details

> Configure project :workflow-ui-android
WARNING: API 'variant.getJavaCompile()' is obsolete and has been replaced with 'variant.getJavaCompileProvider()'.
It will be removed at the end of 2019.
For more information, see https://d.android.com/r/tools/task-configuration-avoidance.
To determine what is calling variant.getJavaCompile(), use -Pandroid.debug.obsoleteApi=true on the command line to display more information.

Deprecated Gradle features were used in this build, making it incompatible with Gradle 6.0.
Use '--warning-mode all' to show the individual deprecation warnings.
See https://docs.gradle.org/5.6.3/userguide/command_line_interface.html#sec:command_line_warnings

BUILD SUCCESSFUL in 15s
91 actionable tasks: 8 executed, 83 up-to-date
```

SalvatoreT:
Kirill helped me debug in the Slack channel. The tl;dr is that AdoptOpenJDK's Oracle HotSpot implementation works but OpenJDK's OpenJ9 implementation does not.

[...]

PR #855, opened 2020-01-02
Revert "Upgrade Gradle to 6.0.1."

SalvatoreT:
This is one possible solution to #854.


SalvatoreT:
We found out the root cause was the implementation of Java and not the version of Gradle. See #854.

PR #857, opened 2020-01-02
Bump the Gradle wrapper for 6.0.1

SalvatoreT:
Zach updated the `gradle-wrapper.properties` file in c5ad4ed9abd5461e0fa3769f0a086249a8e606ec, but some of the other files need to change as well.

I ran the following command to get these auto-generated updates.

```
./gradlew wrapper --gradle-version=6.0.1 --distribution-type=all
```


kotlin/gradlew
```diff
@@ -175,14 +175,9 @@ save () {
...
-APP_ARGS=$(save "$@")
+APP_ARGS=`save "$@"`
 
 # Collect all arguments for the java command, following the shell quoting and substitution rules
 eval set -- $DEFAULT_JVM_OPTS $JAVA_OPTS $GRADLE_OPTS "\"-Dorg.gradle.appname=$APP_BASE_NAME\"" -classpath "\"$CLASSPATH\"" org.gradle.wrapper.GradleWrapperMain "$APP_ARGS"
 
-# by default we should be in the correct project dir, but when run from Finder on Mac, the cwd is wrong
-if [ "$(uname)" = "Darwin" ] && [ "$HOME" = "$PWD" ]; then
-  cd "$(dirname "$0")"
-fi
```
kirillzh: This looks like our own custom setup that we should keep. @zach-klippenstein, @rjrjr can you confirm?

zach-klippenstein: I don't believe so, we should be able to remove this.

PR #863, opened 2020-01-07
Optimize re-rendering workers.

zach-klippenstein:
Based on the work in #800, and applies that same algorithm change to workers. Factors out some
of the logic of managing the active/staging lists into the `ActiveStagingList` class, which effectively
replaces `SubtreeManager` with a much less-weird API.

Change to tracing diagnostic listener's golden value reflects the fact that workers are now started before
the render method returns.


rjrjr:
> Change to tracing diagnostic listener's golden value reflects the fact that workers are now started before the render method returns.

I like the sound of that change in its own right. How synchronous is it, though? Could it lead to a recursive `render` pass before the current one finishes? Is that okay?

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/WorkerChildNode.kt
```diff
@@ -0,0 +1,66 @@
...
+ */
+package com.squareup.workflow.internal
+
+import com.squareup.workflow.Worker
+import com.squareup.workflow.WorkflowAction
+import com.squareup.workflow.internal.InlineLinkedList.InlineListNode
+import kotlinx.coroutines.channels.ReceiveChannel
+
+/**
+ * TODO kdoc
```
zach-klippenstein: @zach-klippenstein TODO

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/ActiveStagingList.kt
```diff
@@ -0,0 +1,64 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.workflow.internal
+
+import com.squareup.workflow.internal.InlineLinkedList.InlineListNode
+
+/**
+ * TODO write documentation
```
zach-klippenstein: @zach-klippenstein TODO

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/WorkerChildNode.kt
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * Holds the channel representing the outputs of a worker, as well as a tombstone flag that is
+ * true after the worker has finished and we've reported that fact to the workflow. This is to
+ * prevent the workflow from entering an infinite loop of getting `Finished` events if it
+ * continues to listen to the worker after it finishes.
+ *
+ * @param channel TODO
+ * @param tombstone TODO
+ */
+/* ktlint-disable parameter-list-wrapping */
```
rjrjr: You're not wrapping the param list.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/WorkflowNode.kt
```diff
@@ -157,6 +129,26 @@ internal class WorkflowNode<PropsT, StateT, OutputT : Any, RenderingT>(
...
 
+  override fun <T> runningWorker(
+    worker: Worker<T>,
+    key: String,
+    handler: (T) -> WorkflowAction<StateT, OutputT>
+  ) {
+    // Prevent duplicate workflows with the same key.
+    workers.staging.forEach {
+      require(!(it.matches(worker, key))) {
+        "Expected keys to be unique for $worker: key=$key"
```
rjrjr: Helpful to include the toString of both the new worker and the conflicting one in such a message.

kotlin/workflow-runtime/src/test/java/com/squareup/workflow/internal/ActiveStagingListTest.kt
```diff
@@ -0,0 +1,20 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.workflow.internal
+
+class ActiveStagingListTest {
+  // TODO write tests
```
rjrjr: TODO

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/WorkerChildNode.kt
zach-klippenstein: Not what ktlint thinks – it fails without this. It's because of the weird indentation due to the generics.

kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/WorkflowNode.kt
zach-klippenstein: It's there: `$worker`

zach-klippenstein:
> How synchronous is it, though? Could it lead to a recursive render pass before the current one finishes? Is that okay?

Just saw this – no, it can't recurse. It's just creating the coroutine + channel slightly earlier.

PR #867, opened 2020-01-08
Snips vestigial key refs in Worker.kt

rjrjr:


rjrjr:
Oh good, Travis is failing.

PR #868, opened 2020-01-10
Adds ContainerHints to ShowRenderingTag.

rjrjr:
We're already storing the most recent rendering in view tags. It's weird that we don't do so for the ContainerHints that are always provided at render time, and the lack is actually getting in my way as I build custom containers.

(Note additional commit to bump the version number.)

### DIFF
--- a/kotlin/gradle.properties
+++ b/kotlin/gradle.properties
@@ -21,7 +21,7 @@ android.useAndroidX=true
 #android.debug.obsoleteApi=true
 
 GROUP=com.squareup.workflow
-VERSION_NAME=0.22.2
+VERSION_NAME=0.22.3-SNAPSHOT
 
 POM_DESCRIPTION=Reactive workflows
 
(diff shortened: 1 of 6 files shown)


kotlin/gradle.properties
```diff
@@ -21,7 +21,7 @@ android.useAndroidX=true
 #android.debug.obsoleteApi=true
 
 GROUP=com.squareup.workflow
-VERSION_NAME=0.22.2
+VERSION_NAME=0.22.3-SNAPSHOT
```
zach-klippenstein: Are the iOS versions out of date too?

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ViewShowRendering.kt
```diff
@@ -101,20 +105,25 @@ fun <RenderingT : Any> View.showRendering(
  * has never been called.
  */
 fun <RenderingT : Any> View.getRendering(): RenderingT? {
+  // Can't use a val because of the parameter type.
   @Suppress("UNCHECKED_CAST")
   return when (val showing = showRenderingTag?.showing) {
     null -> null
     else -> showing as RenderingT
```
kirillzh: Is this the same as `showRenderingTag?.showing as?`?

kotlin/gradle.properties
rjrjr: They sure are.

rjrjr: Fixed.

PR #872, opened 2020-01-14
Flag out WorkflowTesting APIs when in Release mode

dhavalshreyas:
**Issue:**
`pod trunk push WorkflowTesting.podspec` fails during validation.

**Reason:**
`Workflow` is built with `ENABLE_TESTABILITY=NO` for `Release` builds, causing `@testable import WorkflowTesting` to fail compilation in `Release`. 

**Proposal:**
`WorkflowTesting` is meant to be used for Unit Testing only. Flag out the relevant bits when not in `Debug` mode.

### DIFF
--- a/swift/WorkflowTesting/Sources/WorkflowActionTester.swift
+++ b/swift/WorkflowTesting/Sources/WorkflowActionTester.swift
@@ -13,6 +13,12 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
+// WorkflowTesting only available in Debug mode.
+//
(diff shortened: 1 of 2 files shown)


AquaGeek:
Have we verified this against a sample app?

dhavalshreyas:
> Have we verified this against a sample app?

@AquaGeek: Yes, I did try this with a few sample apps.

@JustinDSN  brought up a good point, we don't have `WorkflowTesting-Unit-Tests` in the test matrix. I'll add it on my todos to add something there.

PR #873, opened 2020-01-14
Add back button support to Dungeon sample game screen.

zach-klippenstein:
Based on #874.


kotlin/samples/dungeon/app/src/main/java/com/squareup/sample/dungeon/GameLayoutRunner.kt
```diff
@@ -52,6 +53,8 @@ class GameLayoutRunner(view: View) : LayoutRunner<GameRendering> {
     rendering: GameRendering,
     containerHints: ContainerHints
   ) {
+    view.backPressedHandler = rendering.onBackPressed
```
zach-klippenstein: This is currently broken by #875.

PR #876, opened 2020-01-15
WIP Proposal for new WorkflowDebugger that allows pausing and inspecting the runtime.

zach-klippenstein:
This preserves the `WorkflowDiagnosticListener` functionality, but moves it around a bit.

 ## Initialization

Instead of setting a listener on the `WorkflowSession`, you call `WorkflowSession.debug` and pass an "initializer" function. This initializer replaces the `WorkflowDiagnosticListener.onRuntimeStarted` method, and receives one additional parameter: a `WorkflowDebugger`.

## Diagnostic Events

The `WorkflowDiagnosticListener` interface still exists, but instead of passing it to the session directly, the initializer function registers it on the `WorkflowDebugger` by calling `addDiagnosticListener`.

## Scoping

Since each `WorkflowDiagnosticListener` is now associated with a specific workflow runtime's execution, it is much more natural to write stateful listeners that need to be specialized to a specific runtime session, since the listener can be instantiated at the time the particular runtime session starts. 

## Composition

It used to be that you could compose multiple `WorkflowDiagnosticListener`s with the `andThen` extension function. With this proposal, composition happens at the "initializer" level: instead of composing listeners together, you compose initializers. This is more powerful, since each initializer can still add its own diagnostic listener(s), but they can also interact with the debugger. An initializer can delegate to other initializer functions as necessary.

## Debugging

In addition to accepting diagnostic listeners, the `WorkflowDebugger` lets you actually do interactive debugging.

### Inspecting the Workflow tree

It provides a stream of `WorkflowHierarchyDebugSnapshot` values, emitted after every render pass. A `WorkflowHierarchyDebugSnapshot` is an immutable description of an entire tree, including every workflow's props, state, and rendering, and all its child workflows and workers. These are now generated by the `WorkflowNode`s directly, instead of being reconstructed from diagnostic events by `DebugSnapshotRecordingListener`. This emission can be enabled or disabled by setting a property on the debugger.

### Pausing and resuming the runtime

It also lets you set a "breakpoint" after every render pass, via another debugger property. When this breakpoint is enabled, the runtime will pause (suspend) after emitting every render pass rendering and debug snapshot. In order to resume processing, the debugger's `resume` method must be called after pausing (clearing the flag does not resume). Whenever the breakpoint is enabled, debug snapshots are also always emitted, so one could implement a debugging UI that displays the tree snapshot and provides a "Resume" button.

The breakpoint functionality is implemented by simply suspending the main workflow loop at the breakpoint. Workers will still run while the runtime is paused, but they will experience backpressure. Rendering events will be queued.


kotlin/workflow-runtime/src/main/java/com/squareup/workflow/debugging/RealWorkflowDebugger.kt
```diff
@@ -0,0 +1,90 @@
...
+  @Volatile override var emitTrees: Boolean = false
+
+  @Volatile override var breakpointEnabled: Boolean = false
+
+  var diagnosticListener: WorkflowDiagnosticListener? = null
+    private set
+
+  private var isRunning = false
+
+  private val resumeSignal = Channel<Unit>(capacity = 1)
```
Kiskae: Wouldn't a `RendezvousChannel` be appropriate here since the usage should exclusive be "one runtimewaits, then one debugger resumes". It would also allow using the return value of `offer` to indicate whether the debugger did actually resume.

zach-klippenstein: It's got a buffer to avoid a race condition in the scenario where a collector of `trees` calls `resume` before we actually start receiving on the resume signal. In that case, the resume call would get dropped. The API is less complicated to reason about this way, because by the time the tree is emitted, any `resume` call will actually resume the runtime. Do you have a use case for `resume` returning whether it actually succeeded or not? I could see it maybe being useful to expose a property indicating whether we're paused or not, but that would be exposed as its own Flow, not a return value from `resume`.

zach-klippenstein:
Closing because inactive, might reopen later.

PR #880, opened 2020-01-15
Fix racy broken WorkerStressTests.

### LINKED ISSUE #878: WorkerStressTest seems to be broken
Running `./gradlew build` locally will hang at `WorkerStressTest`.

I suspect this might be why all our PRs in the last few days have been timing out as well.

I noticed this first on my Kotlin 1.3.70-eap branch, and assumed it had something to do with the new version, but this is happening on master. The only coroutines-related thing that I can think of that changed recently was making workers launch on the `Unconfined` dispatcher, but that PR went green.

zach-klippenstein:
Fixes #878.

### DIFF
--- a/kotlin/workflow-testing/src/test/java/com/squareup/workflow/WorkerStressTest.kt
+++ b/kotlin/workflow-testing/src/test/java/com/squareup/workflow/WorkerStressTest.kt
@@ -1,23 +1,25 @@
 package com.squareup.workflow
 
-import com.squareup.workflow.WorkflowAction.Companion.noAction
 import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.FlowPreview
 import kotlinx.coroutines.cancelChildren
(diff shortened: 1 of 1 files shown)


kotlin/workflow-testing/src/test/java/com/squareup/workflow/WorkerStressTest.kt
```diff
@@ -29,31 +31,36 @@ class WorkerStressTest {
...
+      val outputs = launchWorkflowIn(this, workflow, flowOf(Unit)) {
+        // Collect to a channel immediately so that we don't miss any outputs.
+        it.outputs.produceIn(this)
+      }
 
       // This should just work, and the test will finish, but this is broken by
       // https://github.com/Kotlin/kotlinx.coroutines/issues/1584 and will crash instead if
       // receiveOrClosed is used.
       channel.close()
-      delay(1)
```
zach-klippenstein: Not sure why this delay was ever here. This test should never have passed.

PR #881, opened 2020-01-15
Allow empty snapshots to be passed into launchWorkflowIn.

### LINKED ISSUE #879: EOFException thrown when starting workflow with WorkflowRunner
## Problem

We've seen some internal reports from apps in the wild crashing with `EOFException` on launch.

The crash happens because `restoreState` is trying to read from an empty `Snapshot`. Empty snapshots should not be passed to the runtime, only null. 

I think this has something to do with the `WorkflowRunnerViewModel` defaulting its snapshot cache to `Snapshot.EMPTY`, and always writing that property to the `SavedStateRegistry`. It should default to null instead. I'm not sure how to reproduce yet, since this property should be updated as soon as the runtime starts – seems like it should only happen if the app is killed immediately, before the runtime actually starts. It could be that the system isn't calling `SavedStateProvider.saveState` often enough.

## Solution

There are two possible fixes:

1. Make `WorkflowRunnerViewModel` default to a null snapshot instead of empty, or don't write the snapshot to the bundle if it's empty.
2. Make the runtime more resilient, and treat empty snapshots as null ones, ignoring them.

cc @Blisse

zach-klippenstein:
Fixes #879.

cc @Blisse

### DIFF
--- a/kotlin/workflow-runtime/src/main/java/com/squareup/workflow/LaunchWorkflow.kt
+++ b/kotlin/workflow-runtime/src/main/java/com/squareup/workflow/LaunchWorkflow.kt
@@ -85,7 +85,7 @@ internal typealias Configurator <O, R, T> = CoroutineScope.(
  * will fail (report an exception up through [scope]). If this flow completes _after_ emitting at
  * least one value, the runtime will _not_ fail or stop, it will continue running with the
  * last-emitted input.
- * @param initialSnapshot If not null, used to restore the workflow.
+ * @param initialSnapshot If not null or empty, used to restore the workflow.
  * @param beforeStart Called exactly once with the flows for renderings/snapshots and outputs.
  * It also gets a sub-scope of [scope] with a newly created child [Job] which defines the lifetime
  * of the launched workflow tree. Cancelling that job ends the new workflow session.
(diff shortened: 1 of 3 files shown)


JustinDSN:
@zach-klippenstein are there any other additional edge cases we should protect from when trying to restore from the initialSnapshot? For example: The snapshot contains bytes, but can't be inflated / de-serialized. In addition to the fact that the Snapshot's size is 0?

rjrjr:
> are there any other additional edge cases we should protect from when trying to restore from the initialSnapshot? For example: The snapshot contains bytes, but can't be inflated / de-serialized. In addition to the fact that the Snapshot's size is 0?

We generally prefer to fail fast for that kind of thing -- expose the bug and get it fixed rather than being forgiving and make it hard to discover. I'd rather crash if I break serialization — especially since our main use case, the Android hibernation system, ensures that we won't get stuck in a crash loop.

In this specific case, the distinction between empty and null is way too subtle for us to expect devs to learn.

zach-klippenstein:
What Ray said. I intentionally kept things as strict as possible for that reason, so we could detect invalid usage as soon as possible. But empty snapshots are so easy to get wrong, and it's so obvious what the intent of passing an empty snapshot is, that it doesn't really make sense to be so strict in this case.

JustinDSN:
Sounds good!

PR #883, opened 2020-01-15
Add ReactiveSwift FAQ entry

rjrjr:


docs/faq.md
```diff
@@ -49,3 +49,9 @@ MVC.
 
 Workflow defines strong boundaries and contracts between separate parts of the application to ensure
 that our code remains predictable and maintainable as the size and complexity of the codebase grows.
+
+## Why ReactiveSwift instead of RxSwift?
+
+Swift has a powerful type system. ReactiveSwift was designed to take advantage of the language – and, as it turns out, that resulted in many common design decisions between ReactiveSwift and Combine (which is/will become the defacto reactive architecture on the platform).
+
+RxSwift, on the other hand, was designed to be similar to rx on other platforms above all else. It lacks typed errors, has weird convention breaking design patterns, etc.
```
zach-klippenstein: This is good, would be nice to elaborate a bit more on "weird convention-breaking design patterns" and stuff though.

zach-klippenstein:
@dhavalshreyas or @AquaGeek, is this worth fixing up and merging?

docs/faq.md
dhavalshreyas: I would personally change this to something like:

```
The team considered various options here (including `RxSwift`) and decided to go with `ReactiveSwift` due to it's support for `hot` vs `cold` observables and typed errors.

However, we want to support `Workflow` for projects which use other reactive frameworks. That work is being tracked here: #933
```

zach-klippenstein: Do you want to take over this PR?

zach-klippenstein:
This PR has been open and inactive for a while, closing. Feel free to re-open if someone wants to actually land it.

PR #886, opened 2020-01-17
Improved animation for BackStackContainer, PanelContainer.

rjrjr:
@zach-klippenstein demonstrates that dialog based modals can work just fine with tab / IME navigation, and with very simple code. And fixes mismatched speed on default `BackStackContainer` transition.

phone
tablet

### DIFF
--- a/kotlin/samples/containers/android/src/main/java/com/squareup/sample/container/panel/PanelContainer.kt
+++ b/kotlin/samples/containers/android/src/main/java/com/squareup/sample/container/panel/PanelContainer.kt
@@ -39,6 +39,8 @@ import kotlin.math.min
 object PanelContainer : ViewBinding<PanelContainerScreen<*, *>>
 by ModalContainer.forContainerScreen(
     R.id.panel_container,
+    // This theme defines custom enter and exit animation styles for panel windows.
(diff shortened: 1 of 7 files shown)


kotlin/samples/containers/android/src/main/java/com/squareup/sample/container/panel/PanelContainer.kt
```diff
@@ -39,6 +39,7 @@ import kotlin.math.min
 object PanelContainer : ViewBinding<PanelContainerScreen<*, *>>
 by ModalContainer.forContainerScreen(
     R.id.panel_container,
+    dialogThemeResId = R.style.PanelDialog,
```
zach-klippenstein: Add a comment about why this is needed?

rjrjr:
```kotlin
    // This theme defines custom enter and exit animation styles for panel windows.
```

And added this in `style.xml`.

```
    <!-- For some reason this is required to get our windowEnterAnimation honored. -->
    <item name="android:windowIsFloating">true</item>
```

PR #803, opened 2019-12-07
WIP Optimize rendering by caching renderings when possible.

zach-klippenstein:
[Spreadsheet](https://docs.google.com/spreadsheets/d/1e4OBW0RmKscQyQPIYMLHf0Tzj2WbBnOndVLnEcJ5MAg/edit#gid=916243004)

Note below the `renderPassAlternateOneLeaf | SQUARE` case – this is the most similar case to most real-world ones. A tree that is a little deep and a little bushy, where only one path some node up to the root actually updates and needs re-rendering. Here we only render 4 nodes, and all other renderings are simply returned from the cache.

Benchmark | (treeShape) | Mode | Cnt | Score (before) | Error | Score (after) | Error | Units | Speedup
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
WorkflowNodeBenchmark.renderPassAlternateLeaves | DEEP | avgt | 3 | 310.463 | 7.559 | 311.437 | 83.952 | us/op | 1.00
WorkflowNodeBenchmark.renderPassAlternateLeaves | BUSHY | avgt | 3 | 563.374 | 50.840 | 556.091 | 17.334 | us/op | 1.01
WorkflowNodeBenchmark.renderPassAlternateLeaves | SQUARE | avgt | 3 | 504.244 | 8.309 | 538.211 | 230.249 | us/op | 0.94
WorkflowNodeBenchmark.renderPassAlternateOneLeaf | DEEP | avgt | 3 | 315.172 | 18.100 | 312.197 | 11.376 | us/op | 1.01
WorkflowNodeBenchmark.renderPassAlternateOneLeaf | BUSHY | avgt | 3 | 294.026 | 22.039 | 178.340 | 12.926 | us/op | 1.65
WorkflowNodeBenchmark.renderPassAlternateOneLeaf | SQUARE | avgt | 3 | 273.824 | 5.861 | 10.249 | 0.794 | us/op | 26.72
WorkflowNodeBenchmark.renderPassAlternateWorkers | DEEP | avgt | 3 | 323.121 | 20.304 | 327.485 | 10.509 | us/op | 0.99
WorkflowNodeBenchmark.renderPassAlternateWorkers | BUSHY | avgt | 3 | 556.811 | 80.929 | 552.366 | 22.018 | us/op | 1.01
WorkflowNodeBenchmark.renderPassAlternateWorkers | SQUARE | avgt | 3 | 502.989 | 7.628 | 508.815 | 31.266 | us/op | 0.99
[...]

PR #887, opened 2020-01-19
Don't onPropsChanged unless the old and new props are actually unequal.

zach-klippenstein:
This looks like an optimization, but since most `onPropsChanged` are no-ops,
it's not really. This is to prevent code from starting to rely on the fact
that `onPropsChanged` is currently called on ever render pass. This is a small
step towards the behavior in #803, which lets `WorkflowNode`s skip a render
pass for their entire subtree when unnecessary. Making this change now will
help prevent code from being written that will break when that change lands.

The actual change is just surrounding three lines with an if, most of the changes in this PR are updates to tests.


zach-klippenstein:
Should manually test the Dungeon sample and make sure the time machine still works.

kotlin/workflow-runtime/src/test/java/com/squareup/workflow/internal/WorkflowDiagnosticListenerIntegrationTest.kt
```diff
@@ -235,6 +221,21 @@ class WorkflowDiagnosticListenerIntegrationTest {
...
+              "onBeforeWorkflowRendered",
+              "onAfterWorkflowRendered",
+              "onAfterRenderPass",
+              "onBeforeSnapshotPass",
+              "onAfterSnapshotPass"
+          ), listener.consumeEventNames()
+      )
+
+      propsChannel.send("new props")
+      yield()
```
rjrjr: Maybe doc these yields?

zach-klippenstein: Will follow up.

PR #889, opened 2020-01-19
Upgrade Dokka to 0.10.0 and maven publish to 0.9.0-SNAPSHOT.

### LINKED ISSUE #569: Sources aren't being published for AARs
Non-android artifacts have sources, Android artifacts seem to be missing them.

### LINKED ISSUE #651: Upgrade to Dokka 0.10.0
Dokka 0.10.0 was just [released](https://github.com/Kotlin/dokka/releases/tag/0.10.0), fixes a bunch of bugs and introduces multiplatform support.

zach-klippenstein:
Closes #651.

This technically fixes #569 as well (according to https://github.com/vanniktech/gradle-maven-publish-plugin/issues/54),
and after uploading the artifact server has the sources jar with the actual sources inside. However, when
consuming via gradle, IntelliJ still isn't showing the sources for AARs (it is for JARs). The android artifacts
have a `-api` suffix however, so I'm wondering if the source jars are getting lost in some transformation
(maybe Jetifier?).

Note that the javadocs jars are on the server but they're all empty (for android and non-android
artifacts), but IntelliJ seems to be just fine pulling the kdoc from the source jars, so I guess the source jars
are all that is required for documentation.

### DIFF
--- a/kotlin/.buildscript/configure-dokka.gradle
+++ b/kotlin/.buildscript/configure-dokka.gradle
@@ -27,15 +27,17 @@ afterEvaluate { project ->
   project.tasks.dokka {
     outputDirectory = "$rootDir/../docs/kotlin/api"
     outputFormat = 'gfm'
-    reportUndocumented = false
-    skipDeprecated = true
-    jdkVersion = 8
-    packageOptions {
-      prefix = "com.squareup.workflow.internal"
-      suppress = true
-    }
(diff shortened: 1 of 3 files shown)


zach-klippenstein:
UI tests failed, but they pass locally. They all failed, so I'm hoping this was a transient problem on Travis' side.

PR #896, opened 2020-01-23
Increase worker counts in WorkerStressTest to 1000

zach-klippenstein:


zach-klippenstein:
I tried going up to 5000, 500000, and we actually start to see terrible performance pretty quickly since we go through the entire list of rendered workers on every worker (`n!`). To solve this, we need Workers to explicitly expose their type so we can use hashing to detect collisions instead of linear search.

Issue #402, opened 2019-06-10
Lost events from subscriptions if multiple fire at the same time


davidapgar:
Prior to the concurrency changes in 0.17, if two signals fired at the same time (same runloop), both actions would be received (across two render passes, assuming the subscription was still maintained).

With the most recent changes, only the first subscription's action is received, and the second is silently dropped. This appears to be due to how subscription/signal handling was changed. Prior, it was done through a signal of signals, that were flatMap'd with "latest" and with a constant observation on the output of events. This changed to have the signal subscribe on each pass (after being merged together).

The behavior, at a minimum, is now surprising.

Either
1) Restore the previous behavior (to not drop events)
or
2) assert if multiple signals fire on the same render loop.

If restoring the previous behavior, we must assure that it's the updated action handling - there may have previously been a bug where the second signal firing was handled with the rules of the previous render loop - and thus might have caused an action that wasn't declared (need to investigate this).

davidapgar:
With an overly specific test case, the previous behavior was a "bug". Example:

Two signals being subscribed to, signalA and signalB, and a state where if we're not "loading" we don't want to subscribe anymore, eg:
```
enum State {
  case idle
  case loading
}

enum Action {
  case signalFired
  apply() {
    switch self {
      case .signalFired:
        state = .idle
    }
  }
}
render() {
  if state == .loading {
    context.subscribe(signalA.map { return .signalFired })
    context.subscribe(signalB.map { return .signalFired })
  }  
}
```

Both `signalA` and `signalB` fire at the same time (same runloop).
The action from `signalA` is received, the state updates to `.idle`. Render happens again, and the signals are *not* subscribed to.

`signalB` will still fire with the previous implementation, and the `.signalFired` action will be handled. This is unexpected behavior, as that signal was not subscribed to after `signalA` fired.

davidapgar:
Talked with @zach-klippenstein about how this is handled on the Kotlin side:
(If i understood correctly) Since they are using Channels, events from signals are queued since Channels act as a blocking queue. The mapping from a signal is done after the event is received from a channel.

Thus, the behavior in the example above would be:
`signalA` and `signalB` both send events at the same time. The whichever one "wins" (in this example, signalA) gets handled first. The event is mapped to an action and handled.
A render pass happens, and signalB is still subscribed to. Its event is received from the channel, and mapped to the action from the most recent render pass and handled as well.

To add this behavior, we need to:
[...]

PR #907, opened 2020-01-23
[Exploration] Use a run loop observer to trigger render passes asynchronously

AquaGeek:
This changes the render pass to be asynchronous instead of synchronous and will happen **at most** once per run loop. There are likely issues with how this is set up, but I wanted to get some eyes on it sooner rather than later.

I _believe_ this addresses #402.

### DIFF
--- a/swift/Workflow/Sources/RunLoopObserver.swift
+++ b/swift/Workflow/Sources/RunLoopObserver.swift
@@ -18,8 +18,6 @@ import CoreFoundation
 
 /// Swift wrapper for `CFRunLoopObserver`
 internal final class RunLoopObserver {
-    private let runLoop: CFRunLoop
-    private let runLoopModes: CFRunLoopMode
     private var observer: CFRunLoopObserver!
 
     /// Creates a `RunLoopObserver` and adds it to the given run loop for the given run loop modes. See the docs for `CFRunLoopObserverCreateWithHandler` for documentation.
@@ -31,8 +29,6 @@ internal final class RunLoopObserver {
         runLoopModes: CFRunLoopMode = .defaultMode,
         callback: @escaping (_ activityStage: CFRunLoopActivity) -> Void
     ) {
-        self.runLoop = runLoop
-        self.runLoopModes = runLoopModes
         observer = CFRunLoopObserverCreateWithHandler(
             kCFAllocatorDefault,
             activityStages.rawValue,
@@ -47,6 +43,6 @@ internal final class RunLoopObserver {
 
     deinit {
         // Clean up the observer
-        CFRunLoopRemoveObserver(runLoop, observer, runLoopModes)
+        CFRunLoopObserverInvalidate(observer)
     }
 }
--- a/swift/Workflow/Sources/RunLoopObserver.swift
+++ b/swift/Workflow/Sources/RunLoopObserver.swift
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2020 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import CoreFoundation
+
+/// Swift wrapper for `CFRunLoopObserver`
+internal final class RunLoopObserver {
+    private let runLoop: CFRunLoop
+    private let runLoopModes: CFRunLoopMode
+    private var observer: CFRunLoopObserver!
+
+    /// Creates a `RunLoopObserver` and adds it to the given run loop for the given run loop modes. See the docs for `CFRunLoopObserverCreateWithHandler` for documentation.
+    init(
+        runLoop: CFRunLoop = CFRunLoopGetCurrent(),
+        activityStages: CFRunLoopActivity,
+        repeats: Bool = true,
+        order: CFIndex = 0,
+        runLoopModes: CFRunLoopMode = .defaultMode,
+        callback: @escaping (_ activityStage: CFRunLoopActivity) -> Void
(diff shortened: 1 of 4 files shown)


swift/Workflow/Sources/WorkflowHost.swift
```diff
@@ -65,6 +69,23 @@ public final class WorkflowHost<WorkflowType: Workflow> {
             self?.handle(output: output)
         }
 
+        runLoopObserver = RunLoopObserver(activityStages: .beforeWaiting, runLoopModes: .commonModes) { [weak self] activity in
+            self?.renderIfNeeded()
+        }
+    }
+
+    public func setNeedsRender() {
```
timdonnelly: Why is this public?

AquaGeek: 🤷‍♀ Maybe you need to poke it?

It should likely start off `internal` or even `private`.

swift/Workflow/Sources/RunLoopObserver.swift
```diff
@@ -0,0 +1,48 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import CoreFoundation
+
+/// Swift wrapper for `CFRunLoopObserver`
+internal final class RunLoopObserver {
+    private var observer: CFRunLoopObserver!
```
bencochran: nit: I think this can be a `let` without the `!` since nothing in the initializer needs `self`?

swift/Workflow/Sources/WorkflowHost.swift
bencochran: Yeah, I don’t think there’s a scenario where the outside world should ever need to force a render (without, say, an `update(workflow:)` call)

swift/Workflow/Sources/RunLoopObserver.swift
AquaGeek: Oh, yes, it can probably be a `let`. Sorry, this went through a few revisions. The `CFRunLoopObserverCreateWithHandler` returns an implicitly unwrapped optional, so we need to handle that somewhere (maybe this should be a `let observer: CFRunLoopObserver?`).

bencochran: Yeah, that was my guess.

Looks like that function doesn’t have nullability annotations. I’d say we should make this `guard let … else fatalError` since the documentation mentions no scenario in which it will return null. And without this observer we’re going to have a very broken pile of software.

zach-klippenstein:
We looked at doing this on Android a while back. One issue with asynchronous render passes is that you can basically end up in a state where you're running the wrong action.

If you have a workflow that renders the same child in different states, with different actions specified to handle outputs in those states, like this:
```kotlin
when(state) {
  is Foo → renderChild(Child) { FooAction }
  is Bar → renderChild(Child) { BarAction } 
}
```
If `Child` emits outputs twice synchronously (e.g. `Child` subscribes to the same signal twice) and we don't render for each update, then both outputs will be handled by `FooAction`. This would be weird because if I'm debugging this, I'm going to see `FooAction` handle the first output, and my state update, but then when `Child` outputs the second time, even though I'm in the `Bar` state I'm going to see `FooAction` run again.

swift/Workflow/Sources/RunLoopObserver.swift
AquaGeek: True. It feels a bit weird to add it to this mostly-generic RAII wrapper for run loop observers, but I guess I could make the initializer failable and we could `fatalError` in `WorkflowHost`.

zach-klippenstein:
> I wonder if we actually have this kind of usage that you bring up, Zach, in practice.

I can't say, but this definitely came up a while back on Android, someone actually ran into it and reported it as a bug.

zach-klippenstein:
Anyway, I think a more robust solution to this is queueing. But not necessarily the queue-per-worker model that the Kotlin implementation uses. Queue-per-worker is actually more complicated and probably less performant than necessary. 

A simpler implementation could involve a single, global (to the instance of the workflow runtime) queue that stores functions that look like: `() -> Boolean`. When a worker is run for the first time, you create a lambda that does something like this:
```kotlin
fun updateFromValue(value): Boolean {
  if (subscription.isDisposed) return false
  val action = currentActionForThisWorker(value) // need to update this on every render pass in case a different action is specified
  applyActionToThisNode(action) // responsible for propagating output to parent if necessary
  return true
}
```

Then you subscribe to your Signal/SignalProducer with something like `subscribe { queue.enqueue({ updateFromValue($0) }) }` (basically curry the emitted value into the `updateFromValue` function and enqueue that). This means all workers and signals are only subscribed _once_, the first time they are ran/subscribed, and the node just needs to keep track of all those disposables so it can dispose them when the workflow is torn down.

Rendering event handlers would do something similar, but instead of checking if a subscription is disposed they just need to check if the workflow is still active.

Then at the `WorkflowHost` level, after the render pass finishes, you dequeue the next function from the queue and run it. If it returns false it means that function represents an update for a worker/subscription/rendering-event that is now stale, so repeat until a function returns true. You also need a way to handle if the queue is empty, which could be as simple as registering a callback that will be invoked on the next `enqueue` call. Once a dequeued function returns true, it means the update has been applied, and you can start the next render pass.

I think this implementation is actually not that much work, especially since you don't have to worry about thread safety as long as you ensure all signals/workers emit on your main thread. You get all the queueing behavior you need, and render passes still keep up with updates. I just ran this by @AquaGeek and Ray in person, and it passed the sniff test for them.

AquaGeek:
Well, I patched this against `master` in our internal monorepo, and a lot of UI tests that use `UICollectionView` started failing. (Text field contents get out of sync with the data model.) ☹️

timdonnelly:
> Well, I patched this against `master` in our internal monorepo, and a lot of UI tests that use `UICollectionView` started failing. (Text field contents get out of sync with the data model.) ☹️

Good to know – collection view resolves updates during `layoutSubviews`, so it would be interesting to figure out when that is happening before the run loop observer fires.

AquaGeek:
I spent some time trying to track down exactly where things get out of sync, but I haven't been able to pinpoint it yet. For now I'm closing this pull, as I don't think this is a viable solution. I'm working on building out a sample app with both KIF and XCUITests in it to try and flush out the issue/provide a stable base that we can use to ensure we don't regress in the future.

PR #909, opened 2020-01-24
Make verifyAction and verifyActionResult support no processed action.

### LINKED ISSUE #885: renderTester API should support verifying that render event did not send to sink
I have a workflow and I want to test that when I'm rendering a certain state, that no actions were processed when the user submits invalid data.

Right now, `RenderTestResult` has `verifyAction()` and `verifyActionResult()` methods, but there isn't an easy way to verify that there was no action.

A `verifyNoAction()` method would be helpful for testing this kind of workflow.

zach-klippenstein:
Closes #885.

cc @ijwhelan

### DIFF
--- a/kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RenderTestResult.kt
+++ b/kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RenderTestResult.kt
@@ -30,6 +30,8 @@ interface RenderTestResult<StateT, OutputT : Any> {
    * Asserts that the render pass handled either a workflow/worker output or a rendering event, and
    * passes the resulting [WorkflowAction] to [block] for asserting.
    *
+   * If the workflow didn't process any actions, [block] will be passed [WorkflowAction.noAction].
+   *
    * This is useful if your actions are a sealed class or enum. If you need to test an anonymous
    * action, use [verifyActionResult].
    */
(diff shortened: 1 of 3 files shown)


kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RenderTestResult.kt
```diff
@@ -35,6 +35,13 @@ interface RenderTestResult<StateT, OutputT : Any> {
    */
   fun verifyAction(block: (WorkflowAction<StateT, OutputT>) -> Unit)
 
+  /**
+   * Asserts that the render pass either didn't result in any [WorkflowAction] being performed
+   * (e.g. in the case where a rendering event is disabled by being mapped to a no-op), or that
+   * an action was performed that didn't change the state or emit an output.
```
zach-klippenstein: Is this too lenient? Should this method _require_ no action to have been processed at all? The current behavior could let through an action that has complex logic that just happens to set the state to it's previous value.

rjrjr: My knee-jerk is that a literal "no action was fired" assertion is too much of an implementation detail to encourage in testing. But the real question is whether this satisfies the original requester.

zach-klippenstein: An alternative is to allow `verifyActionResult` to handle the case where no action was fired and just treat it like a no-op action.

ijwhelan: So the original use case I had in mind for this was verifying the following behavior:
- User has entered invalid data into a simple form
- User presses the submit button
- Ensure that no network request was made to actually submit the data

The motivation for this test was to ensure that I'm being properly defensive against the UI not respecting the data validity flag in the state and leaving the submit button enabled. I want to make sure that the lambda I'm handing to the UI to fire when the submit button is clicked is actually a no-op.

I agree that actions are implementation details and I don't want to expose them outside of the workflow, but I think allowing no-op actions might open up some gaps in a test like the one I'm trying to write.

zach-klippenstein:
Reworked this PR: no more `verifyNoAction`, instead `verifyAction` and `verifyActionResult` let you assert `noAction`.

rjrjr:
Still LG.

rjrjr:
Still LGTM

PR #911, opened 2020-01-24
Revert "Accept the change to .idea/misc.xml that AS 4 insists on making."

zach-klippenstein:
This reverts commit 53979140c741be254f7f6de9295c4aa500e65cdf.

My work laptop now insists on reverting this, so I think this was only ever
a problem on my personal machine.


zach-klippenstein:
This doesn't affect the actual code, and I'm pretty sure I'm the only person this is affecting right now, so I'm going to merge without approval.

PR #912, opened 2020-01-24
Adds orchestrator to espresso tests.

rjrjr:


rjrjr:
Branch shard passed, PR shard failed with the same "No tests found" garbage that happens at the CLI. Why is this so hard.

PR #913, opened 2020-01-24
Drastically simplifies ModalViewContainer.

rjrjr:
Trying to protect container authors from calling the `Dialog`
constructor was foolish. Didn't actually simplify anything,
and made it very difficult to manage window size.

Panel sample behavior around soft keyboard is a lot better now.

### DIFF
--- a/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ModalViewContainer.kt
+++ b/kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ModalViewContainer.kt
@@ -25,30 +25,57 @@ import android.view.ViewGroup
 import android.view.ViewGroup.LayoutParams.MATCH_PARENT
 import android.view.ViewGroup.LayoutParams.WRAP_CONTENT
 import androidx.annotation.IdRes
-import androidx.annotation.StyleRes
+import com.squareup.workflow.ui.ModalViewContainer.Companion.binding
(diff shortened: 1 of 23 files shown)


kotlin/.idea/misc.xml
```diff
@@ -19,4 +19,4 @@
   <component name="ProjectType">
     <option name="id" value="Android" />
   </component>
-</project>
\ No newline at end of file
+</project>
```
rjrjr: Can't seem to revert that. Thanks git, thanks AS 3.5 v AS 4.0

kotlin/workflow-ui-android/src/main/java/com/squareup/workflow/ui/ModalViewContainer.kt
```diff
@@ -96,24 +108,34 @@ internal class ModalViewContainer @JvmOverloads constructor(
...
         }
       }
   )
+
+  companion object {
+    /**
+     * Creates a [ViewBinding] for modal container screens of type [H].
+     *
+     * Each view created for [HasModals.modals] will be shown in a [Dialog]
+     * whose window is set to size itself to `WRAP_CONTENT` (see [android.view.Window.setLayout]).
```
rjrjr: Lies

rjrjr:
Update fixes stale kdoc on `PanelContainer`.

PR #922, opened 2020-01-28
Upgrade Kotlin to 1.3.61.

zach-klippenstein:


zach-klippenstein:
We pulled it out so dependabot could notify us. Dependabot can't handle version vars.

PR #940, opened 2020-02-05
Make the context used to start workers configurable for tests.

zach-klippenstein:
This is helpful for tests that need to be able to control time.

This functionality was covered by the test's dispatcher, but since #851 changed the dispatcher used for workers, we need to be able to configure workers' contexts separately.

### DIFF
--- a/kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/Workers.kt
+++ b/kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/Workers.kt
@@ -32,6 +32,7 @@ import kotlinx.coroutines.flow.collect
 import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.flow.produceIn
 import kotlinx.coroutines.plus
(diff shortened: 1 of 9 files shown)


kotlin/workflow-runtime/src/main/java/com/squareup/workflow/internal/Workers.kt
```diff
@@ -131,8 +133,9 @@ private fun <T> Flow<T>.transformToValueOrDone(): Flow<ValueOrDone<T>> = flow {
 
 private fun CoroutineScope.createWorkerScope(
   worker: Worker<*>,
-  key: String
-): CoroutineScope = this + CoroutineName(worker.debugName(key)) + Unconfined
+  key: String,
+  workerContext: CoroutineContext
+): CoroutineScope = this + CoroutineName(worker.debugName(key)) + Unconfined + workerContext
```
tehras: Ahh, I see what you meant when you were describing `CouroutineScope` at your talk.

PR #944, opened 2020-02-07
Make ViewRegistry's key set strictly-typed to KClass<*> instead of Any.

zach-klippenstein:
The lack of typing caused a bug integrating with some legacy code, and wasn't
needed or useful, so fixing the type.


JustinDSN:
LGTM

PR #948, opened 2020-02-10
Add JetBrain's new kotlin binary compatibility validator to start tracking changes to our ABI.

### LINKED ISSUE #243: Should we enforce binary compatibility for Kotlin libraries?
Okio has tooling to ensure they don't accidentally break backwards binary compatibility for minor version bumps. I've not seen a lot of libraries that do this, but it seems to make sense especially for libraries that are likely to be depended on transitively. Kotlin makes it really easy to accidentally change your ABI, due to things like type inference. Should we follow suit?

## Related Reading
 - [Maintaining Compatibility in Kotlin Libraries article](https://zsmb.co/maintaining-compatibility-in-kotlin-libraries/)

zach-klippenstein:
Ignores all leaf sample projects individually, wildcards are not supported (see https://github.com/Kotlin/binary-compatibility-validator/issues/3).

Closes #243.

### DIFF
--- a/kotlin/build.gradle
+++ b/kotlin/build.gradle
@@ -72,6 +72,7 @@ buildscript {
       'timber': "com.jakewharton.timber:timber:4.7.1",
 
       'kotlin': [
+          'binaryCompatibilityValidatorPlugin': "org.jetbrains.kotlinx:binary-compatibility-validator:0.1.1",
           'gradlePlugin': "org.jetbrains.kotlin:kotlin-gradle-plugin:1.3.61",
           'stdLib': [
               'common': "org.jetbrains.kotlin:kotlin-stdlib-common",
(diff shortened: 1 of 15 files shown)


kotlin/build.gradle
```diff
@@ -206,3 +210,31 @@ subprojects {
     ]
   }
 }
+
+// We use JetBrain's Kotlin Binary Compatibility Validator to track changes to our public binary
+// APIs.
+// When making a change that results in a public ABI change, the apiCheck task will fail. When this
+// happens, run ./gradlew apiDump to generate updated *.api files, and add those to your commit.
+// See https://github.com/Kotlin/binary-compatibility-validator
+apply plugin: 'binary-compatibility-validator'
```
tehras: I think you could consider moving this to a separate gradle file if you think this build.gradle will keep growing, but doesn't seem like the pattern right now anyways (based on this file).

zach-klippenstein: Great idea! I'll do that now to keep this file from getting bloated.

zach-klippenstein:
This tool doesn't process Android modules unfortunately. Filed: https://github.com/Kotlin/binary-compatibility-validator/issues/4

zach-klippenstein:
Latest master supports android modules! Will update dependency and API dump once v0.1.2 is released.

PR #960, opened 2020-02-12
[WIP][DNR] Add `key` to Screen

bencochran:
This allows us to drop the key arguments in containers and let screens provide their own keys.

Once we have the ability to have generic screen types (e.g. #815), we can provide something like the following to bring back easy customization of key values by consumers of a screen:

```swift
public struct KeyedScreen<Key: Hashable, ScreenType: Screen>: Screen {
    public init(key: Key, screen: ScreenType) { … }
}

extension Screen {
    func keyed<Key: Hashable>(_ key: Key) -> KeyedScreen<Key, Self> { … }
}
```

### DIFF
--- a/swift/WorkflowUI/Sources/Screen/Screen.swift
+++ b/swift/WorkflowUI/Sources/Screen/Screen.swift
@@ -19,4 +19,21 @@
 ///
 /// Conforming types contain any information needed to populate a screen: data,
 /// styling, event handlers, etc.
-public protocol Screen {}
+public protocol Screen {
+
+    /// The type of key for this screen.
+    associatedtype Key: Hashable = ObjectIdentifier
+
+
+    /// The key for this screen. This is often used by containers to determine screen equality.
+    /// The default value is the object identifier of the screen type.
+    var key: Key { get }
+
+}
+
+
+extension Screen where Key == ObjectIdentifier {
+    public var key: ObjectIdentifier {
+        return ObjectIdentifier(Self.self)
+    }
+}
--- a/swift/Samples/BackStackContainer/Sources/BackStackScreen.swift
+++ b/swift/Samples/BackStackContainer/Sources/BackStackScreen.swift
@@ -27,45 +27,21 @@ public struct BackStackScreen: Screen {
 extension BackStackScreen {
     /// A specific item in the back stack. The key and screen type is used to differentiate reused vs replaced screens.
     public struct Item {
-        public var key: AnyHashable
         public var screen: AnyScreen
-        var screenType: Any.Type
         public var barVisibility: BarVisibility
 
-        public init<ScreenType: Screen, Key: Hashable>(key: Key?, screen: ScreenType, barVisibility: BarVisibility) {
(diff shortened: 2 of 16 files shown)


JustinDSN:
Looks cool and pushes the concern to a more appropriate place. Have you considered using the `Identifiable` protocol?

https://nshipster.com/identifiable/

```
protocol Identifiable {
    associatedtype ID: Hashable
    var id: ID { get }
}
```

```
public protocol Screen : Identifiable {}
```

JustinDSN:
After talking with the team today, I understand a lot more about how SwiftUI uses Generic Types like this to better describe things like "Identifiable" via a concrete type.

It looks like SwiftUI has `IDView<Content, ID>` and `EquatableView<Content>`. 

```
/// A uniquely identified view that can be inserted or removed.
@available(iOS 13.0, OSX 10.15, tvOS 13.0, watchOS 6.0, *)
public struct IDView<Content, ID> where Content : View, ID : Hashable {

    public var content: Content

    public var id: ID

    @inlinable public init(_ content: Content, id: ID)

    /// The type of view representing the body of this view.
    ///
    /// When you create a custom view, Swift infers this type from your
    /// implementation of the required `body` property.
    public typealias Body = Never
}

extension IDView : View {
}

/// A view type that compares itself against its previous value and
/// prevents its child updating if its new value is the same as its old
/// value.
@available(iOS 13.0, OSX 10.15, tvOS 13.0, watchOS 6.0, *)
public struct EquatableView<Content> where Content : Equatable, Content : View {

    public var content: Content

    @inlinable public init(content: Content)

    /// The type of view representing the body of this view.
    ///
    /// When you create a custom view, Swift infers this type from your
    /// implementation of the required `body` property.
    public typealias Body = Never
}
```

swift/WorkflowUI/Sources/Screen/Screen.swift
```diff
@@ -19,4 +19,21 @@
...
-public protocol Screen {}
+public protocol Screen {
+
+    /// The type of key for this screen.
+    associatedtype Key: Hashable = ObjectIdentifier
+
+
+    /// The key for this screen. This is often used by containers to determine screen equality.
+    /// The default value is the object identifier of the screen type.
+    var key: Key { get }
```
AquaGeek: The ergonomics of this are much better, but doesn't this create conflicts if two unrelated screens happen to provide the same `Key`? Having this on the container API means the scope of the key is limited to that container.

rjrjr:
I think we need some kind of solution for this use case on the Swift side. It's super common. Ideally it'd be somewhat consistent w/kotlin's scheme, but that's not a deal breaker.

https://github.com/square/workflow/tree/master/kotlin/workflow-ui/core-common/src/main/java/com/squareup/workflow/ui

AquaGeek:
Please reopen this against the Swift repo:

```bash
git remote add swift git@github.com:square/workflow-swift.git
git push swift bc/screen-keys:bc/screen-keys
```

PR #962, opened 2020-02-13
Fix release instructions to pull master instead of resetting.

zach-klippenstein:


dhavalshreyas:
@zach-klippenstein : Any reason we should do `git pull` instead of `git reset --hard origin/master`?

zach-klippenstein:
Don't want to merge into a stale local `master` branch.

RELEASING.md
```diff
@@ -76,7 +76,7 @@ and assign it to @bencochran or @aquageek.*
 1. If this was a fix release, merge changes to the master branch:
    ```bash
    git checkout master
-   git reset --hard origin/master
+   git pull
```
dhavalshreyas: ahh, good point! 

can we do:
```suggestion
   git fetch
   git reset --hard origin/master
```

This will protect us if we have some accidental commits in master.

zach-klippenstein: Reset hard will _lose_ accidental commits, pull will abort and ask to resolve. Also pull is just simpler, these instructions are complicated enough.

dhavalshreyas: Sgtm!

Issue #930, opened 2020-02-03
[swift] Remove all project files for Swift implementations


dhavalshreyas:
1. Use `cocoapods-generate` to generate the project files.
2. Update .travis.yml to gen projects on CI
3. Delete .xcodeproj project files

amorde:
Initially we were considering migrating the Tutorial project to `pod gen`, but I think we should keep it as-is for two reasons:

1. Engineers don't work on the tutorial often, so merge conflicts in the project file are not likely to be an issue
2. It helps for people new to the project to be able to open a workspace and get started quickly, without having to use `pod gen`. Only `pod install` is required.

Going to close as complete, but we can open another issue if we think the tutorial needs to be migrated as well.

PR #964, opened 2020-02-13
[swift] Use generated projects for development & sample apps

amorde:
Progress towards #930 (last remaining project is `Tutorial`)

This introduces a new podspec `Development.podspec` that contains app specs for each of the tutorials.

For development setup, run the following:

```shell
bundle install
bundle exec pod gen Development.podspec --auto-open
```

This will open a project with the following schemes:

Screen Shot 2020-02-12 at 6 13 59 PM

The `Development` framework depends on both `Workflow` and `WorkflowUI`, and can be used to run all unit tests with `CMD-U`.

Screen Shot 2020-02-12 at 6 18 51 PM

While the `Development` framework could also run all tests on CI, I kept the jobs separate for consistency and job parallelism.

## Snapshot Tests

Reference images are stored in `swift/Samples/SnapshotTests/`. For new modules, a folder should be added with the module name, ex. `swift/Samples/SnapshotTests/SplitScreenContainer/`

Each module can include their own `ReferenceImages_64` which will be picked up by `FBSnapshotTestCase`.

Failure diffs will be stored in `swift/FailureDiffs`


amorde:
Looks like the swiftdoc build script will need updating as well

Development.podspec
```diff
@@ -0,0 +1,91 @@
...
+  s.source_files = 'swift/Samples/Dummy.swift'
+
+  s.subspec 'Dummy' do |ss|
+  end
+
+  s.default_subspecs = 'Dummy'
+
+  dir = Pathname.new(__FILE__).dirname
+
+  s.scheme = {
```
JustinDSN: Do we need the environment variables in both places? What happens when we add a second Sample App with their own Reference Images? We could handle that as a fast follow, but eventually I'll be merging my vertical stack container screen which also will have snapshot tests.

https://github.com/square/workflow/pull/964/files#diff-fd7235628a2b67d5e50864050de83789R75

Development.podspec
```diff
@@ -0,0 +1,91 @@
...
+        'FB_REFERENCE_IMAGE_DIR' => dir.join('swift/Samples/SplitScreenContainer/SnapshotTests/ReferenceImages'),
+        'IMAGE_DIFF_DIR' => dir.join('swift/Samples/SplitScreenContainer/SnapshotTests/FailureDiffs'),
+      }
+    }
+
+  s.app_spec 'SampleApp' do |app_spec|
+    app_spec.source_files = 'swift/Samples/SampleApp/Sources/**/*.swift'
+    app_spec.resources = 'swift/Samples/SampleApp/Resources/**/*.swift'
+    app_spec.pod_target_xcconfig = {
+      'INFOPLIST_FILE' => '${PODS_ROOT}/../../swift/Samples/SampleApp/Configuration/Info.plist'
```
JustinDSN: What's this all about?

Gemfile
```diff
@@ -1,4 +1,4 @@
 source 'https://rubygems.org'
 
 gem 'cocoapods', '1.8.4'
-gem 'cocoapods-generate', '~> 1.0'
+gem 'cocoapods-generate', git: 'https://github.com/square/cocoapods-generate', ref: '2dcee88'
```
JustinDSN: Can we add a comment as to why we need to use this ref and when we can go back to a normal released version?

Development.podspec
amorde: We need them in both places if we want to be able to run these tests in both the `Development` and `Development-Unit-SplitScreenTests` schemes. We could remove them from this scheme but then switching to the `Development-Unit-SplitScreenTests` will be required.

That said, I can pull them out into a variable so they aren't defined twice

Development.podspec
amorde: There was already an existing plist for this sample - we can let `pod gen` create one for us instead if its not needed, but I had added this because one of the samples needed the iOS 13 scene delegation that was specified in their plist (I believe it was the SwiftUI one that isn't supported ATM)

amorde: I'll double check and remove this if it isn't necessary

amorde:
@JustinDSN this is ready for another look - I updated the PR description to include details about snapshot tests, including a shared folder for reference images

Development.podspec
JustinDSN: This file looks to be deleted?

Development.podspec
```diff
@@ -0,0 +1,103 @@
...
+    test_spec.dependency 'Development/SampleSplitScreen'
+    test_spec.app_host_name = 'Development/SampleSplitScreen'
+    test_spec.requires_app_host = true
+    test_spec.source_files = 'swift/Samples/SplitScreenContainer/SnapshotTests/**/*.swift'
+
+    test_spec.framework = 'XCTest'
+
+    test_spec.dependency 'iOSSnapshotTestCase'
+
+    test_spec.resources = 'swift/Samples/SnapshotTests/SplitScreenContainer/ReferenceImages_64'
```
JustinDSN: I don't think this is required.

amorde: This is required, since we are now copying the resources into the bundle. We had discussed this change offline, but I can go back to using global paths if you prefer

Development.podspec
amorde: Yep, good catch

Development.podspec
amorde: Extra context: we now only have the `IMAGE_DIFF_DIR` specified as an environment variable

https://github.com/square/workflow/pull/964/files#diff-fd7235628a2b67d5e50864050de83789R23

amorde: Discussed offline, but we decided on using both methods so that:

1. When recording, it places the created reference images in a sane location
2. When running the tests from different contexts (ex. an app host vs. the test bundle itself) the  resources will always be available without naming conflicts. By default the name of the bundle is a component of the path

PR #969, opened 2020-02-15
Fix type parameter for expectWorker accepting Worker instance.

zach-klippenstein:
The worker type should be unbounded, _not_ the same as the _Workflow's_ output type.

### DIFF
--- a/kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RenderTester.kt
+++ b/kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RenderTester.kt
@@ -265,9 +265,9 @@ interface RenderTester<PropsT, StateT, OutputT : Any, RenderingT> {
 /* ktlint-disable parameter-list-wrapping */
 fun <PropsT, StateT, OutputT : Any, RenderingT>
     RenderTester<PropsT, StateT, OutputT, RenderingT>.expectWorker(
-  doesSameWorkAs: Worker<OutputT>,
+  doesSameWorkAs: Worker<*>,
   key: String = "",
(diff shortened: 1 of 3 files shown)


zach-klippenstein:
Fixed the `EmittedOutput` type.

kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RenderTester.kt
```diff
@@ -265,9 +265,9 @@ interface RenderTester<PropsT, StateT, OutputT : Any, RenderingT> {
 /* ktlint-disable parameter-list-wrapping */
 fun <PropsT, StateT, OutputT : Any, RenderingT>
     RenderTester<PropsT, StateT, OutputT, RenderingT>.expectWorker(
-  doesSameWorkAs: Worker<OutputT>,
+  doesSameWorkAs: Worker<*>,
   key: String = "",
-  output: EmittedOutput<OutputT>? = null
+  output: EmittedOutput<Any?>? = null
```
tehras: Why is this required?

zach-klippenstein: This is how you "fake" an output from this worker. This type is the type of the output from the worker.

tehras:
Oh duhhh. Makes sense

kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RenderTester.kt
tehras: Oh duhhh.

PR #966, opened 2020-02-13
Add basic UI tests for every Android sample app.

zach-klippenstein:
Closes #965.

PR #970, opened 2020-02-15
Add more UI tests for the poetry app.

zach-klippenstein:
~TODO: These tests are flaky on some devices. See https://github.com/square/workflow/pull/966#discussion_r379545241~

Tests seem to be a lot more stable with Github Actions. I think we're good to merge unless future runs show these are still flaky.


zach-klippenstein:
Nope, still flaky.

PR #973, opened 2020-02-17
[WIP] Update Swift CI to Xcode 11.3

bencochran:


amorde:
I've seen the snapshot tests failures locally as well - the diff doesn't show any noticeable difference, so there might be a 1 pixel or < 1 pixel difference.

@JustinDSN Should increase the tolerance slightly?

amorde:
Ah nevermind - I've rerun them against an iOS 13.2 simulator and I'm now seeing significant changes. Here's the diff I get when running on iOS 13.2.2:

diff_test_splitRatio_third@2x

dhavalshreyas:
Fixed in #1052

PR #974, opened 2020-02-17
[swift] Remove view registry

### LINKED ISSUE #815: [Swift] Proposal: Remove ViewRegistry

bencochran:
This removes the view registry in favor of having screens provide their own view controllers (via `ViewControllerDescription`).

This improves compile-time safety as well as allows for easier container types (we can now have generic screen types, and pass-through containers no longer must provide a wrapper view
controller).

Closes #815


swift/WorkflowUI/Sources/Screen/ScreenViewController.swift
```diff
@@ -59,4 +71,15 @@ open class ScreenViewController<ScreenType: Screen>: UIViewController {
 
 }
 
+extension ScreenViewController {
+
+    public final class func description(for screen: ScreenType) -> ViewControllerDescription {
```
bencochran: Needs doc comment

swift/Samples/SplitScreenContainer/Sources/SplitScreenContainerViewController.swift
```diff
@@ -16,61 +16,44 @@
...
+    private var leadingContentViewController: DescribedViewController
     private lazy var leadingContainerView: ContainerView = .init()
 
     private lazy var separatorView: UIView = .init()
 
-    private var trailingContentViewController: ScreenViewController<TrailingScreenType>? = nil
+    private var trailingContentViewController: DescribedViewController
     private lazy var trailingContainerView: ContainerView = .init()
     
-    private var currentRatio: CGFloat {
```
bencochran: got rid of these redundant copies of the values while I was in here

swift/Samples/TicTacToe/Sources/Authentication/LoginScreen.swift
```diff
@@ -24,33 +24,23 @@ struct LoginScreen: Screen {
...
-    required init(screen: LoginScreen, viewRegistry: ViewRegistry) {
-        welcomeLabel = UILabel(frame: .zero)
-        emailField = UITextField(frame: .zero)
-        passwordField = UITextField(frame: .zero)
-        button = UIButton(frame: .zero)
-        super.init(screen: screen, viewRegistry: viewRegistry)
-
-        update(with: screen)
-    }
+fileprivate final class LoginViewController: UIViewController {
```
bencochran: Made this one not be a subclass of `ScreenViewController` just to show that is’s not necessary as a superclass

swift/Samples/TicTacToe/Sources/Authentication/LoginScreen.swift
```diff
@@ -24,33 +24,23 @@ struct LoginScreen: Screen {
...
-}
-
-
-extension ViewRegistry {
 
-    public mutating func registerLoginScreen() {
-        self.register(screenViewControllerType: LoginViewController.self)
+    var viewControllerDescription: ViewControllerDescription {
+        return ViewControllerDescription(
+            build: { LoginViewController() },
```
bencochran: Need to pass the screen into this init since I made the view description not always immediately call update (unless we decided that’s the way to go)

swift/WorkflowUI/Sources/ViewControllerDescription/DescribedViewController.swift
```diff
@@ -19,51 +19,60 @@
 import UIKit
 
 
-internal final class AnyScreenViewController: ScreenViewController<AnyScreen> {
+public final class DescribedViewController: UIViewController {
```
bencochran: This isn’t a rename, but a rewrite. I suppose they do perform mostly the same job, though.

swift/WorkflowUI/Sources/Screen/AnyScreen/AnyScreen.swift
```diff
@@ -19,24 +19,22 @@
 import UIKit
 
 public struct AnyScreen: Screen {
```
bencochran: Happy to see this collapse into just a passthrough for the view controller description

swift/WorkflowUI/Sources/ViewControllerDescription/ViewControllerDescription.swift
```diff
@@ -0,0 +1,77 @@
...
+            update(viewController)
+        }
+    }
+
+    /// Construct and update a new view controller as described by this view
+    /// controller description.
+    internal func buildViewController() -> UIViewController {
+        let viewController = build()
+        assert(canUpdate(viewController: viewController), "View controller description built a view controller it cannot update (\(viewController) is not exactly type \(viewControllerType))")
+        return viewController
```
bencochran: I initially considered calling update with the view controller after building it. Technically that’s redundant for our ScreenViewController subclasses (since they all take the screen in in init), but I’m worried that omitting it will cause a lot of people to initially build a view controller that’s not fully populated without realizing it. Anyone opposed to calling `update` eagerly here?

swift/WorkflowUI/Sources/ViewControllerDescription/ViewControllerDescription.swift
```diff
@@ -0,0 +1,77 @@
...
+    ///
+    /// - Parameters:
+    ///   - type: The type of view controller produced by this description.
+    ///           Typically, should should be able to omit this parameter, but
+    ///           in cases where type inference has trouble, it’s offered as
+    ///           an escape hatch.
+    ///   - build: Closure that produces a new instance of the view controller
+    ///   - update: Closure that updates the given view controller
+    public init<VC: UIViewController>(type: VC.Type = VC.self, build: @escaping () -> VC, update: @escaping (VC) -> Void) {
+        self.viewControllerType = type
```
dhavalshreyas: Instead of doing this, what if we hold-on to the `canUpdate` block?

```
self.canUpdate = { untypedVC in
  return untypedVC is VC
}
```

bencochran: I played with that locally, but found it really useful for debugging to have the type around in a printable manner. I also don’t think we want `is` because we only want exact matches, not subclasses IMO.

swift/Samples/SplitScreenContainer/Sources/SplitScreenContainerViewController.swift
```diff
@@ -16,61 +16,44 @@
...
-        super.init(screen: screen, viewRegistry: viewRegistry)
+    required init(screen: ContainerScreen) {
+        leadingContentViewController = DescribedViewController(screen: screen.leadingScreen)
+        trailingContentViewController = DescribedViewController(screen: screen.trailingScreen)
+        super.init(screen: screen)
     }
 
-    override public func screenDidChange(from previousScreen: ContainerScreen) {
+    override internal func screenDidChange(from previousScreen: ContainerScreen) {
+        if screen.ratio != previousScreen.ratio {
```
JustinDSN: Nice usage of `screenDidChange` to detect and set the `needsAnimatedLayout` flag. Didn't think about using this API for this.

swift/WorkflowUI/Sources/Screen/Screen.swift
```diff
@@ -14,9 +14,18 @@
  * limitations under the License.
  */
 
+#if canImport(UIKit)
```
JustinDSN: This will effectively remove the Screen protocol for non UIKit targeting consumers. Just calling it out, we may not care right away but this may break external consumers if they build for MacOS for example.

swift/WorkflowUI/Sources/ViewControllerDescription/DescribedViewController.swift
```diff
@@ -19,51 +19,60 @@
...
             }
-            currentViewController.removeFromParent()
-
-            currentViewController = screen.makeViewController(from: viewRegistry)
-            addChild(currentViewController)
+            currentViewController = description.buildViewController()
             if isViewLoaded {
+                addChild(currentViewController)
                 view.addSubview(currentViewController.view)
+                currentViewController.view.frame = view.bounds
```
JustinDSN: Do we need to set the frame here? It's set in `viewDidLayoutSubviews`, also, we're not setting it in `viewDidLoad`, curious the heuristics of when to set frame as I came across this in the split screen container work.

swift/WorkflowUI/Sources/ViewControllerDescription/ViewControllerDescription.swift
amorde: could that cause an issue if an `update` implementation expects the view controller to be inside a window?

AquaGeek:
Cases where we want to change the resulting view controller type (e.g. adapt a `Screen` defined elsewhere to our use case) now require you to wrap that in your own `Screen` type (or pull off the various properties that you need/care about), right?

Also, do we care that this requires that we use view controllers as the currency globally (as opposed to e.g. views)?

swift/WorkflowUI/Sources/ViewControllerDescription/ViewControllerDescription.swift
bencochran: `canUpdate` is just the type check, the `update` implementation doesn’t play in here. Though, if someone wrote an `update` method that required being in a window, I’d be very sad at them.

swift/WorkflowUI/Sources/ViewControllerDescription/ViewControllerDescription.swift
bencochran: Since a single view controller description is responsible for building and updating, I can’t imagine a scenario where you’d reasonably build a type of a superclass and then be passed a value of a subclass, except by mistake.

AquaGeek: Do we actually want to be able to pass in something other than the default (`VC.self`) here?

swift/WorkflowUI/Sources/Screen/Screen.swift
bencochran: We could. But I doin’t think any of WorkflowUI makes sense outside of UIKit. If we wanted macOS support we’d probably want a separate module for that. It’s totally valid for a non-uikit related workflow to produce a rendering type that’s not a `Screen`.

swift/WorkflowUI/Sources/ViewControllerDescription/DescribedViewController.swift
bencochran: We either need to set the frame or `view.setNeedsLayout()`. I guess I opted for the former

bencochran:
> Cases where we want to change the resulting view controller type (e.g. adapt a `Screen` defined elsewhere to our use case) now require you to wrap that in your own `Screen` type (or pull off the various properties that you need/care about), right?

Yep, that’s the intended pattern should we need it. It would also be valid for a workflow to render a type that’s not a `Screen` then have a consumer conform it’s type to `Screen` to provide a view controller, though I like that approach less.

> Also, do we care that this requires that we use view controllers as the currency globally (as opposed to e.g. views)?

Not personally. It’s already the case today with view registry. We could provide a pretty easy convenience for `ViewControllerDescription` that describes a `UIView` if we wanted.

swift/WorkflowUI/Sources/ViewControllerDescription/ViewControllerDescription.swift
bencochran: In the case of the extension on `ScreenViewController`, it was inferring a type of `ScreenViewController<FooScreen>` instead of `FooViewController` with older Swift versions (where I couldn’t do `Self(screen: screen)` and instead had to do `self.init(screen: screen)`. Adding this hinting allowed me to fix that.

swift/WorkflowUI/Sources/Screen/Screen.swift
amorde: what about projects that want to use Workflow with SwiftUI?

amorde: I know UIViewController's can be used in SwiftUI but UIKit technically isn't a requirement to develop in SwiftUI

bencochran: We have WorkflowSwiftUI, which doesn’t use WorkflowUI `Screen`

dhavalshreyas:
> This is awesome, but how bad is updating existing code to use this going to be? Should we leave the `ViewRegistry` parameters in for now and do a release with support for the new technique, so we can migrate existing code over incrementally?

We tried a few approaches (#816) which would allow us to migrate incrementally, however there were some risks with this approach. Since we'll need the same `ViewRegistry` to get carried over to the leafs, if we use a non-ViewRegistry `Screen` in between, there will be a run-time crash. 

The hard-removal approach is a lot more up-front work to do the migration, however we'll have compile-time safety and will be a cleaner migration.

swift/WorkflowUI/Sources/ViewControllerDescription/ViewControllerDescription.swift
bencochran: Oh, I totally misattributed your question and thought it was about the `assert(canUpdate…`. So, yeah doing an update eagerly here would break a view controller’s ability to assume it’s in a window when it gets updated. But I strongly believe that’s not an assumption we should allow people to make.

PR #977, opened 2020-02-20
Upgrade Compose to dev05.

zach-klippenstein:
Release notes:
 - https://developer.android.com/jetpack/androidx/releases/compose#0.1.0-dev05
 - https://developer.android.com/jetpack/androidx/releases/ui#0.1.0-dev05

### DIFF
--- a/kotlin/build.gradle
+++ b/kotlin/build.gradle
@@ -16,7 +16,7 @@
 buildscript {
   ext.versions = [
       'targetSdk': 29,
-      'compose': '0.1.0-dev04',
+      'compose': '0.1.0-dev05',
   ]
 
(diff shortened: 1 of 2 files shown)


kotlin/.buildscript/configure-compose.gradle
```diff
@@ -19,6 +19,7 @@ android {
     compose true
   }
   composeOptions {
+    kotlinCompilerVersion "1.3.61-dev-withExperimentalGoogleExtensions-20200129"
```
tehras: Just curious - why is this required?

zach-klippenstein: The docs said to add it, but they were wrong about other things – I just removed it and it seems to work fine.

PR #980, opened 2020-02-23
Bump maven publish gradle plugin to 0.9.0

bnvinay92:


zach-klippenstein:
Thanks, but we already have a PR for this: https://github.com/square/workflow/pull/947

PR #985, opened 2020-02-26
Introduce a view binding function that works with AndroidX ViewBindings.

### LINKED ISSUE #984: Explore AndroidX ViewBinding integration for LayoutRunner

zach-klippenstein:
Initially proposed by @0legg.

Closes #984.

```kotlin
val HelloBinding: ViewBinding<HelloWorkflow.Rendering> =
  bindViewBinding(HelloGoodbyeLayoutBinding::inflate) { rendering, containerHints ->
    helloMessage.text = rendering.message
    helloMessage.setOnClickListener { rendering.onClick }
  }
```

### DIFF
--- a/kotlin/samples/hello-workflow/build.gradle
+++ b/kotlin/samples/hello-workflow/build.gradle
@@ -35,5 +35,6 @@ dependencies {
   implementation project(':workflow-runtime')
 
   implementation deps.androidx.appcompat
+  implementation deps.androidx.viewbinding
   implementation deps.rxjava2.rxjava2
 }
--- a/kotlin/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/LayoutRunner.kt
+++ b/kotlin/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/LayoutRunner.kt
@@ -20,7 +20,11 @@ import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import androidx.annotation.LayoutRes
+import androidx.viewbinding.ViewBinding
+import com.squareup.workflow.ui.LayoutRunner.Companion.bind
 import kotlin.reflect.KClass
+
+typealias ViewBindingInflater<BindingT> = (LayoutInflater, ViewGroup?, Boolean) -> BindingT
 
 /**
  * A delegate that implements a [showRendering] method to be called when a workflow rendering
@@ -51,6 +55,31 @@ import kotlin.reflect.KClass
  *    val TicTacToeViewBuilders = ViewRegistry(
  *        NewGameLayoutRunner, GamePlayLayoutRunner, GameOverLayoutRunner
  *    )
+ *
+ * ## AndroidX ViewBinding
+ *
+ * [AndroidX ViewBinding][ViewBinding] is supported in two ways.
+ * In most cases, you can use the `bind` function that takes a function and avoid implementing
+ * [LayoutRunner] at all.
+ *
+ * If you need to perform some set up before [showRendering] is called, use the
+ * `bind` overload that takes:
+ *  - a reference to a `ViewBinding.inflate` method and
+ *  - a [LayoutRunner] constructor that accepts a [ViewBinding]
+ *
+ *   class HelloLayoutRunner(private val binding: HelloGoodbyeLayoutBinding) : LayoutRunner<Rendering> {
+ *
+ *     override fun showRendering(rendering: Rendering) {
+ *       binding.messageView.text = rendering.message
+ *       binding.messageView.setOnClickListener { rendering.onClick(Unit) }
+ *     }
+ *
+ *     companion object : ViewFactory<Rendering> by bind(
+ *         HelloGoodbyeLayoutBinding, ::HelloLayoutRunner
+ *     )
+ *   }
+ *
+ * If the view does not need to be initialized, the [bind] function can be used instead.
  */
 interface LayoutRunner<RenderingT : Any> {
   fun showRendering(
@@ -69,8 +98,7 @@ interface LayoutRunner<RenderingT : Any> {
       contextForNewView: Context,
       container: ViewGroup?
     ): View {
-      return LayoutInflater.from(container?.context ?: contextForNewView)
-          .cloneInContext(contextForNewView)
+      return contextForNewView.viewBindingLayoutInflater(container)
           .inflate(layoutId, container, false)
           .apply {
             bindShowRendering(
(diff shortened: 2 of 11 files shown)


kotlin/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/LayoutRunner.kt
```diff
@@ -92,6 +120,31 @@ interface LayoutRunner<RenderingT : Any> {
       noinline constructor: (View) -> LayoutRunner<RenderingT>
     ): ViewBinding<RenderingT> = Binding(RenderingT::class, layoutId, constructor)
 
+    inline fun <BindingT : androidx.viewbinding.ViewBinding, reified RenderingT : Any> bind(
+      noinline bindingInflater: ViewBindingInflater<BindingT>,
```
0legg: Passing method reference here will allow us to avoid reflection, and will make consumer side more configurable. On the other hand, it'll increase boilerplate a bit.

zach-klippenstein: Yea, not a huge fan of this reflection approach, but I'm not as worried as I'd usually be because it's reflecting on generated code that's guaranteed to have this method. But I would probably rather not merge it and only have the explicit version at least for now. It's really not that much boilerplate, and much less magic.

zach-klippenstein: So the `buildView` API passes you a `ViewGroup` parent for the view-to-inflate, but it's nullable. Unfortunately, unlike the `LayoutInflater.inflate` method, the parent parameter to `ViewBinding.inflate` is _not_ nullable. That means this binding needs to check if it's null, and decide which `inflate` method to call based on whether a container is provided or not. It's really awkward to pass two separate `inflate` references to the binding, especially since they have the same name. 

For this reason, I'm wondering if the reflection-based binding is the only reasonable one. We can do the reflection lookup lazily and cache the result, to minimize the impact on performance.

kotlin/samples/hello-workflow/src/main/java/com/squareup/sample/helloworkflow/HelloLayoutRunner.kt
```diff
@@ -15,25 +15,12 @@
...
-  override fun showRendering(
-    rendering: HelloWorkflow.Rendering,
-    containerHints: ContainerHints
-  ) {
-    messageView.text = rendering.message
-    messageView.setOnClickListener { rendering.onClick() }
+val HelloBinding: ViewBinding<HelloWorkflow.Rendering> =
+  bindViewBinding(HelloGoodbyeLayoutBinding::inflate) { rendering, _ ->
+    helloMessage.text = rendering.message
+    helloMessage.setOnClickListener { rendering.onClick }
```
zach-klippenstein: Should we leave this sample using the `View`-based `LayoutRunner`, make it use a binding-based `LayoutRunner`, or show this most succinct style?

zach-klippenstein: Should we leave this sample using the `View`-based `LayoutRunner`, make it use a binding-based `LayoutRunner`, or show this most succinct style?

kotlin/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/LayoutRunner.kt
```diff
@@ -51,6 +51,27 @@ import kotlin.reflect.KClass
...
+ *       binding.messageView.text = rendering.message
+ *       binding.messageView.setOnClickListener { rendering.onClick(Unit) }
+ *     }
+ *
+ *     companion object : ViewBinding<Rendering> by bind(
+ *         HelloGoodbyeLayoutBinding, ::HelloLayoutRunner
+ *     )
+ *   }
+ *
+ * If the view does not need to be initialized, the [bindViewBinding] function can be used instead.
```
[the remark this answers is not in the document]
rjrjr: And we _really_ have to rename our own `ViewBinding` class. I don't think we should cram that into this PR, but we should do it before the next release. :(

zach-klippenstein:
Looks like a legit UI test failure.

kotlin/samples/hello-workflow/build.gradle
```diff
@@ -28,6 +28,8 @@ android {
 
     testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
   }
+
+  buildFeatures.viewBinding = true
```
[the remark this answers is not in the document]
0legg: It definitely should

kotlin/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/AndroidXViewBinding.kt
```diff
@@ -0,0 +1,33 @@
...
+ * ```
+ * val HelloBinding: ViewBinding<Rendering> =
+ *   bindViewBinding(HelloGoodbyeLayoutBinding::inflate) { rendering, containerHints ->
+ *     helloMessage.text = rendering.message
+ *     helloMessage.setOnClickListener { rendering.onClick(Unit) }
+ *   }
+ * ```
+ *
+ * If you need to initialize your view before [showRendering] is called, create a [LayoutRunner]
+ * and create a binding using `LayoutRunner.bind` instead.
```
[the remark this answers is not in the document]
zach-klippenstein: This was intentional, because this links to the wrong `bind` method and the link adds more confusion than anything. There's a super old dokka issue about this: https://github.com/Kotlin/dokka/issues/80

kotlin/samples/hello-workflow/build.gradle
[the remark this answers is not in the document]
zach-klippenstein: It should be if we want to recommend ViewBinding as the best practice. I was thinking of this more as "if you want to use ViewBinding, we'll make it easy for you", and "if you want to use ViewBinding" implies that you know how to turn it on. But I would not be opposed to recommending ViewBinding as the default over raw views.

kotlin/samples/hello-workflow/src/main/java/com/squareup/sample/helloworkflow/HelloLayoutRunner.kt
zach-klippenstein: I think the only one might be the screen in the dungeon sample where we initialize the recycler view.

zach-klippenstein:
`ViewBinding` rename is filed as #988.

kotlin/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/LayoutRunner.kt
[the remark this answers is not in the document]
zach-klippenstein: I didn't want to hang this off `LayoutRunner` because its API doesn't have anything to do with `LayoutRunner`, it's just an implementation detail. But "`LayoutRunner` as one-stop-shopping for dealing with Layout files" makes sense, so I'll change it.

kotlin/samples/hello-workflow/src/main/java/com/squareup/sample/helloworkflow/HelloLayoutRunner.kt
zach-klippenstein: I converted a bunch of samples, but wow are there a lot. Need to timebox this for now. We shouldn't merge this until all samples are converted though, so it's not confusing what the "right" way to do things is.

zach-klippenstein:
> Do any of the other current samples do the kind of pre-render setup that would justify LayoutRunner + Binding

@rjrjr I've already found a few.

zach-klippenstein:
I'm going to only convert the HelloWorkflow sample in this PR, and then convert all the other samples in a follow-up so it doesn't block the 0.24.0 branch cut. See #1012.

PR #993, opened 2020-03-03
Upgrade Kotlin to 1.3.70 and fix related build errors.

zach-klippenstein:
2 commits, first one factors the version number back out into a variable.

### DIFF
--- a/kotlin/build.gradle
+++ b/kotlin/build.gradle
@@ -17,6 +17,7 @@ buildscript {
   ext.versions = [
       'targetSdk': 29,
       'compose': '0.1.0-dev05',
+      'kotlin': '1.3.61',
   ]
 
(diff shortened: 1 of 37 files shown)


kotlin/build.gradle
```diff
@@ -17,6 +17,7 @@ buildscript {
   ext.versions = [
       'targetSdk': 29,
       'compose': '0.1.0-dev05',
+      'kotlin': '1.3.70',
```
tehras: Didn't you previously mention that dependabot doesn't pick these up?

zach-klippenstein: Yea, but as mentioned in the commit message it doesn't really matter. dependabot will only upgrade one dep at a time, and so we always have to manually edit the PRs to update everything at once anyway.

Issue #739, opened 2019-11-12
ContainerHints for Swift


rjrjr:

dhavalshreyas:
Addressed in #999

PR #999, opened 2020-03-04
Add ViewEnvironment (#739)

bencochran:
(ViewEnvironment is the Swift implementation of Kotlin’s ContainerHints. I included the straw-man rename of them to ViewEnvironment.)

ViewEnvironment is a container of strongly-typed key/value pairs that can be passed down the view-side of a workflow tree. This allows containers to provide view-oriented hinting to their children as well as allows for view-side changes to view-only environment without re-rendering the workflow tree (for example, runtime theme changes).


swift/WorkflowUI/Sources/Screen/ViewEnvironment/EnvironmentScreen.swift
```diff
@@ -0,0 +1,57 @@
...
+
+}
+
+#if canImport(UIKit)
+
+extension Screen {
+
+    /// Returns a new screen that will render the contents of this screen, with
+    /// the given environment key path set to the given value.
+    public func withEnvironment<Value>(_ keyPath: WritableKeyPath<ViewEnvironment, Value>, value: Value) -> EnvironmentScreen<Value, Self> {
```
bencochran: This extension is getting a little `SwiftUI`-y. Thoughts?

swift/Samples/SplitScreenContainer/Sources/Environment+SplitScreen.swift
```diff
@@ -0,0 +1,41 @@
...
+
+    internal(set) public var splitScreenPosition: SplitScreenPosition {
+        get { return self[SplitScreenPositionKey.self] }
+        set { self[SplitScreenPositionKey.self] = newValue }
+    }
+
+}
+
+private enum SplitScreenPositionKey: ViewEnvironmentKey {
+    static var defaultValue: SplitScreenPosition = .none
```
JustinDSN: This is cool! I'm kinda blown away that a private type's `internal static var` is accessible from outside this file.

dhavalshreyas: `SplitScreenPositionKey` is not accessed from outside this file.

swift/WorkflowUI/Sources/Screen/ViewEnvironment/EnvironmentScreen.swift
dhavalshreyas: Can we make it *more* `SwiftUI`-y?

bencochran: For the full effect I could spell it this way:
```swift
public func environment<Value>(_ keyPath: WritableKeyPath<ViewEnvironment, Value>, _ value: Value) -> EnvironmentScreen<Value, Self>
```

bencochran: I tend to shy away from unnamed parameters, but in the case of this one I think it’s actually alright?

ex:
```
someScreen
    .environment(\.theme, DarkTheme())
    .environment(\.padding, 24)
```

dhavalshreyas: I like that.

swift/Samples/SplitScreenContainer/Sources/Environment+SplitScreen.swift
JustinDSN: But it's default value var is.

swift/WorkflowUI/Sources/Screen/ViewEnvironment/EnvironmentScreen.swift
```diff
@@ -0,0 +1,57 @@
...
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+
+/// A wrapper screen that overrides environment values
+public struct EnvironmentScreen<Value, Content> {
```
dhavalshreyas: Is there value in this? Without this `ViewEnvironment` is fully contained within `viewControllerDescription` and `UIViewController`s.

bencochran: `ViewControllerDescription` itself doesn’t have any knowledge of `ViewEnvironment`.

My desire for this is for wrapper workflows to be able to provide environment values without having to implement a wrapper screen themselves. Let’s use automatic day/night theme switching as an example (ignoring iOS’s dark mode support for a moment). I could have a workflow that has itself subscribes to some signal for time of day, changing its state to `.day` or `.night` as appropriate. With this extension, it’s `render` could be:
```
func render(state: State, context: RenderContext<AutoNightModeWorkflow>) -> Rendering {
    return wrappedWorkflow
        .mapRendering { $0.environment(\.theme, Theme(style: state.timeOfDay) }
        .rendered(with: context)
}
```

swift/WorkflowUI/Sources/Screen/ViewEnvironment/ViewEnvironment.swift
```diff
@@ -0,0 +1,97 @@
...
+        get {
+            if let value = storage[ObjectIdentifier(key)] as? Key.Value {
+                return value
+            } else {
+                return Key.defaultValue
+            }
+        }
+
+        set {
+            storage[ObjectIdentifier(key)] = newValue
```
zach-klippenstein: Given struct's value semantics, does this actually mutate the current value? In the Kotlin API `ContainerHints` is entirely immutable, we don't provide any mutating operators.

swift/WorkflowUI/Sources/Screen/ViewEnvironment/EnvironmentScreen.swift
zach-klippenstein: Doesn't this break the UI/Workflow decoupling? `ViewEnvironment` is purely a view concern, so the workflows shouldn't know about it.

bencochran: IMO, it’s okay for a workflow to put values _into_ the environment. But not to take them out. We absolutely could make a workflow like the one described above do the same job by implementing a wrapper screen type, but it’s effectively going to do the same thing and be the same flow of information.

swift/WorkflowUI/Sources/Screen/ViewEnvironment/ViewEnvironment.swift
bencochran: In a `struct` this is effectively a `mutating func`, so the mutability follows our standard value semantics (isn’t shared).

So for example you cannot do:
```swift
func update(screen: MyScreen, environment: ViewEnvironment) {
    environment[SomeKey.self] = 14
}
```

swift/WorkflowUI/Sources/Screen/ViewEnvironment/EnvironmentScreen.swift
rjrjr: Can we have this discussion in a separate PR, please? It's a big change from the Kotlin behavior, for a speculative use case. And it's not required by the general machinery here, right?

bencochran: Sure. I’ll kick it out of this PR when I update to fix the tutorial code.

rjrjr: While I'm thinking about it, though: one thing to note in defense of this class is that container views themselves can transform child renderings into other screen classes. I've already done that a few times, very handy.

That said, I still don't think this class should be a part of this PR.

dhavalshreyas: Actually, thinking more about this - this might be the primary way for us to set environment values.

rjrjr: https://github.com/square/workflow/issues/1001

bencochran:
Removed `EnvironmentScreen` (into a branch waiting to be PR’d once this lands), fixed broken Tutorials, and updated docs and templates. Let me know if people think this is ready to land.

PR #1002, opened 2020-03-04
Add Swift tutorial code to the Development podspec

bencochran:
I missed the Tutorials when doing ViewEnvironment stuff because they weren’t in the development project. This adds them to that.

### DIFF
--- a/.gen_config.yml
+++ b/.gen_config.yml
@@ -2,6 +2,14 @@ local_sources:
   - .
   - swift/Samples/SplitScreenContainer
   - swift/Samples/BackStackContainer
+  - swift/Samples/Tutorial/Frameworks/TutorialViews
+  - swift/Samples/Tutorial/Frameworks/TutorialBase
+  - swift/Samples/Tutorial/Frameworks/Tutorial1Complete
+  - swift/Samples/Tutorial/Frameworks/Tutorial2Complete
+  - swift/Samples/Tutorial/Frameworks/Tutorial3Complete
+  - swift/Samples/Tutorial/Frameworks/Tutorial4Complete
+  - swift/Samples/Tutorial/Frameworks/Tutorial5Complete
+  - swift/
 platforms:
   - ios
 podspec_paths:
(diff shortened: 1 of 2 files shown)


.gen_config.yml
```diff
@@ -2,6 +2,14 @@ local_sources:
...
   - swift/Samples/SplitScreenContainer
   - swift/Samples/BackStackContainer
+  - swift/Samples/Tutorial/Frameworks/TutorialViews
+  - swift/Samples/Tutorial/Frameworks/TutorialBase
+  - swift/Samples/Tutorial/Frameworks/Tutorial1Complete
+  - swift/Samples/Tutorial/Frameworks/Tutorial2Complete
+  - swift/Samples/Tutorial/Frameworks/Tutorial3Complete
+  - swift/Samples/Tutorial/Frameworks/Tutorial4Complete
+  - swift/Samples/Tutorial/Frameworks/Tutorial5Complete
+  - swift/
```
amorde: is this line needed? I don't think there are any other local podspecs in the `swift` folder

Development.podspec
```diff
@@ -89,6 +89,32 @@ Pod::Spec.new do |s|
...
+      environment_variables: snapshot_test_env
+    }
+  end
+
+  s.test_spec 'Tutorial5Tests' do |test_spec|
+    test_spec.dependency 'BackStackContainer'
+    test_spec.dependency 'TutorialBase'
+    test_spec.dependency 'Tutorial5'
+    test_spec.dependency 'WorkflowTesting'
+    test_spec.requires_app_host = true
```
amorde: can we also do `test_spec.app_host = 'Development/Tutorial'` so the tutorial app is the host? makes it easier to run the tests while working on the tutorial

amorde: actually, the tutorial app might be in a weird state for compiling since its, ya know, a tutorial. probably better to use a generated host

zach-klippenstein:
Should this PR be closed/moved to the swift repo?

AquaGeek:
Yeah, this should be moved over to the Swift repo.

PR #1005, opened 2020-03-05
Rename ContainerHints to ViewEnvironment to match Swift.

### LINKED ISSUE #1000: Rename ContainerHints to ViewEnvironment
Presuming #999 merges, Kotlin should be updated to use the new name introduced there.

zach-klippenstein:
Closes #1000.

Keeps `ContainerHints` around as a deprecated type alias with replacement action.

### DIFF
--- a/kotlin/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/ViewEnvironment.kt
+++ b/kotlin/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/ViewEnvironment.kt
@@ -17,41 +17,48 @@ package com.squareup.workflow.ui
 
 import kotlin.reflect.KClass
 
+@Suppress("unused")
+@Deprecated(
+    "Renamed to ViewEnvironment.",
+    replaceWith = ReplaceWith("ViewEnvironment", "com.squareup.workflow.ui.ViewEnvironment")
+)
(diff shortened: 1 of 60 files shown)


kotlin/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/ViewEnvironment.kt
```diff
@@ -17,54 +17,61 @@ package com.squareup.workflow.ui
 
 import kotlin.reflect.KClass
 
+@Suppress("unused")
+@Deprecated(
+    "Renamed to ViewEnvironment.",
+    replaceWith = ReplaceWith("ViewEnvironment", "com.squareup.workflow.ui.ViewEnvironment")
+)
+typealias ContainerHints = ViewEnvironment
```
rjrjr: Nice. Should you do the same for the key type?

zach-klippenstein: I don't think it's used as ubiquitously (at least an order of magnitude less common to define new keys than to reference the container type), so I think we can just update those references immediately.

zach-klippenstein:
Only remaining red shard is an iOS error, this is ready to merge.

PR #1011, opened 2020-03-06
Enable AndroidX test orchestrator to try to reduce flakiness.

zach-klippenstein:
2 commits:
1. Factor out Android UI test gradle config into a separate file.
2. Enable AndroidX test orchestrator to try to reduce flakiness.

Note this was originally attempted in #912, but we never got it working.


vRallev:
Note that the test orchestrator has a significant impact on performance. I'm surprised that you need it for this repository since you probably save little data. Are there many threading issues?

zach-klippenstein:
There's no threading in most of the tests, and most apps only have a single UI test, so I don't know how this would help either. But I haven't been able to repro the flakes locally so trying this.

PR #1022, opened 2020-03-11
Remove Workers/Subscriptions/ReactiveSwift from Workflow, add SubscriptionProvider to WorkflowUI

dhavalshreyas:


swift/WorkflowUI/Sources/Screen/SubscriptionWorkflow.swift
```diff
@@ -0,0 +1,120 @@
+import Foundation
+import ReactiveSwift
+import Workflow
+
+public struct SubscribedScreen<S: Screen, SubPro: SubscriptionProvider>: Screen {
```
AquaGeek: How does this interact with the nested `Screen`'s view controller? It gets wrapped by `ViewController` below, right?

Is that going to cause issues with any of our containers?

swift/Workflow/Sources/SubtreeManager.swift
```diff
@@ -416,84 +375,6 @@ extension WorkflowNode.SubtreeManager {
 
 }
 
-// MARK: - Workers
-
-extension WorkflowNode.SubtreeManager {
```
AquaGeek: I _love_ that all of this goes away. 🔥

AquaGeek:
This is an interesting approach. I would have thought that it would take more work to remove the Rx bits.

In general, I think there is something interesting in moving subscriptions into a reference type (the view controllers, here). I think that much more closely mirrors the expectation around their lifetimes.

However, I am wary of moving subscriptions into the view layer (even though this sort of backdoor). What happens if a parent Workflow decides not to convert a given screen into a view (due to business logic or programmer error)?

swift/Workflow/Sources/SubtreeManager.swift
dhavalshreyas: Yeah, so much red!

swift/WorkflowUI/Sources/Screen/SubscriptionWorkflow.swift
dhavalshreyas: Not sure how. The big negative is we'll be adding a pretty heft VC just to subscribe.

bencochran:
An example use-case of the “not convert the given screen into a view” is a workflow that manages a tab bar. One of the unselected tabs may want to go do a network request in order to determine what its tab should be badged with. But the tab bar screen isn’t showing its view and will not build a view controller for it.

swift/Workflow/Sources/WorkflowHost.swift
```diff
@@ -81,10 +76,11 @@ public final class WorkflowHost<WorkflowType: Workflow> {
     }
 
     private func handle(output: WorkflowNode<WorkflowType>.Output) {
-        mutableRendering.value = rootNode.render()
+        rendering = rootNode.render()
+        renderingListener.listener?(rendering)
```
zach-klippenstein: I wish kotlin had this optional-invoke operator. We have to do `?.invoke()`

swift/WorkflowUI/Sources/Screen/AnyScreen/AnyScreen.swift
```diff
@@ -42,4 +42,10 @@ public struct AnyScreen: Screen {
 
 }
 
+public extension Screen {
+    func asAny() -> AnyScreen {
+        AnyScreen(self)
+    }
+}
```
zach-klippenstein: I'm surprised this didn't exist already.

PR #1024, opened 2020-03-12
Updating Release instructions

dhavalshreyas:
This is based on some inconsistencies I bumped into during the recent release cycle. It's possible I've been doing something incorrectly, in which case do let me know!

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -13,7 +13,7 @@ and assign it to @bencochran or @aquageek.*
 ---
 1. Merge an update of [the change log](CHANGELOG.md) with the changes since the last release.
 
-1. Make sure you're on the `master` branch (or fix branch, e.g. `v0.1-fixes`).
+1. Create a new release branch `release-v0.1.x` (or checkout fix branch, e.g. `release-v0.0.x`).
 
 1. Confirm that the kotlin build is green before committing any changes
    ```bash
@@ -22,11 +22,17 @@ and assign it to @bencochran or @aquageek.*
 
 1. In `kotlin/gradle.properties`, remove the `-SNAPSHOT` prefix from the `VERSION_NAME` property.
    E.g. `VERSION_NAME=0.1.0`
+   ```bash
+   sed -i '' -e 's/-SNAPSHOT//g' kotlin/gradle.properties
+   ```
 
 1. Create a commit and tag the commit with the version number:
    ```bash
    git commit -am "Releasing v0.1.0."
+   git push origin release-v0.1.x
+
    git tag v0.1.0
+   git push origin v0.1.0
    ```
 
 1. Upload the kotlin artifacts:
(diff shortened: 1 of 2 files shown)


RELEASING.md
```diff
@@ -22,11 +22,17 @@ and assign it to @bencochran or @aquageek.*
...
+   sed -i '' -e 's/-SNAPSHOT//g' kotlin/gradle.properties 
+   ```
 
 1. Create a commit and tag the commit with the version number:
    ```bash
    git commit -am "Releasing v0.1.0."
+   git push origin release-v0.1.x
+
    git tag v0.1.0
+   git push origin v0.1.0
```
dhavalshreyas: We need to push the tag before we can do the next step `pod trunk push...`

RELEASING.md
```diff
@@ -13,7 +13,7 @@ and assign it to @bencochran or @aquageek.*
 ---
 1. Merge an update of [the change log](CHANGELOG.md) with the changes since the last release.
 
-1. Make sure you're on the `master` branch (or fix branch, e.g. `v0.1-fixes`).
+1. Create a new release branch `release-v0.1.x` (or checkout fix branch, e.g. `release-v0.0.x`).
```
zach-klippenstein: This isn't strictly required, but probably a good habit to get into anyway. Thanks for adding.

RELEASING.md
```diff
@@ -44,25 +50,24 @@ and assign it to @bencochran or @aquageek.*
...
-1. Bump the version
+1. Checkout `master` and bump the version (if doing a new release)
    - **Kotlin:** Update the `VERSION_NAME` property in `kotlin/gradle.properties` to the new
      snapshot version, e.g. `VERSION_NAME=0.2.0-SNAPSHOT`.
    - **Swift:** Update `s.version` in `*.podspec` to the new version, e.g. `0.2.0`.
+   ```bash
+   git checkout master
+   sed -i '' -e 's/0.1.0-SNAPSHOT/0.2.0-SNAPSHOT/g' kotlin/gradle.properties 
+   ls *.podspec | xargs sed -i '' -e 's/0.1.0/0.2.0/g'
+   ```
```
zach-klippenstein: Using commands to update config manually and then just committing to master blindly makes me nervous, because if someone copy/pastes this command and forgets to change both values, they've suddenly pushed bad code to master without noticing.

RELEASING.md
```diff
@@ -77,7 +82,7 @@ and assign it to @bencochran or @aquageek.*
    ```bash
    git checkout master
    git pull
-   git merge --no-ff v0.1-fixes
+   git merge --no-ff release-v0.0.x
```
zach-klippenstein: Your previous examples used the release branch `release-v0.1.x`, you probably want to use the same here.
```suggestion
   git merge --no-ff release-v0.1.x
```

RELEASING.md
```diff
@@ -86,7 +91,7 @@ and assign it to @bencochran or @aquageek.*
 
 ## Deploying the documentation website
 
-Official Workflow documentation lives at <https://squareup.github.io/workflow>. The website content
+Official Workflow documentation lives at <https://square.github.io/workflow>. The website content
```
zach-klippenstein: Oops, good catch.

deploy_website.sh
```diff
@@ -56,7 +56,7 @@ git checkout $SHA
 (cd kotlin && ./gradlew dokka)
 
 # Generate the Swift API docs.
-(cd swift/Samples/SampleApp && $SWIFT_DOCS_SCRIPT $SWIFT_API_DIR)
+($SWIFT_DOCS_SCRIPT $SWIFT_API_DIR)
```
zach-klippenstein: Doesn't the swift script need to be updated to use different relative paths?

RELEASING.md
dhavalshreyas: I had used `release-v0.0.x` for the "fix-branch". But maybe that's a very subtle difference :D

deploy_website.sh
dhavalshreyas: @amorde changed this, `Development.podspec` is now at the root.

RELEASING.md
dhavalshreyas: I added it in because we have like 4 `.podspec` files now :) 

But maybe we can do that on our own 🤷‍♂ 

I'll drop this.

dhavalshreyas:
```suggestion
```

RELEASING.md
zach-klippenstein: Ah, I missed that

deploy_website.sh
zach-klippenstein: Oh cool. So this has just been flat out wrong for a while?

amorde: Since https://github.com/square/workflow/pull/964

zach-klippenstein: This doesn't need the parentheses anymore – that was just there to scope the `cd`.
```suggestion
$SWIFT_DOCS_SCRIPT $SWIFT_API_DIR
```

AquaGeek:
If there are still inconsistencies, please open a PR against the [Swift repo](github.com/square/workflow-swift).

PR #1025, opened 2020-03-13
Require Signals to get wrapped in a Worker

AquaGeek:
This _should_ address #402. The aesthetics of the `SignalWorker` are very much open for feedback. We could also leave `context.subscribe` in place and create the worker in the infrastructure layer instead.

### DIFF
--- a/swift/Workflow/Sources/SignalWorker.swift
+++ b/swift/Workflow/Sources/SignalWorker.swift
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2020 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import ReactiveSwift
+
+/// A `Worker` that wraps a `Signal`
+public struct SignalWorker<Key: Equatable, Value>: Worker {
+    let key: Key
+    let signal: Signal<Value, Never>
+
+    public init(key: Key, signal: Signal<Value, Never>) {
+        self.key = key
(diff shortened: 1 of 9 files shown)


AquaGeek:
> Did you consider making `SignalWorker` abstract and having its own `run()` method that returns a `Signal`, mirroring `Worker`? Seems odd that the api for creating `Signal` workers vs `SignalProducer` workers is different, although I'm sure there's a good reason (I'm just curious what it is).

I'm not entirely sure I follow what you're suggesting. `SignalWorker` is a convenience wrapper — for `SignalProducer` Workers we currently require you to define them yourself. With `Signal`s I don't think it makes sense to define your own `Worker` type — it's a hot signal.

swift/WorkflowTesting/Sources/WorkflowRenderTester.swift
```diff
@@ -270,16 +270,14 @@ fileprivate final class RenderTestContext<T: Workflow>: RenderContextType {
         let sink = Sink<Action> { action in
             observer.send(value: AnyWorkflowAction(action))
         }
-        subscribe(signal: signal)
```
AquaGeek: Because `subscribe` has been removed I moved its logic up into this function and am subscribing directly — i.e. I'm not wrapping the `Signal` here in a worker (it creates test failures due to unexpected workers being encountered).

zach-klippenstein:
I don't see what the temperature of the stream has to do with it. Making it an extendable type means implementors can implement `isEquivalent`, vs passing in a key. That's the only real difference that I can tell.

swift/Workflow/Sources/SignalWorker.swift
```diff
@@ -0,0 +1,42 @@
...
+        self.key = key
+        self.signal = signal
+    }
+
+    public func run() -> SignalProducer<Value, Never> {
+        return SignalProducer(signal)
+    }
+
+    public func isEquivalent(to otherWorker: SignalWorker) -> Bool {
+        return key == otherWorker.key
```
AquaGeek: @bencochran Is `Key` comparison enough here? Do we run into the "need consumers to understand that `Key` needs to be differentiated enough" problem?

AquaGeek:
We're going to split this PR into two:
* Introduce the `SignalWorker`
* Remove the `subscribe` function from the `RenderContext`

That way, we can land the first, integrate it (incrementally), and fix any issues that come out of that.

JustinDSN:
Sounds great, incremental is better and less painful!

Issue #1021, opened 2020-03-11
Implement keyed side effects in place of / as underpinnings to Worker (Grand Unified Workflow Theory)


rjrjr:
There are lot of nagging questions and bugs around Workers, Workflows and their consistency between each other and the two platforms.

 * Should they have identity / key properties? (#961)
 * Should Swift have `SignalProducer.asWorker`? (#581)
 * Should we fix the bugs in `Context.subscribe`, or drop that function entirely? (#744)
 * How do we resolve their run / render APIs? (#1019, #996, #544)
 * How do we resolve their lifecycles? (#994)
 * Does Swift need StatelessWorkflow? (#236)

After a marathon discussion this afternoon, @zach-klippenstein @bencochran @dhavalshreyas and I are pretty convinced that most of these issues can be resolved, or at least become a lot simpler to address, if we drop Worker as a separate concept. We replace it with one new Workflow feature that allows them to do the same kind of work, `RenderContext.onFirstRender`. 

It should be possible to retrofit existing Workers as convenience implementations built on this call, allowing this to be a transparent change. And this seems like no more work than addressing all of the issues listed above in isolation — probably a lot less when you take into account the related design discussions that never seem to resolve.

I'm not going to try to capture all of the details, just enough that we don't forget the conversation. Counting on @zach-klippenstein and @bencochran to iterate here.

```kotlin
// Current:
override fun render(context: RenderContext<State, Output>): R {
  context.runningWorker(stream.asWorker(), “key”) {
    setOutput(it)
  }
}

// Proposed:
override fun render(
  props: RequestParameters,
  state: State,
  context: RenderContext<State, Output>
) {
  // key is required! No conveniences on this API, we use it to build simpler ones.
  context.onFirstRender(key = props + state) { onDispose ->
    val subscription = stream.subscribe {
      context.actionSink.send { action { … } }
    }
    onDispose { subscription.unsubscribe() }
  }
  context.onFirstRender(“times”) {}
}
```

It is not expected that anyone would call that method inline. Rather, it's the primitive we use to build various operators.

Worker could be reproduced something like this:

```kotlin
// <Props, State, Output, Rendering>
abstract class Worker<T> : Workflow<Worker<T>, T, Unit> {
  abstract fun run(): Flow<T>
  abstract fun isEquivalent(other: Worker<*>): Boolean

  final override asStatefulWorkflow() = WWF(this)
}

class WWF<T>(
  private val worker: Worker<T>
[...]

PR #1026, opened 2020-03-13
Bc/guwt prototype

zach-klippenstein:


swift/Workflow/Sources/RenderContext.swift
```diff
@@ -158,3 +152,107 @@ extension RenderContext {
...
+    typealias State = Void
+    func makeInitialState() -> State {
+        return ()
+    }
+    func workflowDidChange(from previousWorkflow: SignalProducerWorkflow, state: inout State) {
+    }
+    typealias Rendering = Void
+    func render(state: State, context: RenderContext<SignalProducerWorkflow>) -> Rendering {
+        let sink = context.makeSink(of: AnyWorkflowAction.self)
+        context.runSideEffect(key: UUID()) { [signalProducer] lifetime in
```
zach-klippenstein: Isn't this going to restart the `SignalProducer` on every render pass?

swift/Workflow/Sources/RenderContext.swift
```diff
@@ -158,3 +152,107 @@ extension RenderContext {
...
+                .map { value in
+                    AnyWorkflowAction { state in
+                        state = value
+                        return nil
+                    }
+                }
+                .observe(on: QueueScheduler.workflowExecution)
+                .observeValues(sink.send)
+        }
+        return property.value
```
zach-klippenstein: Should this be `state`?  Otherwise not sure why the state is updated on line 249.

swift/Workflow/Sources/Worker.swift
```diff
@@ -44,3 +44,48 @@ extension Worker where Self: Equatable {
...
+    func workflowDidChange(from previousWorkflow: WorkerWorkflow<WorkerType>, state: inout UUID) {
+        if !worker.isEquivalent(to: previousWorkflow.worker) {
+            state = UUID()
+        }
+    }
+
+    typealias Rendering = Void
+
+    func render(state: State, context: RenderContext<WorkerWorkflow>) -> Rendering {
+        context.runSideEffect(key: state) { sink, lifetime in
```
zach-klippenstein: I know you ended up removing this sink, but passing an explicit sink here would let us keep backpressure, as I raised in https://github.com/square/workflow/issues/1021#issuecomment-598976131.

swift/Workflow/Sources/RenderContext.swift
bencochran: Ah, yep! I copied too much from SignalWorkflow. Good catch.

swift/Workflow/Sources/RenderContext.swift
bencochran: Ah, yeah. Should definitely be `state`.

zach-klippenstein:
I think I should close this PR and let you open your own when you're ready @bencochran, you good with that?

bencochran:
👍 (confirmed offline)

PR #1031, opened 2020-03-19
Update mkdocs config with new Kotlin UI module structure.

zach-klippenstein:
Also fixes some other issues with the deployment script.

### DIFF
--- a/deploy_website.sh
+++ b/deploy_website.sh
@@ -53,10 +53,10 @@ pushd $DIR
 git checkout $SHA
 
 # Generate the Kotlin API docs.
-(cd kotlin && ./gradlew dokka)
+(cd kotlin && ./gradlew assemble && ./gradlew dokka)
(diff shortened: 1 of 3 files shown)


deploy_website.sh
```diff
@@ -53,10 +53,10 @@ pushd $DIR
 git checkout $SHA
 
 # Generate the Kotlin API docs.
-(cd kotlin && ./gradlew dokka)
+(cd kotlin && ./gradlew assemble && ./gradlew dokka)
 
 # Generate the Swift API docs.
-(cd swift/Samples/SampleApp && $SWIFT_DOCS_SCRIPT $SWIFT_API_DIR)
+$SWIFT_DOCS_SCRIPT $SWIFT_API_DIR
```
zach-klippenstein: This change duplicates one in #1024.

PR #1032, opened 2020-03-19
Revert "Enable AndroidX test orchestrator to try to reduce flakiness."

zach-klippenstein:
This reverts commit 18e4dcfdfa110289a077b7ca9a55b1caeb07d89d.

Tests are still super flaky, so no reason to use the orchestrator, since it
increases the time each test takes to run.


rjrjr:
I wonder if there's something wrong with our idle detection code.

zach-klippenstein:
@rjrjr The failure on this PR is that install exception again. See #1024.

zach-klippenstein:
Only Swift shards are red, all android shards are green, merging.

PR #1043, opened 2020-03-25
Add kotlin slack badge to README.

zach-klippenstein:
As documented [here](https://kotlinlang.slack.com/archives/C2YH04E3S/p1571313373000900).


zach-klippenstein:
Install exception.

PR #1051, opened 2020-03-30
Vend `preferredContentSize` through the `DescribedViewController`

zradke:
Because the `DescribedViewController` wraps all view controllers used in
workflows, if it does not vend its current child's
`preferredContentSize`, it will be lost. This makes it very difficult to
create dynamic workflows that rely on self-sizing view controllers
without introducing workarounds to have the children communicate their
intended size to the parent, or by removing the convenience of the
`DescribedViewController` entirely. This change solves the issue by
having the `DescribedViewController` update its own
`preferredContentSize` to match whatever the `currentViewController` is,
and when that child's size changes.

### DIFF
--- a/swift/WorkflowUI/Tests/DescribedViewControllerTests.swift
+++ b/swift/WorkflowUI/Tests/DescribedViewControllerTests.swift
@@ -1,0 +1,334 @@
+/*
+ * Copyright 2020 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


swift/WorkflowUI/Tests/DescribedViewControllerTests.swift
```diff
@@ -0,0 +1,318 @@
+//
```
zach-klippenstein: Nit: we don't use comment headers like this.

PR #1052, opened 2020-03-30
Replace Travis with Github Actions for Swift CI.

dhavalshreyas:
Sibling to #1053 

* Switch over all relevant builds to Github Actions
* Bumps Xcode to 11.3.1
* Bumps iOS version to 13.3


.github/workflows/swift.yaml
```diff
@@ -0,0 +1,41 @@
...
+  
+  build_docs:
+    runs-on: macOS-latest
+    needs: bundle
+    
+    steps:
+    - name: Documentation
+      run: |
+        brew install sourcedocs
+        .buildscript/build_swift_docs.sh /tmp/swiftdocs
```
zach-klippenstein: Does this job serve a purpose? Looks like this output isn't used at all. I'm planning to make a separate workflow to build and validate the entire doc site.

.github/workflows/swift.yaml
```diff
@@ -0,0 +1,41 @@
+name: Swift - Workflow - Cocoapods (Xcode 11.2.1)
+
+on: [push]
+
+jobs:
+  bundle:
```
zach-klippenstein: I don't think this job is going to do what you expect. Each job runs in its own VM instance. I think you need to set up a cache to get the output. See these docs:
- https://github.com/actions/cache/blob/master/examples.md#swift-objective-c---cocoapods
- https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows

.github/workflows/swift.yaml
dhavalshreyas: That makes sense. I'll skip it from this one then.

.github/workflows/swift.yaml
dhavalshreyas: Yes! Looks like we should use `artifacts` to carry data from one job to another.

dhavalshreyas: I was hoping for something like Google Cloud Builder's `workspace` which carries a dir from one job to the next.

zach-klippenstein: That's what the cache action does.

.github/workflows/swift.yaml
```diff
@@ -0,0 +1,45 @@
+name: Swift - Workflow - Cocoapods (Xcode 11.2.1)
+
+on: [push]
```
zach-klippenstein: You probably want to include a file filter so you don't run this job for Android PRs. See this doc: https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths

.github/workflows/swift.yaml
zach-klippenstein: Oh, never mind – i just read our travis config again and realized this was there to validate that the build actually passed. I guess we broke it at one point. We should leave this in then (but also copy the comment back over).

.github/workflows/swift.yaml
```diff
@@ -0,0 +1,99 @@
...
+on:
+  push:
+    branches:
+      - master
+      - '!gh-pages'
+    paths:
+      # Rebuild when workflow configs change.
+      - .github/workflows/**
+      # Or when swift code changes.
+      - swift/**
```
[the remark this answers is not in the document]
dhavalshreyas: Good point, I can also filter for the top-level files like `*.podspec`, `*.swift`. But ignoring paths is better.

.github/workflows/swift.yaml
```diff
@@ -0,0 +1,106 @@
...
+
+  spm:
+    runs-on: macOS-latest
+
+    steps:
+    - uses: actions/checkout@v1
+
+    - name: Swift Package Manager - iOS
+      run: |
+        xcodebuild -scheme "Workflow-Package" test -destination "name=iPhone 11"
```
AquaGeek: Shouldn't this be using SPM to build/test?

.github/workflows/swift.yaml
```diff
@@ -0,0 +1,106 @@
...
+        gem update --system
+        gem install bundler
+
+    - name: Bundle Install
+      run: bundle install
+
+    - name: Tutorial App
+      run: |
+        cd swift/Samples/Tutorial
+        bundle exec pod repo update
```
amorde: if this is taking a while the podfile could be updated to use `https://cdn.cocoapods.org` and remove the call to `pod repo update`

.github/workflows/swift.yaml
bencochran: Hmm, actually I don’t think we can since WorkflowUI depends on UIKit, which SPM doesn’t know about (in my experiment locally `swift test` works just fine but it skips all of the WorkflowUI tests because `#if canImport(UIKit)`.

.travis.yml
```diff
@@ -153,115 +141,3 @@ matrix:
...
-        - gem install bundler
-        - bundle install
-        - bundle exec pod gen Development.podspec
-
-    - language: swift
-      name: "Swift - SPM - iOS (Xcode 11.1)"
-      os: osx
-      osx_image: xcode11.1
-      script:
-        - xcodebuild -scheme "Workflow-Package" test -destination "name=iPhone 11"
```
dhavalshreyas: @AquaGeek, @amorde: This is the SPM test that I copied over to the Github Actions. Not entirely sure what's happening here, but my assumption was `xcodebuild` is using `Package.swift` to build and test the `Workflow` package.

.github/workflows/swift.yaml
bencochran: (Now that WorkflowUI is pretty useless without UIKit, we should probably make that a hard dependency…)

.github/workflows/swift.yaml
```diff
@@ -0,0 +1,106 @@
...
+    - name: Install
+      run: |
+        gem update --system
+        gem install bundler
+        bundle install
+        brew install sourcedocs
+        
+    - name: Swiftdocs
+      run: |
+        .buildscript/build_swift_docs.sh /tmp/swiftdocs
```
zach-klippenstein: Super nit (don't block the PR on this): `${{ runner.temp }}` [is a thing](https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#runner-context). Seems like the only benefit of using it is that it's guaranteed to be empty, which doesn't matter here.

.github/workflows/swift.yaml
```diff
@@ -0,0 +1,106 @@
...
+  push:
+    branches:
+      - master
+      - '!gh-pages'
+    paths:
+      - '*.podspec'
+      - 'Gemfile*'
+      - 'Package.swift'
+      - 'swift/**'
+      - '.github/workflows/swift.yaml'
```
zach-klippenstein: Just realized, this is missing `.buildscript/build_swift_docs.sh`

.travis.yml
amorde: that makes sense 👍

PR #1053, opened 2020-03-30
Replace Travis with Github Actions for Android CI.

zach-klippenstein:
This workflow (naming won't be confusing at all) will:
 * Only run if files are changed in the kotlin or github workflows directories.
 * Validate the gradle wrapper.
 * Install Java 8.
 * Share gradle's wrapper and cache.
   * Dependencies are shared between all runs with the same gradle config,
     but each SHA gets its own cache for artifacts.
 * Assemble all modules.
 * Run unit tests, checks, and instrumentation tests in parallel.
   * Instrumentation tests are ran on a macOS worker, all other shards are on ubuntu.
   * All check types are run in parallel.

This commit also removes the Android config from our .travis file.


.github/workflows/android.yml
```diff
@@ -0,0 +1,50 @@
...
+    - name: Build with Gradle
+      run: ./gradlew build
+  
+  instrumentation-tests:
+    name: Instrumentation tests
+    runs-on: macos-latest
+    timeout-minutes: 20
+    strategy:
+      matrix:
+        api-level: [21, 29]
```
dhavalshreyas: Neat!

zach-klippenstein: So much more useable than Travis' matrix functionality.

.github/workflows/android.yml
```diff
@@ -0,0 +1,52 @@
+name: Android CI
+
+on:
+  push:
+    branches: [ master ]
+    paths: kotlin/**
+  pull_request:
```
JustinDSN: Do you also need to add the `paths` here so these tasks are only ran for the `kotlin` directory or does it inherit the `push` settings?

JustinDSN: Oh you just added it.

.github/workflows/android.yml
```diff
@@ -0,0 +1,93 @@
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v2
+      - uses: gradle/wrapper-validation-action@v1
+      - name: set up JDK 1.8
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+      - uses: actions/cache@v1
```
dhavalshreyas: This is interesting. I'm not sure what is inside `~/.gradle/caches`, but could there be a race if there're multiple builds in-progress due to to multiple PRs. Could the cache from the `assemble` step from one PR be used in the `check` job in another.

dhavalshreyas: I think `Artifacts` is a better action for this: https://help.github.com/en/actions/configuring-and-managing-workflows/persisting-workflow-data-using-artifacts

zach-klippenstein: There are no races, caches are stored as tar files, and downloaded to the VM by this action before running your build. But yes, this cache is used so that the assemble job does all the common work needed by the other jobs. To get this to work, you seem to need both the cache configured with the same keys on all jobs, and have the right dependency graph configured via `needs`.

zach-klippenstein: I looked at a number of other open source Android projects, as well as the official docs for this action, and this seems to be the standard way of doing it.

zach-klippenstein:
A bunch of tests are failing but only for API 21, 29 works fine. I'm going to try adding a few more.

.github/workflows/android.yml
dhavalshreyas: Gotcha, the cache will be used across different PRs/branches..etc But I guess if `key` hashes the contents of `~/.gradle/caches`, we're all good.

zach-klippenstein: We're not hashing the contents of the cache. We're only hashing the contents of the build scripts. The reason for this, and I will document this in the script better, is so that if the build config changes, we start fresh to avoid any bad config polluting the cache. However, gradle does its own caching, so we don't need to clear out the cache if the build configs haven't changed.

zach-klippenstein:
This is ready to review again. I've also rebased/retargeted it directly to master, and removed our Travis config – don't see any reason to wait.

zach-klippenstein:
I wasn't actually getting the SHA, fixing…

.github/workflows/android.yml
```diff
@@ -0,0 +1,150 @@
+name: Android CI
+
+on:
+  push:
+    branches:
+      - master
+      - '!gh-pages'
+    paths:
+      # Rebuild when workflow configs change.
+      - .github/workflows/**
```
zach-klippenstein: Be more specific: only consider this actual file. No need to rebuild Android when iOS config changes.

.github/workflows/android.yml
```diff
@@ -0,0 +1,150 @@
+name: Android CI
```
zach-klippenstein: Rename to Kotlin (here and the file)

.github/workflows/android.yml
```diff
@@ -0,0 +1,150 @@
...
+        uses: actions/cache@v1
+        with:
+            path: ~/.gradle/caches
+            # Include the SHA in the hash so this step always adds a cache entry. If we didn't use the SHA, the artifacts
+            # would only get cached once for each build config hash.
+            # Don't use ${{ runner.os }} in the key so we don't re-assemble for UI tests.
+            key: gradle-${{ hashFiles('**/gradle/wrapper/gradle-wrapper.properties') }}-${{ hashFiles('**/*.gradle*') }}-${{ github.sha }}
+            # The first time a SHA is assembled, we still want to load dependencies from the cache.
+            # Note that none of jobs dependent on this one need restore keys, since they'll always have an exact hit.
+            restore-keys: |
```
rjrjr: This is like science fiction. 😺

rjrjr: I don't quite get what's going on w/the caching, but I don't see that any change is needed to the comments here. Just need to get myself up to speed.

.github/workflows/android.yml
```diff
@@ -0,0 +1,150 @@
...
+      - .github/workflows/**
+      - kotlin/**
+
+jobs:
+  assemble:
+    name: Assemble
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      # These setup steps should be common across all jobs in this workflow.
```
rjrjr: Kind of aggravating that these need to be copy / pasted all over the place. Oh well.

zach-klippenstein: Tell me about it.

.github/workflows/android.yml
zach-klippenstein: I'm not even sure caching is the right solution for per-run files. I think the cache is most useful for dependencies, things that don't change often across commits. But maybe we should use artifacts for communicating between jobs. Something we can play with after this merges.

zach-klippenstein:
UI test shard failed with a network issue, restarted (unfortunately you can't just restart a single job, you have to restart the whole workflow, [known issue](https://github.community/t5/GitHub-Actions/Ability-to-rerun-just-a-single-job-in-a-workflow/m-p/41629#M4691)).

PR #1060, opened 2020-03-31
Add a Github Workflow to automatically publish the doc site when release tags are pushed.

zach-klippenstein:
This workflow is run whenever a release tag is pushed (starts with `v` prefix), or when
a pull request is made that modifies the workflow, so the changes can be validated. In
the latter case, the `deploy_website.sh` script is run in dry mode so the docs are built
but not actually deployed. This workflow re-uses the gradle artifact and bundle caches
from the Kotlin and Swift workflows, respectively.

This requires a few changes to the `deploy_website.sh` script:
 - Use HTTPS instead of SSH to clone.
 - Allow custom username/password to be passed in via environment variables,
   so the GH Workflow can authenticate using its token.
 - Enable dry runs by setting the `DRY_RUN` environment variable, so pull requests that
   modify this workflow definition can still run the build steps in CI to validate them
   without actually publishing the doc site.
 - Shallow clone only the actual ref that is going to be deployed from.
 - Try to resolve the deploy ref with best-effort, but fallback to the raw value if it
   can't be resolved. Github does a shallow clone of the repo, so when building a pull
   request, the head ref won't actually be present in the worker's repo, so we just
   assume it will resolve when cloning fresh.
 - Build Kotlin source quietly.
 - Turn off auto-logging (`set -x`) since the git command might have a password in it
   that we don't want to expose.

This change also updates the `RELEASING` file to mention that the doc site is now built
and deployed automatically.

### DIFF
--- a/.github/workflows/deploy-doc-site.yml
+++ b/.github/workflows/deploy-doc-site.yml
@@ -1,0 +1,82 @@
+name: Deploy documentation site
+
+on:
+  push:
+    tags: [ v* ]
+  pull_request:
+    # Run this workflow as a dry run when it's edited, so we can test it.
+    paths: .github/workflows/deploy-doc-site.yml
+
+jobs:
+  deploy:
+    runs-on: macOS-latest
+    env:
+      # Default to deploying the current REF. If this is a pull request run, that ref will be invalid
+      # (local-only merge commit), so we'll override it with the GITHUB_HEAD_REF below.
+      DEPLOY_REF: ${{ github.ref }}
+    steps:
+      - uses: actions/checkout@v2
+
+      # Gradle caches (keys must match those defined in kotlin.yml)
+      # Don't use the gradle wrapper cache, since there's only one job we're downloading the whole wrapper once either way.
+      - name: Cache gradle artifacts
+        uses: actions/cache@v1
+        with:
+          path: ~/.gradle/caches
+          key: gradle-${{ hashFiles('**/gradle/wrapper/gradle-wrapper.properties') }}-${{ hashFiles('**/*.gradle*') }}-${{ github.sha }}
+          restore-keys: |
+            gradle-${{ hashFiles('**/gradle/wrapper/gradle-wrapper.properties') }}-${{ hashFiles('**/*.gradle*') }}-
+
+      # Swift caches (keys must match those defined in swift.yml)
+      - name: Load gem cache
+        uses: actions/cache@v1
+        with:
+          path: .bundle
+          key: gems-${{ hashFiles('Gemfile.lock') }}
+
+      - name: Set up Swift environment
+        run: |
+          # Set global bundle path so it gets used by build_swift_docs.sh running in the nested repo as well.
+          bundle config --global path "$(pwd)/.bundle"
+          bundle check || bundle install
+          # Don't need to run pod gen, the website script does that itself.
+          brew install sourcedocs
+          sudo xcode-select -s /Applications/Xcode_11.3.1.app
+
+      # Docs dependencies
+      - name: Set up Python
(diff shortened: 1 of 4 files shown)


zach-klippenstein:
PR-based runs work now.

zach-klippenstein:
I've been testing tag pushes using the tag `v0.26.0-docs.0`. When this PR lands, delete that tag and create a commit + tag for `v0.25.1-docs`.

.github/workflows/deploy-doc-site.yml
```diff
@@ -0,0 +1,82 @@
...
+          key: gems-${{ hashFiles('Gemfile.lock') }}
+
+      - name: Set up Swift environment
+        run: |
+          # Set global bundle path so it gets used by build_swift_docs.sh running in the nested repo as well.
+          bundle config --global path "$(pwd)/.bundle"
+          bundle check || bundle install
+          # Don't need to run pod gen, the website script does that itself.
+          brew install sourcedocs
+          sudo xcode-select -s /Applications/Xcode_11.3.1.app
```
dhavalshreyas, line 44: nit: We should probably move the xcodeversion to some global config, but we can clean it up later.

.github/workflows/deploy-doc-site.yml
```diff
@@ -0,0 +1,82 @@
...
+          echo DEPLOY_REF=$DEPLOY_REF
+          echo DRY_RUN=$DRY_RUN
+          git remote -v
+
+      ## Main steps
+      - name: Build and deploy website
+        env:
+          WORKFLOW_GOOGLE_ANALYTICS_KEY: ${{ secrets.WORKFLOW_GOOGLE_ANALYTICS_KEY }}
+          GIT_USERNAME: ${{ github.actor }}
+          GIT_PASSWORD: ${{ secrets.GITHUB_TOKEN }}
```
dhavalshreyas, line 81: Nice!

.github/workflows/deploy-doc-site.yml
```diff
@@ -0,0 +1,82 @@
...
+on:
+  push:
+    tags: [ v* ]
+  pull_request:
+    # Run this workflow as a dry run when it's edited, so we can test it.
+    paths: .github/workflows/deploy-doc-site.yml
+
+jobs:
+  deploy:
+    runs-on: macOS-latest
```
rjrjr, line 12: Outside of the scope of this PR, but I wonder if we really want to be using macOS-latest. Not hermetic.

.github/workflows/deploy-doc-site.yml
```diff
@@ -0,0 +1,82 @@
...
+      - uses: actions/checkout@v2
+
+      # Gradle caches (keys must match those defined in kotlin.yml)
+      # Don't use the gradle wrapper cache, since there's only one job we're downloading the whole wrapper once either way.
+      - name: Cache gradle artifacts
+        uses: actions/cache@v1
+        with:
+          path: ~/.gradle/caches
+          key: gradle-${{ hashFiles('**/gradle/wrapper/gradle-wrapper.properties') }}-${{ hashFiles('**/*.gradle*') }}-${{ github.sha }}
+          restore-keys: |
```
rjrjr, line 27: Is the `|` a yaml multi-line?

zach-klippenstein: Yep

zach-klippenstein:
Holding off on pushing the docs tag until I've fixed some issues with kdoc.

zach-klippenstein:
Pushed tag `v0.25.1-docs` to trigger doc site deploy.

https://github.com/square/workflow/runs/553365404?check_suite_focus=true

PR #1066, opened 2020-04-01
Fix Gradle config to generate correct multi-module dokka docs.

zach-klippenstein:
We are now using dokka in two distinct ways:
1. The maven publish plugin we're using runs the `dokka` task on each individual
   module to generate the javadocs jar maven artifact. We can use the default config for this.
2. The new `siteDokka` task generates Github-flavored Markdown files for all the modules
   we care to publish documentation for. This config is defined in the root gradle file.

This change means the URLs for kotlin api docs changes, so it also adds
redirects from some of the most-viewed Kotlin API pages to their new locations
in case links exist to those pages somewhere.

cc @dhavalshreyas for Github Workflows changes.


kotlin/build.gradle
```diff
@@ -224,3 +224,49 @@ subprojects {
...
+
+  configuration {
+    reportUndocumented = false
+    skipDeprecated = true
+    jdkVersion = 8
+
+    // TODO(#1063) Add source links using same automated process as subProjects.
+
+    perPackageOption {
+      prefix = "com.squareup.workflow.internal"
```
rjrjr, line 268: Oh, neat trick.

zach-klippenstein:
This is weird, the UI test runs failed because just restoring the gradle cache took 15 minutes: https://github.com/square/workflow/pull/1066/checks?check_run_id=552922994

PR #1074, opened 2020-04-02
Fix Kotlin build caching.

zach-klippenstein:
Gradle doesn't use the build cache by default, so we weren't actually caching any
build artifacts, just dependencies. This change turns on build caching for all
gradle invocations, and adds some logic to the kotlin workflow to wipe the build cache
before each fresh assemble run.

### DIFF
--- a/.github/workflows/kotlin.yml
+++ b/.github/workflows/kotlin.yml
@@ -14,6 +14,12 @@ on:
       - .github/workflows/kotlin.yml
       - kotlin/**
 
+env:
+  # Must use $HOME here, NOT a tilde, because of the order in which bash does expansion:
+  # Tilde happens before variables, so will be used literally, whereas $HOME will be
+  # recursively expanded.
+  GRADLE_CACHE_PATH: $HOME/.gradle/caches
+
 jobs:
   assemble:
     name: Assemble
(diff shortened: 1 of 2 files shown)


.github/workflows/kotlin.yml
```diff
@@ -29,23 +35,35 @@ jobs:
...
-              gradle-${{ hashFiles('**/gradle/wrapper/gradle-wrapper.properties') }}-${{ hashFiles('**/*.gradle*') }}-${{ hashFiles('**/buildSrc/**') }}-
+          path: ${{ env.GRADLE_CACHE_PATH }}
+          # Include the SHA in the hash so this step always adds a cache entry. If we didn't use the SHA, the artifacts
+          # would only get cached once for each build config hash.
+          # Don't use ${{ runner.os }} in the key so we don't re-assemble for UI tests.
+          key: gradle-${{ hashFiles('**/gradle/wrapper/gradle-wrapper.properties') }}-${{ hashFiles('**/*.gradle*') }}-${{ hashFiles('**/buildSrc/**') }}-${{ github.sha }}
+          # The first time a SHA is assembled, we still want to load dependencies from the cache.
+          # Note that none of jobs dependent on this one need restore keys, since they'll always have an exact hit.
+          restore-keys: |
+            gradle-${{ hashFiles('**/gradle/wrapper/gradle-wrapper.properties') }}-${{ hashFiles('**/*.gradle*') }}-${{ hashFiles('**/buildSrc/**') }}-
```
dhavalshreyas, line 49: What's in `buildSrc`?

.github/workflows/kotlin.yml
```diff
@@ -79,15 +96,15 @@ jobs:
...
         uses: actions/cache@v1
         with:
-            path: ~/.gradle/caches
-            # Don't set restore-keys so cache is always only valid for the current build config.
-            # Also don't use ${{ runner.os }} in the key so we don't re-assemble for UI tests.
-            key: gradle-${{ hashFiles('**/gradle/wrapper/gradle-wrapper.properties') }}-${{ hashFiles('**/*.gradle*') }}-${{ hashFiles('**/buildSrc/**') }}-${{ github.sha }}
+          path: ${{ env.GRADLE_CACHE_PATH }}
+          # Don't set restore-keys so cache is always only valid for the current build config.
+          # Also don't use ${{ runner.os }} in the key so we don't re-assemble for UI tests.
+          key: gradle-${{ hashFiles('**/gradle/wrapper/gradle-wrapper.properties') }}-${{ hashFiles('**/*.gradle*') }}-${{ hashFiles('**/buildSrc/**') }}-${{ github.sha }}
```
dhavalshreyas, line 102: I guess we don't need the `restore-key` here since the prior steps would have populated it?

.github/workflows/kotlin.yml
zach-klippenstein: Right now, just our dependency definitions.

.github/workflows/kotlin.yml
zach-klippenstein: Yes. This is documented on line 47.

PR #1076, opened 2020-04-02
Github Workflow to release Workflow version

dhavalshreyas:
Will be triggered by a Slack action. Currently updates and publishes from a "test" branch into a "test" release branch. Unfortunately, to test the `repository-dispatch` trigger, this will need to be merged to master.

### DIFF
--- a/.buildscript/update_changelog.swift
+++ b/.buildscript/update_changelog.swift
@@ -1,0 +1,33 @@
+#!/usr/bin/swift 
+
+import Foundation
+
+let contents = try! String(contentsOfFile: "CHANGELOG.md", encoding: .utf8)
+
+var lines = contents.components(separatedBy: "\n")
+
+let workflowVersion = ProcessInfo.processInfo.environment["WORKFLOW_VERSION"]!
+let kotlinChangelog = ProcessInfo.processInfo.environment["KOTLIN_CHANGELOG"]!
+let swiftChangelog =  ProcessInfo.processInfo.environment["SWIFT_CHANGELOG"]!
+
+let dateFormatter = DateFormatter()
+dateFormatter.dateFormat = "YYYY-MM-dd"
+
+let newChangelog = """
+
+## Version \(workflowVersion)
+
+_\(dateFormatter.string(from: Date()))_
+
+### Kotlin
+
+\(kotlinChangelog)
+
+### Swift
+
+\(swiftChangelog)
+"""
+
+lines.insert(newChangelog, at: 2)
+
+try! lines.joined(separator: "\n").write(toFile: "CHANGELOG.md", atomically: true, encoding: .utf8)
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,100 @@
+name: Release Workflow
+
+on:
+  repository_dispatch:
+    types: [release]
+
+env:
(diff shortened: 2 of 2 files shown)


zach-klippenstein:
Do we really need to introduce yet another language into this repo? We already have shell scripts and Ruby, can we use Ruby for this? I am concerned because the more languages we're using for config/automation, the harder it's going to be to onboard people, and the more painful it's going to be to work on this code since it's yet another dev environment we all need to have set up.

.buildscript/update_changelog.swift
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/swift 
+
+import Foundation
+
+let contents = try! String(contentsOfFile: "CHANGELOG.md", encoding: .utf8)
+
+var comps = contents.components(separatedBy: "\n")
```
zach-klippenstein, line 7: Nit: Please use the full word, it's more readable.

.buildscript/update_changelog.swift
```diff
@@ -0,0 +1,33 @@
...
+\(kotlinChangelog)
+
+### Swift
+
+\(swiftChangelog)
+"""
+
+comps.insert(newChangelog, at: 2)
+
+try! comps.joined(separator: "\n").write(toFile: "CHANGELOG.md", atomically: true, encoding: .utf8)
```
zach-klippenstein, line 33: What does `atomically` do and why is it necessary?

dhavalshreyas:
> There's a lot of logic in that workflow, and mostly just a lot of shell commands. Could it be written as an actual shell script that we just call from the workflow? I think an actual shell script would be easier to read/maintain than Github's custom workflow syntax, we'd get better editor support/syntax highlighting, and it would probably be easier to test if we can run the command manually/locally.

I actually like how the steps modularizes all the steps, will be easy for us to modify as we keep going. That + being able to visualize steps that succeeded / skipped / failed is pretty nice. I've been running github actions locally using this: https://github.com/nektos/act

.buildscript/update_changelog.swift
dhavalshreyas: Will fix. It's just lines, will call them lines.

.buildscript/update_changelog.swift
dhavalshreyas: `atomically` ensures a file either gets written successfully or fails without modifying (if app crashes, memory issues...etc). This way, we're ensured that the file doesn't end up in a corrupt state. It's slightly more expensive, but doesn't make a difference in this case.

zach-klippenstein:
None of that can't be done by shell scripts. Scripts can be modularized, use functions, use conditions, access environment variables, etc. Scripts can print debug and error "annotations" that will be extracted and printed specially by the actions UI, and of course anything printed to stdout is viewable live as the script runs.

I know shell scripts aren't ideal, but I still think shell scripts are better than yaml. I would rather have special logic like this in a scripting language that, while kinda gross, is still fairly universal and doesn't require special tools to run. 

That said, I don't feel super strongly about this. I just don't see any value provided by writing this all as a big workflow vs a shell script, and it seems like it will just make maintenance more of a pain.

zach-klippenstein:
Just to be extra clear, this is still WAY better than the manual process we have now. Thanks for doing this!

PR #1077, opened 2020-04-03
Migrate workflow-core module to multiplatform common code.

zach-klippenstein:
Picking up from #804, but taking a more cautious approach and migrating one module at a time.

This PR is two commits:
1. Converts the gradle config for `workflow-core` to use the Kotlin multiplatform plugin and config syntax, and renames the source directories appropriately (`main` -> `jvmMain`, etc).
2. Moves most of the source from the JVM target to the Common target, and introduces new common dependencies. The only things that couldn't be moved were some helpers for working with enums reflectively, since there's no way to reflect on enum constants in common code.

cc @atizik


zach-klippenstein:
Oops, forgot to move tests.

zach-klippenstein:
Tests are moved, and I added a bunch more!

kotlin/workflow-core/src/commonMain/kotlin/com/squareup/workflow/Snapshot.kt
```diff
@@ -136,22 +137,6 @@ fun BufferedSource.readByteStringWithLength(): ByteString {
   return readByteString(size.toLong())
 }
 
-inline fun <reified T : Enum<T>> BufferedSource.readOptionalEnumByOrdinal(): T? {
```
steve-the-edwards, line 139: Just removing these from the API?

steve-the-edwards: nvm.  I see it in jvm below.

zach-klippenstein:
Re-opened after some internal discussion about priority and scope. All good!

zach-klippenstein:
This PR breaks maven releases, and javadoc artifacts. Can't merge like this.

PR #1081, opened 2020-04-07
Porchbox: A sample app for porch delivery during COVID with Workflows

steve-the-edwards:
Upgrade gradle to 6.3-rc1 and AGP to 4.1.0-alpha04

Demonstrates a 'full-stack' app with a simple Firebase integration


kotlin/samples/containers/app-porchbox/src/test/java/com/squareup/sample/porchbox/PorchWorkflowTest.kt
```diff
@@ -0,0 +1,116 @@
...
+    workflow.testFromStart(Props) {
+      assertThat(awaitNextRendering().beneathModals is LoginRendering).isTrue()
+    }
+  }
+
+  @Test fun `test login loading`() {
+    PorchWorkflow(FakeAuthSuccess, FakeRealtimeDb).testFromState(
+        Props,
+        LoggingIn(BuildConfig.CLIENT_EMAIL, BuildConfig.CLIENT_PASS)) {
+      assertThat(awaitNextRendering().beneathModals).isEqualTo(LoggingIn(BuildConfig.CLIENT_EMAIL, BuildConfig.CLIENT_PASS))
```
steve-the-edwards, line 103: @zach-klippenstein this test is failing but I am having trouble understanding why.  Yes, I know that my fakeAuth and fakeDb will return immediately, but still there should be a Loading rendering *first* before the data is loaded (regardless of how quickly the Action is handled).  And yet, the result of awaitNextRendering() is always the rendering of the full data set Master Detail?  What am I missing?

.github/workflows/kotlin.yml
```diff
@@ -34,6 +34,17 @@ jobs:
         with:
           java-version: 1.8
 
+      - name: setup environment secrets
+        shell: bash
+        env: # pull in the secrets and environment variables
+          PORCHBOX_GOOGLE_SERVICES: $${{ secrets.GoogleServices }}
```
steve-the-edwards, line 40: Can an admin add these secrets for me if we are OK with this?  @zach-klippenstein @rjrjr

kotlin/samples/dungeon/app/build.gradle.kts
```diff
@@ -35,7 +35,7 @@ android {
 
   compileOptions {
     // Required for SnakeYAML.
-    coreLibraryDesugaringEnabled = true
+    isCoreLibraryDesugaringEnabled = true
```
steve-the-edwards, line 38: This was giving me a sync error without the change but perhaps it was a spurious gradle error?

steve-the-edwards:
Will put in a separate repo for the standalone app.

Issue #971, opened 2020-02-15
Extract common sample gradle config into shared file


zach-klippenstein:
If you think this will be the same for all apps you could just extract it into a separate `.gradle` file and apply to demo apps.

Something like:

./buildscript/configure-sample-app.gradle
```
apply plugin: 'com.android.application'

android {
  defaultConfig {
    testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
  }

  testOptions {
    animationsDisabled = true
  }
}

dependencies {
  androidTestImplementation deps.test.androidx.espresso.core
  androidTestImplementation deps.test.androidx.junitExt
  androidTestImplementation deps.test.androidx.uiautomator
}
```

and just apply that at the top:
```
apply from: rootProject.file('.buildscript/configure-sample-app.gradle')
apply plugin: 'kotlin-android'

android rootProject.ext.defaultAndroidConfig

android {
  defaultConfig {
......
```

_Originally posted by @tehras in https://github.com/square/workflow/pull/966_

wugeorgeq:
Is this still a good first issue if I want to start contributing? :)

PR #1082, opened 2020-04-07
Common sample app gradle config

wugeorgeq:
Taking a stab at [#971 ](https://github.com/square/workflow/issues/971)

### DIFF
--- a/kotlin/.buildscript/android-sample-app.gradle
+++ b/kotlin/.buildscript/android-sample-app.gradle
@@ -1,0 +1,9 @@
+apply from: rootProject.file('.buildscript/configure-android-defaults.gradle')
+
+dependencies {
+  implementation(project(":workflow-core"))
+  implementation(project(":workflow-runtime"))
+
+  implementation(Deps.get("androidx.appcompat"))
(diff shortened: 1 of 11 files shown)


wugeorgeq:
Good call, will do

wugeorgeq:
It looks like we need to write an implementation of `Plugin` to achieve the same goal in kotlin dsl. Going from groovy -> kotlin means going from a dynamic resolution of the scripts to kotlin's static accessor definitions `android{}`, `implementation()` etc. TLDR it's not as simple as converting the common gradle files to `gradle.kts`. @zach-klippenstein, this seems like a good change to do, just wanted to check in to see if this is what we want https://proandroiddev.com/sharing-build-logic-with-kotlin-dsl-203274f73013

wugeorgeq:
Spike here: https://github.com/square/workflow/pull/1086

zach-klippenstein:
Just noticed this is two commits: each commit should be a complete change, so if the first one just adds a file without using it anywhere it looks like dead code. Please squash them, and then we can merge this!

Issue #959, opened 2020-02-12
RenderTester API should not require explicitly expecting all workers 


zach-klippenstein:
[Internal Slack discussion](https://square.slack.com/archives/C1MDTM2FR/p1581085074142600?thread_ts=1581033959.140600&cid=C1MDTM2FR):
> Feature request, by the by: in terms of future-proofing the tests,  can it be that "unexpected" renders and workers are treated as non-triggering no-ops?  I see some advantage to having to make them explicit, but it does mean my test is now sensitive to changes for code paths I'm explicitly not testing...

We could make workers “expected” by default, but we can’t automatically expect child workflows, since `renderChild` must always return a rendering and the test library can’t know how to construct your childrens’ renderings for you.

wugeorgeq:
Is this still a good first issue if I want to start contributing? :)

zach-klippenstein:
Thanks for putting up a PR @wugeorgeq! I just noticed it now, my bad. Would like to get agreement from a few more maintainers before merging it though. cc @rjrjr @bencochran @dhavalshreyas

rjrjr:
If we make this change, do we have a way to express "expect this worker and no other"? That seems important.

bencochran:
Maybe we could instead make this opt-in? Like a “allow unexpected workers” flag of sorts? That would a) make the default still break if you go from no workers to workers unexpectedly and b) give a way to expect some but ignore other workers.

zach-klippenstein:
So we agree this is a good option to have, it's an open question which is the better default behavior.

I think the default should be more lenient. The original reason this came up was because someone added a worker to a workflow, and immediately a bunch of completely unrelated tests broke. I don't think that's a good UX, and if we make strictness opt-out, nobody's going to remember to do that and we will only have solved half the problem. Strictly expecting "no other workers" is useful but is a specific assertion, and more brittle, and so it should be opt-in. 

For comparison, Mockito (popular java mocking library) has similar functionality for verifying methods calls on a mock by either calling the `verifyNoMoreInteractions` for the catch-all case — it is also opt-in.

rjrjr:
I'm fine with lenient by default, so long as I have a way to turn
strictness on.

wugeorgeq:
Sorry for the late reply, I'll make an attempt unless there are objections :)

wugeorgeq:
> The fact that we had to break out a separate function for workers is weird. Why does it make sense to allow leniency for workers but not for workflows?

[...]

PR #1084, opened 2020-04-07
Make expected workers optional in RenderTester

wugeorgeq:
Addressing [#959](https://github.com/square/workflow/issues/959)

### DIFF
--- a/kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RealRenderTester.kt
+++ b/kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RealRenderTester.kt
@@ -117,7 +117,7 @@ internal class RealRenderTester<PropsT, StateT, OutputT : Any, RenderingT>(
     key: String,
     handler: (ChildOutputT) -> WorkflowAction<StateT, OutputT>
   ): ChildRenderingT {
-    val expected = consumeExpectation<ExpectedWorkflow<*, *>>(
+    val expected = consumeExpectedChildWorkflow<ExpectedWorkflow<*, *>>(
         predicate = { it.workflowType.isInstance(child) && it.key == key },
         description = {
           "child workflow ${child::class.java.name}" +
@@ -144,7 +144,7 @@ internal class RealRenderTester<PropsT, StateT, OutputT : Any, RenderingT>(
     key: String,
     handler: (T) -> WorkflowAction<StateT, OutputT>
   ) {
-    val expected = consumeExpectation<ExpectedWorker<*>>(
+    val expected = consumeExpectedWorker<ExpectedWorker<*>>(
         predicate = { it.matchesWhen(worker) && it.key == key },
         description = {
           "worker $worker" +
@@ -154,7 +154,7 @@ internal class RealRenderTester<PropsT, StateT, OutputT : Any, RenderingT>(
         }
     )
 
-    if (expected.output != null) {
+    if (expected?.output != null) {
       check(processedAction == null)
       @Suppress("UNCHECKED_CAST")
       processedAction = handler(expected.output.output as T)
@@ -187,7 +187,7 @@ internal class RealRenderTester<PropsT, StateT, OutputT : Any, RenderingT>(
     }
   }
 
-  private inline fun <reified T : Expectation<*>> consumeExpectation(
+  private inline fun <reified T : Expectation<*>> consumeExpectedChildWorkflow(
     predicate: (T) -> Boolean,
     description: () -> String
   ): T {
@@ -211,6 +211,29 @@ internal class RealRenderTester<PropsT, StateT, OutputT : Any, RenderingT>(
     consumedExpectations += expected
 
     return expected
+  }
+
+  private inline fun <reified T : Expectation<*>> consumeExpectedWorker(
+    predicate: (T) -> Boolean,
+    description: () -> String
+  ): T? {
+    val matchedExpectations = expectations.filterIsInstance<T>()
+        .filter(predicate)
+    return when (matchedExpectations.size) {
+      1 -> {
+        val expected = matchedExpectations[0]
+        // Move the workflow to the consumed list.
+        expectations -= expected
+        consumedExpectations += expected
+
+        expected
+      }
+      0 -> null
+      else -> throw AssertionError(
+          "Multiple expectations matched ${description()}:\n" +
+              matchedExpectations.joinToString(separator = "\n") { "  $it" }
+      )
+    }
   }
 
   private fun deepCloneForRender(): RenderContext<StateT, OutputT> = RealRenderTester(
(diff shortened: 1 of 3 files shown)


kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RealRenderTester.kt
```diff
@@ -213,6 +213,29 @@ internal class RealRenderTester<PropsT, StateT, OutputT : Any, RenderingT>(
     return expected
   }
 
+  private inline fun <reified T : Expectation<*>> consumeExpectedWorker(
```
wugeorgeq, line 216: My kotlin is weak here i think i can define this more explicitly `private inline fun <reified T : ExpectedWorker<*>> consumeExpectedWorker`

wugeorgeq:
Hi all :) I'm on the customers team at square, btw

kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RealRenderTester.kt
```diff
@@ -213,6 +213,29 @@ internal class RealRenderTester<PropsT, StateT, OutputT : Any, RenderingT>(
...
   }
 
+  private inline fun <reified T : ExpectedWorker<*>> consumeExpectedWorker(
+    predicate: (T) -> Boolean,
+    description: () -> String
+  ): T? {
+    val matchedExpectations = expectations.filterIsInstance<T>()
+        .filter(predicate)
+    return when (matchedExpectations.size) {
+      1 -> {
```
rjrjr, line 223: The fact that we had to break out a separate function for workers is a little weird. Why does it make sense to allow leniency for workers but not for workflows?

rjrjr: Answered here: https://github.com/square/workflow/issues/959#issuecomment-631001737

kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RealRenderTester.kt
```diff
@@ -213,6 +213,29 @@ internal class RealRenderTester<PropsT, StateT, OutputT : Any, RenderingT>(
     return expected
   }
 
+  private inline fun <reified T : ExpectedWorker<*>> consumeExpectedWorker(
```
zach-klippenstein, line 216: This method needs kdoc explaining why it returns null, and/or maybe a different name that makes that clear.

kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RealRenderTester.kt
```diff
@@ -154,7 +154,7 @@ internal class RealRenderTester<PropsT, StateT, OutputT : Any, RenderingT>(
         }
     )
 
-    if (expected.output != null) {
+    if (expected?.output != null) {
```
zach-klippenstein, line 157: Would be helpful to have a comment here explaining what `expected == null` implies.

kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RealRenderTester.kt
```diff
@@ -213,6 +213,29 @@ internal class RealRenderTester<PropsT, StateT, OutputT : Any, RenderingT>(
...
+    return when (matchedExpectations.size) {
+      1 -> {
+        val expected = matchedExpectations[0]
+        // Move the workflow to the consumed list.
+        expectations -= expected
+        consumedExpectations += expected
+
+        expected
+      }
+      0 -> null
```
zach-klippenstein, line 231: Nit: This ordering is a bit weird, would read more naturally to me to have the 0 case first.

kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RealRenderTester.kt
```diff
@@ -213,6 +213,29 @@ internal class RealRenderTester<PropsT, StateT, OutputT : Any, RenderingT>(
...
+    val matchedExpectations = expectations.filterIsInstance<T>()
+        .filter(predicate)
+    return when (matchedExpectations.size) {
+      1 -> {
+        val expected = matchedExpectations[0]
+        // Move the workflow to the consumed list.
+        expectations -= expected
+        consumedExpectations += expected
+
+        expected
```
zach-klippenstein, line 229: Nit: It looks really weird to just have a variable name sitting by itself. This should have have a `return`.

kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RealRenderTester.kt
```diff
@@ -213,6 +213,29 @@ internal class RealRenderTester<PropsT, StateT, OutputT : Any, RenderingT>(
...
+      }
+      0 -> null
+      else -> throw AssertionError(
+          "Multiple expectations matched ${description()}:\n" +
+              matchedExpectations.joinToString(separator = "\n") { "  $it" }
```
zach-klippenstein, line 234: Nit: This is error message formatting is duplicated, it could be factored out.

kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RealRenderTester.kt
```diff
@@ -187,7 +187,7 @@ internal class RealRenderTester<PropsT, StateT, OutputT : Any, RenderingT>(
     }
   }
 
-  private inline fun <reified T : Expectation<*>> consumeExpectation(
+  private inline fun <reified T : ExpectedWorkflow<*, *>> consumeExpectedChildWorkflow(
```
zach-klippenstein, line 190: Both these functions share almost all their logic. The only difference is what happens in the 0-expectations-matched case. This could have been kept as a single function with a function parameter for that behavior, and would make it a lot more clear that they are essentially the same function.

wugeorgeq:
Oh I thought we were going to make this opt in perhaps with a defaulted parameter in `RealRenderTester` constructor, should I do a fast-follow PR?

wugeorgeq:
```internal class RealRenderTester<PropsT, StateT, OutputT : Any, RenderingT>(
  private val workflow: StatefulWorkflow<PropsT, StateT, OutputT, RenderingT>,
  private val props: PropsT,
  private val state: StateT,
  private val expectations: MutableList<Expectation<*>> = mutableListOf(),
  private val consumedExpectations: MutableList<Expectation<*>> = mutableListOf(),
  private var childWillEmitOutput: Boolean = false,
  private var processedAction: WorkflowAction<StateT, OutputT>? = null,
  private val allowUnexpectedWorkers: Boolean = true
)```

kotlin/workflow-testing/src/main/java/com/squareup/workflow/testing/RealRenderTester.kt
wugeorgeq: I was playing around with using kotlinreflect's typeOf method but having a function parameter makes more sense now that you mention it. I think the other detail is that the function originally returned nonnullable but now would it return null? Is that awkward for `renderChild` and `runningWorker`?  would we uitilize a `requireNotNull` in `renderChild`?

zach-klippenstein:
This is what I meant: https://github.com/square/workflow/pull/1172

PR #1088, opened 2020-04-10
Makes BackStackContainer.update protected, required for customization.

rjrjr:


zach-klippenstein:
You just need to run `gw apiDump`.

PR #1092, opened 2020-04-14
Add ContainerViewController init with AnyWorkflowConvertable

dhavalshreyas:
Currently, we don't have an API to create a `ContainerViewController` with `AnyWorkflow` type. This adds an extension to make that possible. 

This is mostly a copy-paste from code @bencochran shared with me privately.

### DIFF
--- a/swift/WorkflowUI/Sources/Container/ContainerViewController+AnyWorkflow.swift
+++ b/swift/WorkflowUI/Sources/Container/ContainerViewController+AnyWorkflow.swift
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2019 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#if canImport(UIKit)
+
+import Foundation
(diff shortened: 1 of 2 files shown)


swift/WorkflowUI/Sources/Container/ContainerViewController+AnyWorkflow.swift
```diff
@@ -0,0 +1,39 @@
+import Foundation
+import Workflow
+
+extension ContainerViewController {
```
bencochran, line 4: You’ll need to wrap it all in `#if canImport(UIKit)` since ContainerViewController has that.

dhavalshreyas: Done!

JustinDSN:
Should we add a unit test to make sure this works as expected, and stays working?

bencochran:
> Should we add a unit test to make sure this works as expected, and stays working?

Yeah, that’d probably be a good idea, especially since this relies on Swift knowing that the call to `self.init(workflow:…` is the other (more-specifically-constrained) initializer and not a recursive call to self.

swift/WorkflowUI/Sources/Container/ContainerViewController+AnyWorkflow.swift
```diff
@@ -0,0 +1,56 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#if canImport(UIKit)
+
+    import Foundation
```
bencochran, line 19: nit: we don’t indent inside other `#if` blocks

dhavalshreyas:
> Should we add a unit test to make sure this works as expected, and stays working?

Done!

bencochran:
Test looks good, thanks!

PR #1090, opened 2020-04-14
Modal Container with Modal Sample App

asthatrivedi:
This PR has all the changes addressing the PR i had created a few days ago. Git is not letting me reopen the earlier closed PR since i used force-push in order to fix some commit history issue. I had seen that issue before and have finally found the root cause. I do not foresee this being a problem anymore. I will add everyone's respective comments to make things easier for everyone. For reference, here's the older closed [PR](https://github.com/square/workflow/pull/1087). 

PR includes:

1. Adding ModalContainerScreen and ModalContainerViewController to open source adding a podspec file for the same.
2. Presenting modal view controller as a "confirm to quit" screen to TicTacToe app. 
3. All the PR feedback from the previous PR which i will mention at each line for reference.

PR #1095, opened 2020-04-16
Bump Cocopods Version

JustinDSN:
* Bumps the local development versions of Cocoapods to current version.
* Bumps podspecs to 1.7.0

### DIFF
--- a/swift/Samples/Tutorial/Frameworks/Tutorial1Complete/Tutorial1.podspec
+++ b/swift/Samples/Tutorial/Frameworks/Tutorial1Complete/Tutorial1.podspec
@@ -8,7 +8,7 @@ Pod::Spec.new do |s|
   s.source       = { git: 'Not Published', tag: "podify/#{s.version}" }
 
   # 1.7 is needed for `swift_versions` support
-  s.cocoapods_version = '>= 1.7.0.beta.1'
+  s.cocoapods_version = '>= 1.7.0'
 
   s.swift_versions = ['5.0']
   s.ios.deployment_target = '11.0'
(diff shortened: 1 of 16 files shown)


swift/Samples/Tutorial/Frameworks/Tutorial1Complete/Tutorial1.podspec
```diff
@@ -8,7 +8,7 @@ Pod::Spec.new do |s|
   s.source       = { git: 'Not Published', tag: "podify/#{s.version}" }
 
   # 1.7 is needed for `swift_versions` support
-  s.cocoapods_version = '>= 1.7.0.beta.1'
+  s.cocoapods_version = '>= 1.7.0'
```
dhavalshreyas, line 11: Should this be `1.9.1`?

asthatrivedi: Adding some more info relevant to this upgrade. I'm quoting @amorde here from the modal sample [PR](https://github.com/square/workflow/pull/1090)  

> It's worth keeping something in s.cocoapods_version in case someone tries to pull in workflow while using an old CocoaPods installation, but it definitely doesn't need to be a beta. 1.7.0 should work

JustinDSN: Thanks for adding the relevant comment @asthatrivedi! If we set it to that version, we'd force developers to update their version of cocoapods to use our library. We should try and use the minimum required version where possible.

dhavalshreyas: Makes sense, not sure what's happening with the builds.

amorde: Workflow should support any CocoaPods above `1.7.0` so this LGTM

PR #925, opened 2020-01-30
[swift] Add SplitScreenContainer Sample

JustinDSN:
* Port of the `WorkflowSplitScreenContainer` to open source
* Adds Snapshot Tests (using `iOSSnapshotTestCase` pod)
* Runs Snapshot Tests on Travis
* Improvements:
  * Made `SplitScreenContainerViewController` and `SplitScreenContainerScreen` generic.
  * Handle changes of the ratio in an animated fashion
  * Take `separatorColor` as part of `SplitScreenContainerScreen`
  * Removes `Ratio` in favor of `CGFloat`
  * Updated access levels of members

SplitScreen-ChangeScreen

PR #1098, opened 2020-04-19
[swift] [rfc] Add TabBarScreen Sample

JustinDSN:
* Adds a TabBarScreen sample with snapshot tests
* Demonstrates a way to separate BarItem renderings as a separate Workflow to isolate contained tab screen from bar badge value fetch / calculations.

* Note: Compose Screen concept is out of scope for this PR.

TODO & Open Questions:
- [ ] Should we hang on to only the currently displayed rendering (current implementation) or should we accept an array of `AnyScreen`s and use selected index to control display?
  * Either implementation is technically correct. If we hang on to each screen, then the parent workflow would have to continue to render those workflows each render pass (just like the BackStackScreen).

Single Screen vs. Multiple Screens:
Pros for Single Screen:
  * You don't render screens early (eg: tabs that haven't been visited yet)  

Cons for Single Screen: 
  * View state could be would between renderings (eg: Scroll Position)

- [ ] How do we want to demonstrate fetching / rendering the badge value for the Baz tab? A separate Workflow or a simple Worker that is awaited in `DemoWorkflow`?
  * I'm leaning towards a Worker as its simpler and easier to grok, however I do like demonstrating the map rendering here. WDYT?

- [ ] Determine the right height for the tab bar on all the devices, should this come in on the screen or is it an implementation detail?
   * Also focus more in on the safe area insets thing.

- [ ] Should we snip the ContainerView?

- [x] Any way to tighten up the API (I assume less verbosity)?


swift/Samples/TabBarContainer/DemoApp/Bar/BarScreen.swift
```diff
@@ -0,0 +1,73 @@
...
+
+fileprivate final class BarScreenViewController: ScreenViewController<BarScreen> {
+    
+    private lazy var titleLabel: UILabel = .init()
+    private lazy var tapGestureRecognizer: UITapGestureRecognizer = .init()
+    
+    required init(screen: BarScreen, environment: ViewEnvironment) {
+        super.init(screen: screen, environment: environment)
+        
+        update(with: screen)
```
asthatrivedi, line 39: From what i understand screenDidChange() will now be called right after init so this update function call is not needed anymore.

swift/Samples/TabBarContainer/DemoApp/Bar/BarWorkflow.swift
```diff
@@ -0,0 +1,88 @@
...
+    typealias Output = Never
+
+}
+
+
+// MARK: State and Initialization
+
+extension BarWorkflow {
+
+    struct State {
```
asthatrivedi, line 35: If State is going to be empty then this can be of type Void.

swift/Samples/TabBarContainer/DemoApp/Bar/BarWorkflow.swift
```diff
@@ -0,0 +1,88 @@
...
+
+// MARK: Actions
+
+extension BarWorkflow {
+
+    enum Action: WorkflowAction {
+
+        case viewTapped
+
+        typealias WorkflowType = BarWorkflow
```
asthatrivedi, line 57: Doesn't look like WorkFlowType is being used. Can we remove it or use it?

swift/Samples/TabBarContainer/DemoApp/Foo/FooWorkflow.swift
```diff
@@ -0,0 +1,88 @@
...
+    typealias Output = Never
+
+}
+
+
+// MARK: State and Initialization
+
+extension FooWorkflow {
+
+    struct State {
```
asthatrivedi, line 35: Same comment mentioned in BarWorkflow

swift/Samples/TabBarContainer/DemoApp/Bar/BarWorkflow.swift
amorde: This is for conforming to `WorkflowAction`
https://github.com/square/workflow/blob/8d879b0ebd58d11d7495dc3dd01901c98df26b5a/swift/Workflow/Sources/WorkflowAction.swift#L20-L23

swift/Samples/TabBarContainer/Sources/TabBarScreenViewController.swift
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    override internal func viewDidLoad() {
+        super.viewDidLoad()
+
+        view.addSubview(containerView)
+        view.addSubview(tabBar)
+
+        addChild(contentViewController)
+        containerView.contentView.addSubview(contentViewController.view)
```
bencochran, line 76: Why containerView instead of adding the contentViewController’s view directly?

swift/Samples/TabBarContainer/Sources/TabBarScreenViewController.swift
```diff
@@ -0,0 +1,137 @@
...
+        contentViewController.didMove(toParent: self)
+        
+        update()
+    }
+    
+    internal override func viewDidLayoutSubviews() {
+        super.viewDidLayoutSubviews()
+        
+        let tabBarHeight: CGFloat = 83
```
bencochran, line 85: Can we ask the tab bar it’s height? And how does this work across iPhone X-ish sized phones with bottom safe area insets and iPhone 8-ish ones without them?

swift/Samples/TabBarContainer/SnapshotTests/TabBarContainerSnapshotTests.swift
```diff
@@ -0,0 +1,147 @@
...
+            screen: makeTabBarScreen(),
+            environment: .empty
+        )
+
+        viewController.view.layoutIfNeeded()
+
+        FBSnapshotVerifyView(
+            viewController.view,
+            identifier: name,
+            suffixes: ["_64"]
```
bencochran, line 42: What’s the `_64`for?

swift/Samples/TabBarContainer/DemoApp/Foo/FooWorkflow.swift
```diff
@@ -0,0 +1,79 @@
...
+    typealias State = Void
+
+}
+
+
+// MARK: Actions
+
+extension FooWorkflow {
+
+    enum Action: WorkflowAction {
```
bencochran, line 44: Do we need this action?

swift/Samples/TabBarContainer/DemoApp/DemoWorkflow.swift
```diff
@@ -0,0 +1,122 @@
...
+        }
+        
+        let barBarItem = BarBadgeWorkflow()
+            .mapOutput { (output) -> Action in
+                switch output {
+                case .barTabSelected:
+                    return .barTabSelected
+                }
+        }
+        .rendered(with: context)
```
bencochran, line 105: nit: indent the `}` and `rendered`

swift/Samples/TabBarContainer/DemoApp/Bar/BarScreen.swift
```diff
@@ -0,0 +1,71 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import Workflow
+import WorkflowUI
+
+
+struct BarScreen: Screen {
```
bencochran, line 20: I’d maybe avoid “Bar” as the name here, since it’s easy to confuse with the bar from “TabBar”

swift/Samples/TabBarContainer/DemoApp/Bar/BarBadgeWorkflow.swift
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+    }
+
+}
+
+// MARK: Rendering
+
+extension BarBadgeWorkflow {
+    typealias Rendering = BarItem
```
bencochran, line 100: If I were writing something that had a workflow for its badge value, I’d probably render out just the badge value itself rather than a whole `BarItem`. As it, this ties this workflow to how its consumer is displaying it (e.g. if later we wanted to move this to a sidebar style nav, I’d have to go change this “feature code”)

swift/Samples/TabBarContainer/Sources/TabBarScreen.swift
```diff
@@ -0,0 +1,41 @@
...
+        return TabBarScreenContainerViewController.description(for: self, environment: environment)
+    }
+}
+
+public struct BarItem {
+
+    public var title: String
+    public var image: UIImage
+    public var selectedImage: UIImage?
+    public var badge: String?
```
bencochran, line 30: I know UITabBarItem uses a `String?` for this, but I’d maybe consider modeling this a little more? Like
```swift
enum Badge {
    case none
    case value(Int)
    case text(String)
}
```

I dunno, might just be based on my aversion to `String?` as a type (i.e. what’s the functional difference between `nil` and `""`?)

JustinDSN: I chose to keep this the same semantic as what UITabBarItem uses. Although, I like your idea. 

What number formatter would we use when `.value` is used?

swift/Samples/TabBarContainer/Sources/TabBarScreenViewController.swift
JustinDSN: I received feedback on my SplitScreen PR from @AquaGeek to use a container view to codify a view who sets it's subviews bounds to its frame on every layout pass. Decided to keep the pattern the same.

bencochran: Can’t find that discussion from that PR. But I feel like this adds an unnecessary layer to the code.

swift/Samples/TabBarContainer/Sources/TabBarScreen.swift
bencochran: I’d probably just use `String(value)`

swift/Samples/TabBarContainer/Sources/TabBarScreenViewController.swift
JustinDSN: https://github.com/square/workflow/pull/925#discussion_r376552966

swift/Samples/TabBarContainer/DemoApp/Bar/BarBadgeWorkflow.swift
JustinDSN: Good suggestion, I wanted to get this type of feedback so I thought I'd throw an option out there.

swift/Samples/TabBarContainer/DemoApp/Foo/FooWorkflow.swift
JustinDSN: Nope removing both in Bar and Foo.

swift/Samples/TabBarContainer/DemoApp/Bar/BarScreen.swift
JustinDSN: Will rename to my next favorite name "Baz".

swift/Samples/TabBarContainer/DemoApp/Bar/BarBadgeWorkflow.swift
JustinDSN: Do you like to see a separate Workflow for the BadgeValue or would you just rather use a Worker that is launched in the DemoWorkflow? I'm leaning towards the Worker that lives in the DemoWorkflow.

swift/Samples/TabBarContainer/Sources/TabBarScreenViewController.swift
```diff
@@ -0,0 +1,137 @@
...
+        }
+        
+        tabBar.setItems(tabBarItems, animated: false)
+        tabBar.selectedItem = selectedTabBarItem
+    }
+
+    override internal func viewDidLoad() {
+        super.viewDidLoad()
+
+        view.addSubview(containerView)
```
dhavalshreyas, line 72: Any reason we shouldn't use `UITabBarController` here?

swift/Samples/TabBarContainer/Sources/TabBarScreenViewController.swift
```diff
@@ -0,0 +1,137 @@
...
+        view.addSubview(tabBar)
+
+        addChild(contentViewController)
+        containerView.contentView.addSubview(contentViewController.view)
```
dhavalshreyas, line 76: `UITabBarController` takes in an array of `UIViewController`. When we switch tabs, they're not removed from memory. 

In our version, it looks like we're keeping only the active `UIViewController` in memory. When we switch tabs, we'll be creating new VCs.

Is this the desired behavior?

swift/Samples/TabBarContainer/DemoApp/DemoWorkflow.swift
```diff
@@ -0,0 +1,143 @@
...
+
+
+// MARK: Rendering
+
+extension DemoWorkflow {
+
+    typealias Rendering = TabBarScreen<AnyScreen>
+
+    func render(state: State, context: RenderContext<DemoWorkflow>) -> Rendering {
+        let sink = context.makeSink(of: Action.self)
```
dhavalshreyas, line 78: This seems like a lot of code to render the tab bar view. Anything we can do to tighten this up?

swift/Samples/TabBarContainer/DemoApp/Bar/BarWorkflow.swift
```diff
@@ -0,0 +1,88 @@
...
+
+    }
+
+    func makeInitialState() -> BarWorkflow.State {
+        return State()
+    }
+
+    func workflowDidChange(from previousWorkflow: BarWorkflow, state: inout State) {
+
+    }
```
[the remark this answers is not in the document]
JustinDSN: Done :)

swift/Samples/TabBarContainer/Sources/TabBarScreenViewController.swift
JustinDSN: I've seen two internal implementations that use BluePrint to render the tab bar *and* only accept the currently displayed screen. 

This allows greater control of what the UI rendering / view behavior looks like. For this sample, I didn't want to use Blueprint, so decided to use a simple UITabBar. 

The question of holding onto an array of multiple screens is a good one and should be discussed in more detail.  I'd like a few more folks to weigh their opinions on this PR before we circle up as a group to decide.

swift/Samples/TabBarContainer/DemoApp/DemoWorkflow.swift
dhavalshreyas: Explored a similar direction here: https://github.com/square/workflow/pull/1106/files#diff-09f70fe452e73d20c303e8566efc935bR87

JustinDSN: Accepted @dhavalshreyas's suggestions.

swift/Samples/TabBarContainer/Sources/TabBarScreenViewController.swift
JustinDSN: Accepted Dhaval's suggestion and we're continuing to render all the screens in the parent workflow, so you won't loose your position.

swift/Samples/TabBarContainer/Sources/TabBarContainerScreen.swift
```diff
@@ -0,0 +1,81 @@
...
+        self.selectedImage = selectedImage
+        self.badge = badge
+    }
+
+}
+
+extension BarItem {
+    public enum Badge {
+        case none
+        case value(Int)
```
AquaGeek, line 67: Why a separate case for this? Why not force consumers to do the text conversion?

AquaGeek:
> How do we want to demonstrate fetching / rendering the badge value for the Baz tab? A separate Workflow or a simple Worker that is awaited in DemoWorkflow?

Can we use a map function to convert the `Output` of the hosted workflows to a badge value? e.g.:

* `MessageListWorkflow` has an output of "unread messages," which we map to a badge value when rendering it into a tab screen.
* `SettingsWorkflow` has an output of "available updates," which we also map to a badge value.

This way, if we reparent these workflows into a different container (e.g. hamburger menu side bar) we don't have to change their content at all.

swift/Samples/TabBarContainer/Sources/TabBarScreenViewController.swift
AquaGeek: For context, IIRC that discussion was around adding static container subviews in the split screen so that you don't have to redo all the sizing math when swapping out contents — you just hand the new content view to the appropriate container and it adds it as a subview and does a `view.frame = bounds`.

bencochran:
I’d say badge value should be part of rendering of children, not outputs (i.e. it’s their state, not events).

swift/Samples/TabBarContainer/Sources/TabScreen.swift
```diff
@@ -0,0 +1,39 @@
...
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+import Foundation
+import WorkflowUI
+
+public struct TabScreen {
+    public let barItem: BarItem
+    public let content: AnyScreen
+    public let onSelect: (() -> ())
```
JustinDSN, line 22: Snip this, and handle the selection at the parent.

swift/Samples/TabBarContainer/DemoApp/Baz/BazWorkflow.swift
```diff
@@ -0,0 +1,90 @@
...
+        func apply(toState state: inout BazWorkflow.State) -> BazWorkflow.Output? {
+            return .tabSelected
+        }
+    }
+}
+
+// MARK: Rendering
+
+extension BazWorkflow {
+    typealias Rendering = TabScreen
```
JustinDSN, line 56: This should be just BazScreen

swift/Samples/TabBarContainer/DemoApp/Baz/BazWorkflow.swift
```diff
@@ -0,0 +1,90 @@
...
+
+}
+
+// MARK: State and Initialization
+
+extension BazWorkflow {
+
+    typealias State = Void
+    enum Output {
+        case tabSelected
```
JustinDSN, line 33: Create an worker output for the badgeCountUpdate.

JustinDSN:
> I’d say badge value should be part of rendering of children, not outputs (i.e. it’s their state, not events).

@bencochran could you clarify this a little more. Do you mean that you'd rather see the rendering be a tuple of screen and badge count?

JustinDSN:
Closing for now - will re-open post repo move and when I have time to wrap up.

Issue #744, opened 2019-11-12
[iOS] If 2 signals fire in the same run loop, the second is ignored


bizzemfrog:
If a workflow is subscribed to 2 signals, when both signals send a value in the same runloop, the second signal's value is ignored.

What I have observed is that the both of the signal's map closures get called (ie `context.subscribe(signal: signal.map { Foo -> Action in ... })`) but only the fist emitted action gets `apply` called on it. 

Under the hood I see both actions get enqueued in the `WorkflowNode.SubtreeManager.Subscriptions`' Pipe but once the fist action is applied, the context is reset, which invalidates the `Subscriptions`' Pipe, causing the queued event to get dropped.

I wouldn't expect a signal's value get ignored, and I wouldn't expect developers to ensure that signals don't fire in the same run loop. Seems like subscription events should get queued in a system apart from the current render context.

dhavalshreyas:
@bizzemfrog pushed a branch with a failing test to help investigate this issue:  https://github.com/square/workflow/commit/ee284f3b1bcd2c19daee4791e5ca3af0a357b79d

To help with exploration, I have a rough version of the fix: https://github.com/square/workflow/commit/970931366218627d803d8b38f5060e5509513fa3

I'm now trying to figure out what the right/clean fix for this should be. @davidapgar has some excellent analysis and documentation for a similar issue here: https://github.com/square/workflow/issues/348

I'll be digging into this next week, meanwhile if anybody has any ideas/suggestions, it'll be much appreciated 🙏🏼

AquaGeek:
#1025 Should fix this

dhavalshreyas:
#1048 fixes this.

PR #1111, opened 2020-04-27
Remove context.subscribe

AquaGeek:
This is a rebase of #1025. It removes `context.subscribe` altogether, now that we've migrated our internal clients. **This is a breaking change.**

This finishes the work on #402 / #744.

### DIFF
--- a/swift/Workflow/Sources/SubtreeManager.swift
+++ b/swift/Workflow/Sources/SubtreeManager.swift
@@ -36,9 +36,6 @@ extension WorkflowNode {
         /// The current array of workers
         private (set) internal var childWorkers: [AnyChildWorker] = []
 
-        /// Subscriptions from the outside world.
-        private var subscriptions: Subscriptions = Subscriptions(eventSources: [], eventPipe: EventPipe())
-
         init() {}
 
         /// Performs an update pass using the given closure.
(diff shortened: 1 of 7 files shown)


swift/Workflow/Sources/SubtreeManager.swift
```diff
@@ -470,31 +458,6 @@ extension WorkflowNode.SubtreeManager {
 }
 
 
-// MARK: - Subscriptions
-
-extension WorkflowNode.SubtreeManager {
-    fileprivate final class Subscriptions {
```
JustinDSN, line 476: Bye bye bye 👋

JustinDSN:
The commits are a little weird. Do we need all of them? Should we squash them into a single commit?

swift/Workflow/Sources/SubtreeManager.swift
dhavalshreyas: The red lines make me happy :D

AquaGeek:
> The 8 commits are a little weird. Do we need all of them? Should we squash them into a single commit?

We don't need all of them. Some got pulled out into the other PR that introduced `SignalWorker`. I tried rebasing and squashing some of them, but there were conflicts. I'll just squash all of them.

PR #1113, opened 2020-04-27
Add dependency update as a change type to the pull request template

zach-klippenstein:
Also fixes a typo.


dhavalshreyas:
🤔 Why didn't this PR have a template? 

Another idea, if we have a checklist item for "Should highlight in changelog", the release bot can automatically pick it up.

zach-klippenstein:
It did have a template, I just deleted it since most of it was irrelevant since this isn't a change to actual code.

I like the changelog idea!

PR #1118, opened 2020-04-30
Update Compose to dev10.

zach-klippenstein:
# Description

Dev10 finally brings support for coroutines!

This is the first compose update in a while to actually bump the compiler version
as well as the plugin version, which is why it requires a few more changes
to work.
Fixes # (issue)

## Type of change

Please select at least one

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] API change this is not Backward compatible (fix or feature that would cause existing functionality to not work as expected or change the API contract)
- [ ] This change requires a documentation update
- [x] Dependency update

## Testing Checklist
How was this change tested?

- [ ] Unit Tests
- [x] UI Tests

## Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] Any dependent changes have been merged and published

### DIFF
--- a/kotlin/samples/hello-compose-binding/build.gradle.kts
+++ b/kotlin/samples/hello-compose-binding/build.gradle.kts
@@ -1,3 +1,5 @@
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
 /*
  * Copyright 2019 Square Inc.
  *
@@ -28,6 +30,9 @@ android {
 }
 
 apply(from = rootProject.file(".buildscript/configure-compose.gradle"))
+tasks.withType<KotlinCompile> {
+  kotlinOptions.apiVersion = "1.3"
+}
 
 dependencies {
   implementation(project(":workflow-ui:core-compose"))
--- a/kotlin/samples/hello-compose-binding/src/androidTest/java/com/squareup/sample/hellocomposebinding/HelloBindingTest.kt
+++ b/kotlin/samples/hello-compose-binding/src/androidTest/java/com/squareup/sample/hellocomposebinding/HelloBindingTest.kt
@@ -15,8 +15,8 @@
  */
 package com.squareup.sample.hellocomposebinding
 
-import androidx.test.ext.junit.rules.ActivityScenarioRule
 import androidx.test.ext.junit.runners.AndroidJUnit4
+import androidx.ui.test.android.AndroidComposeTestRule
 import androidx.ui.test.assertIsDisplayed
 import androidx.ui.test.doClick
 import androidx.ui.test.findByText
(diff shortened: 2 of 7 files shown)


kotlin/samples/hello-compose-binding/src/androidTest/java/com/squareup/sample/hellocomposebinding/HelloBindingTest.kt
```diff
@@ -28,7 +28,7 @@ import org.junit.runner.RunWith
 class HelloBindingTest {
 
   // Launches the activity.
-  @Rule @JvmField val scenario = ActivityScenarioRule(HelloBindingActivity::class.java)
+  @Rule @JvmField val composeRule = AndroidComposeTestRule<HelloBindingActivity>()
```
zach-klippenstein, line 31: Compose top-level UI test functions now require a `ComposeTestRule` to be used, or they will throw.

kotlin/workflow-ui/core-compose/src/main/java/com/squareup/workflow/ui/compose/ComposeViewFactory.kt
```diff
@@ -84,7 +85,7 @@ internal class ComposeViewFactory<RenderingT : Any>(
     )
 
     // Entry point to the composition.
-    composeContainer.setContent {
+    composeContainer.setContent(Recomposer.current()) {
```
zach-klippenstein, line 88: Old method is deprecated.

kotlin/samples/hello-compose-binding/build.gradle.kts
```diff
@@ -28,6 +30,9 @@ android {
 }
 
 apply(from = rootProject.file(".buildscript/configure-compose.gradle"))
+tasks.withType<KotlinCompile> {
+  kotlinOptions.apiVersion = "1.3"
+}
```
zach-klippenstein, line 35: The compiler the plugin is built against is in some weird liminal state between 1.3 and 1.4, so these warnings are safe to ignore, this suppresses them.

rjrjr: Seems like this comment should be in the gradle file.

zach-klippenstein: Yep

zach-klippenstein: Done.

zach-klippenstein:
Looks like this might be an actual test failure, but the Actions UI isn't even showing a log for this task so I github might just be having issues.

zach-klippenstein:
UI test failure is now: `Caused by: java.net.UnknownHostException: dl.google.com`.

PR #1131, opened 2020-05-04
Fix broken kotlin documentation build, include in CI.

zach-klippenstein:
This is a known issue with Dokka 0.10.x: https://github.com/Kotlin/dokka/issues/656

## Checklist

- [ ] Unit Tests
- [ ] UI Tests

### DIFF
--- a/.github/workflows/kotlin.yml
+++ b/.github/workflows/kotlin.yml
@@ -63,7 +63,7 @@ jobs:
       ## Actual task
       - name: Assemble with gradle
         working-directory: ./kotlin
-        run: ./gradlew assemble --build-cache --no-daemon --stacktrace
+        run: ./gradlew assemble --build-cache --no-daemon --stacktrace --info
 
   # Runs all check tasks in parallel.
   check:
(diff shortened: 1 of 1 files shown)


zach-klippenstein:
I'm also seeing this with Dokka 0.10.0. But only sometimes. Now it's working locally, without the build cache. Might be a caching issue.

zach-klippenstein:
The CI error looks like it's caused by the gradle cache not being saved to Actions cache, and since `dokka` doesn't depend on `assemble`, it's crashing.

zach-klippenstein:
Seems like Actions stopped recursively expanding environment variables, so `$HOME/.gradle/caches` is being interpreted as a literal path.

zach-klippenstein:
New workflow isn't running, maybe because this PR didn't target master? Rebasing and trying again…

zach-klippenstein:
Yep that seems to do it. I'll merge to master and then cherry-pick into the release branch.

zach-klippenstein:
Now `${{ env.HOME }}` is evaluating to an empty string. Putting this on the back-burner for now,.

zach-klippenstein:
Dokka is now running in CI as of #1153. The issue with caching is tracked as #1152.

PR #1134, opened 2020-05-05
Add os_signpost logs for render() and Workflow and Worker lifecycle

### LINKED ISSUE #1083: Add support for os_signpost
The `os_signpost` APIs introduced in iOS 12 allow us to provide more visibility into the internals of the library. I built a prototype of this (see `stromberg/os_signpost` branch), but it needs to be cleaned up before getting reviewed.

AquaGeek:
Fixes #1083

We are logging:
* `Workflow` creation and destruction*
* `Worker` creation and destruction*
* `Workflow` `render()` calls
* `Sink` events — **note: The format strings specified for these does not include the "public" specifier, so they shouldn't show up in logs unless a debugger is attached.** This is important, as you don't want to e.g. log password entry in text fields.

*Note: We're not instrumenting the `Workflow`s or `Worker`s themselves; we're instrumenting the infrastructure that gets created to hold onto them.

```swift
// If you don't subsequently render this workflow, it won't show up in Instruments:
let workflow = MyWorkflow()
```


AquaGeek:
This is what it looks like in the Tic Tac Toe sample app:

os_signpost

AquaGeek:
```
Testing failed:
	'init(subsystem:category:)' is only available in iOS 10.0 or newer
	'init(subsystem:category:)' is only available in iOS 10.0 or newer
	Testing cancelled because the build failed.

** TEST FAILED **
```

@dhavalshreyas Do we need to bump the spm tests to iOS 10?

swift/Workflow/Sources/SubtreeManager.swift
```diff
@@ -433,11 +439,36 @@ extension WorkflowNode.SubtreeManager {
...
+
             signalProducer
                 .take(during: lifetime)
                 .observe(on: QueueScheduler.workflowExecution)
-                .startWithValues { [weak self] output in
-                    self?.handle(output: output)
+                .start { [weak self] event in
+                    switch event {
+                    case .value(let output):
+                        self?.handle(output: output)
```
JustinDSN, line 455: Does it make sense to log an event to indicate a worker emitted a value and it's about to be handled?

swift/Workflow/Sources/Log.swift
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import os.signpost
+
+extension OSLog {
```
JustinDSN, line 19: Nit: Should this file be called OSLog+Workflow or OSLog+LogSubsystems or something other than just "Log"?

JustinDSN:
> ```
> Testing failed:
> 	'init(subsystem:category:)' is only available in iOS 10.0 or newer
> 	'init(subsystem:category:)' is only available in iOS 10.0 or newer
> 	Testing cancelled because the build failed.
> 
> ** TEST FAILED **
> ```
> 
> @dhavalshreyas Do we need to bump the spm tests to iOS 10?

JustinDSN:
Sorry about the false close, I accidentally hit the button when I tried to cancel a comment.

dhavalshreyas:
> @dhavalshreyas Do we need to bump the spm tests to iOS 10?

Yes! Bumped it up here #1135

swift/Workflow/Sources/SubtreeManager.swift
dhavalshreyas: Maybe move the logging to a different function. `logWorkerState(event:)`?

swift/Workflow/Sources/SubtreeManager.swift
```diff
@@ -213,7 +214,14 @@ extension WorkflowNode.SubtreeManager {
 
             let reusableSink = sinkStore.findOrCreate(actionType: Action.self)
 
+            let signpostRef = NSObject()
+
             let sink = Sink<Action> { action in
+                if #available(iOS 12.0, *) {
```
dhavalshreyas, line 220: This check can get noisy. How do you feel about creating a convenience wrapper (maybe `workflow_os_signpost`) which wraps this check and the `os_signpost` call?

dhavalshreyas:
This is awesome @AquaGeek! Thank you for getting this in!

swift/Workflow/Sources/SubtreeManager.swift
AquaGeek: I know. 😞 I went down that road, but the APIs for `os_signpost` unfortunately don't let you do that: https://github.com/square/workflow/issues/1083#issuecomment-623675549

The message arguments in the `os_signpost` calls get translated into `CVarArg...` in Swift. Apple would need to expose one that supports passing a `va_list`; there's no way to "splat" the resulting array in the wrapper back into variadic arguments.

The only other thing I can think of is to add very specific functions (e.g. `logWorkflowStart/End`) to our wrapper.

dhavalshreyas: Makes sense, ignore this comment then!

swift/Workflow/Sources/SubtreeManager.swift
AquaGeek: I'm exploring using `.on(…)` in the stream (similar to what `.logEvents` does). I'll update this soon.

AquaGeek:
OK. Everything is on this branch now.

I think the last open question is: should we remove the event logging (e.g. sink and worker events) or make them more generic for now? One option would be to log the related Workflow/Worker type instead of the event type (as that's essentially useless, e.g. `Action`).

JustinDSN:
> OK. Everything is on this branch now.
> 
> I think the last open question is: should we remove the event logging (i.e. sink and worker events) or make them more generic for now? One option would be to log the related Workflow/Worker type instead of the event type (as that's essentially useless, e.g. `Action`).

I would vote for making them more generic "Action" along with a comment and a follow up task to log the more detail once we get the sign off from Mosec?

AquaGeek:
> > OK. Everything is on this branch now.
> > I think the last open question is: should we remove the event logging (i.e. sink and worker events) or make them more generic for now? One option would be to log the related Workflow/Worker type instead of the event type (as that's essentially useless, e.g. `Action`).
> 
> I would vote for making them more generic "Action" along with a comment and a follow up task to log the more detail once we get the sign off from Mosec over removing the logging altogether.

To clarify, when I say remove logging I mean remove event logging (for sink events and worker outputs). All the lifecycle stuff can stay.

swift/Workflow/Sources/SubtreeManager.swift
dhavalshreyas: Did the `.on(..)` work out?

swift/Workflow/Sources/SubtreeManager.swift
```diff
@@ -204,7 +204,11 @@ extension WorkflowNode.SubtreeManager {
         func makeSink<Action>(of actionType: Action.Type) -> Sink<Action> where Action: WorkflowAction, WorkflowType == Action.WorkflowType {
             let reusableSink = sinkStore.findOrCreate(actionType: Action.self)
 
+            let signpostRef = NSObject()
```
dhavalshreyas, line 207: Wonder if we should create a type which has the `log...` events and acts as a `ref` instead of creating an arbitrary object to act as `ref`.

AquaGeek: 🤷 I'm not overly-concerned about it. We could make the `log...` calls member functions instead of static functions (and use the object as the object we use to look up the signpost ID. We'd have to figure out how to plumb that object around to the various places that need it.

I think this is fine for now. Thoughts?

dhavalshreyas: I meant, creating something like:

```swift
class SignpostRef: AnyObject {
  func log(...)
}
```

But, this is just a nit, feel free to ship as-is!

JustinDSN:
Regarding the actions, I'm fine with either way as long as someone takes DRI to try and get the approval to add them in. I think it's very valuable, and likely that we'll get approval. My vote would be to keep them in with a comment to the followup work.

AquaGeek:
I changed the event logging to log the `Workflow` or `Worker` type instead of the event information.

Pros:
* Can be public — we already log it with the other signposts
* Easier to distinguish where an event came from (before you had to do the mapping yourself from event type -> `Workflow`/`Worker` type

Cons:
* Less information

PR #1137, opened 2020-05-06
Extract all the os_signpost calls to a single place

AquaGeek:
Another way around the limitations in the `os_signpost` APIs is to create specific logging calls and do all the integration with `os_signpost` behind the scenes (instead of trying to create a wrapper for it).

### DIFF
--- a/swift/Workflow/Sources/OSLog+Workflow.swift
+++ b/swift/Workflow/Sources/OSLog+Workflow.swift
@@ -16,7 +16,82 @@
 
 import os.signpost
 
-extension OSLog {
+fileprivate extension OSLog {
     static let workflow = OSLog(subsystem: "com.squareup.Workflow", category: "Workflow")
     static let worker = OSLog(subsystem: "com.squareup.Workflow", category: "Worker")
+}
+
+// MARK: -
+
+final class WorkflowLogger {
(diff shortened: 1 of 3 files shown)


swift/Workflow/Sources/OSLog+Workflow.swift
```diff
@@ -16,7 +16,82 @@
...
     static let workflow = OSLog(subsystem: "com.squareup.Workflow", category: "Workflow")
     static let worker = OSLog(subsystem: "com.squareup.Workflow", category: "Worker")
 }
+
+// MARK: -
+
+final class WorkflowLogger {
+    // MARK: Workflows
+
+    static func logWorkflowStarted<WorkflowType>(ref: WorkflowNode<WorkflowType>) {
```
amorde, line 29: this has the side effect of losing some information about the call site (I think `os_log` keeps track of `#file` and `#line`, but it seems like we don't really care about that so LGTM

dhavalshreyas:
I like this approach, less noisy, the methods make it clear what we're logging. Not sure how we'd have used the file/line info though, if it doesn't add a lot of value, fine with dropping it.

AquaGeek:
Merging into the other branch. Please stand by.

PR #1148, opened 2020-05-10
Simplify ViewRegistry to act more like a simple map of rendering type to ViewFactory.

zach-klippenstein:
`ViewRegistry` is now just responsible for returning a `ViewFactory` for a given
rendering type, as well as checking that views returned from the factory have been
correctly bound with `bindShowRendering`. `buildView` has been extracted into an
extension method on `ViewRegistry`.

This change simplifies the responsibility of each `ViewRegistry` implementation. It
also makes `ViewRegistry` more flexible for extension. For example, allows a
Compose-based `ViewFactory` to detect recursive Compose bindings and stay in Compose
instead of jumping back out into the legacy view layer every time.

## Checklist

- [x] Unit Tests
- [x] UI Tests

### DIFF
--- a/kotlin/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/ViewFactory.kt
+++ b/kotlin/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/ViewFactory.kt
@@ -27,8 +27,8 @@ import kotlin.reflect.KClass
  *
  * Sets of bindings are gathered in [ViewRegistry] instances.
  */
-interface ViewFactory<RenderingT : Any> {
-  val type: KClass<RenderingT>
+interface ViewFactory<in RenderingT : Any> {
+  val type: KClass<in RenderingT>
 
   /**
    * Returns a View ready to display [initialRendering] (and any succeeding values)
--- a/kotlin/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/ViewRegistry.kt
+++ b/kotlin/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/ViewRegistry.kt
@@ -67,22 +67,23 @@ interface ViewRegistry {
   val keys: Set<KClass<*>>
 
   /**
-   * It is usually more convenient to use [WorkflowViewStub] than to call this method directly.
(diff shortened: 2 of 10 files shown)


kotlin/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/ViewFactory.kt
```diff
@@ -27,8 +27,8 @@ import kotlin.reflect.KClass
  *
  * Sets of bindings are gathered in [ViewRegistry] instances.
  */
-interface ViewFactory<RenderingT : Any> {
-  val type: KClass<RenderingT>
+interface ViewFactory<in RenderingT : Any> {
```
rjrjr, line 30: Now that's a good sign.

kotlin/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/ViewRegistry.kt
```diff
@@ -67,22 +67,23 @@ interface ViewRegistry {
   val keys: Set<KClass<*>>
 
   /**
-   * It is usually more convenient to use [WorkflowViewStub] than to call this method directly.
+   * This method is for general use, use [WorkflowViewStub] instead.
```
rjrjr, line 70: "is not for"?

kotlin/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/ViewRegistry.kt
```diff
@@ -67,22 +67,23 @@ interface ViewRegistry {
...
    *
-   * Creates a [View] to display [initialRendering], which can be updated via calls
-   * to [View.showRendering].
+   * Returns the [ViewFactory] that was registered for the given [renderingType].
    *
-   * @throws IllegalArgumentException if no binding can be find for type [RenderingT]
+   * @throws IllegalArgumentException if no factory can be find for type [RenderingT]
```
rjrjr, line 74: can be _found_

zach-klippenstein:
UI test failures are installation network exceptions.

PR #1150, opened 2020-05-11
Remove RxJava2 dependency from Workflow Android UI integration.

### LINKED ISSUE #1123: Convert workflow-ui/core-android from Rx to Flow
I don't think we're relying on any advanced Rx features that Flow doesn't support, so we should be able to convert it and drop the heavy RxJava dependency.

zach-klippenstein:
Closes #1123.

## Checklist

- [x] Unit Tests
- [ ] UI Tests

### DIFF
--- a/kotlin/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/WorkflowFragment.kt
+++ b/kotlin/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/WorkflowFragment.kt
@@ -21,11 +21,10 @@ import android.view.ViewGroup
 import androidx.fragment.app.Fragment
 import com.squareup.workflow.Workflow
 import com.squareup.workflow.ui.WorkflowRunner.Config
-import io.reactivex.Flowable
 
 /**
  * A [Fragment] that can run a workflow. Subclasses implement [onCreateWorkflow]
- * to configure themselves with a [Workflow], [ViewRegistry] and [inputs][Flowable].
+ * to configure themselves with a [Workflow], [ViewRegistry] and [inputs][Config.props].
  *
  * For a workflow with no inputs, or a static configuration, that's as simple as:
  *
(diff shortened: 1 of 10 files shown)


kotlin/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/WorkflowFragment.kt
```diff
@@ -21,11 +21,10 @@ import android.view.ViewGroup
 import androidx.fragment.app.Fragment
 import com.squareup.workflow.Workflow
 import com.squareup.workflow.ui.WorkflowRunner.Config
-import io.reactivex.Flowable
 
 /**
  * A [Fragment] that can run a workflow. Subclasses implement [onCreateWorkflow]
- * to configure themselves with a [Workflow], [ViewRegistry] and [inputs][Flowable].
+ * to configure themselves with a [Workflow], [ViewRegistry] and [inputs][Config.props].
```
rjrjr, line 27: Oops.

kotlin/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/WorkflowRunnerViewModel.kt
```diff
@@ -64,28 +66,32 @@ internal class WorkflowRunnerViewModel<OutputT : Any>(
...
         .map { it.snapshot }
         .onEach { lastSnapshot = it }
         .launchIn(scope)
+
+    // Cancel the entire workflow runtime after the first output is emitted.
+    // Use the Unconfined dispatcher to ensure the cancellation happens as immediately as possible.
+    scope.launch(Dispatchers.Unconfined) {
+      result.join()
+      scope.cancel(CancellationException("WorkflowRunnerViewModel delivered result"))
+    }
```
zach-klippenstein, line 87: @rjrjr FYI I just reworked this a bit. The result deferred is now created using `async`, which is much more straightforward and means we don't have to manually propagate results and exceptions. The cancel-on-output logic is now responsible _only_ for that, which I think makes it easier to read.

Issue #1152, opened 2020-05-11
Github Actions isn't working with the gradle cache anymore


zach-klippenstein:
Github seems to have stopped recursively expanding environment variables, so our `GRADLE_BUILD_CACHE` variable that is set to `$HOME/.gradle/caches` gets resolved literally as `$HOME/.gradle/caches`, which is obviously not the real path.

Furthermore, `${{ env.HOME }}` resolves to the empty string, so I'm not sure how to get the home directory anymore.

Some exploration of what broke was done in #1131.

Kiskae:
You might want to consider turning the problem around and defining a custom `GRADLE_USER_HOME` instead of using the default `$HOME/.gradle`, then caching `$GRADLE_USER_HOME/caches`. If the caching mechanism is path-relative it might even be possible to isolate it to the current build instead of using the user environment.

zach-klippenstein:
I agree that would be ideal, but I couldn't figure out how to specify a custom cache path when running Gradle. I only saw `--project-cache-dir` which I tried, but it didn't seem to affect all the caches. There's also `--gradle-user-home` which I missed the first time. I'll try again, one of those should definitely work.

Kiskae:
`--project-cache-dir` seems to control where the local build artifacts (that dont end up in `/build`) go (`<projectDir>/.gradle`), while `--gradle-user-home` controls the global wrapper/module cache location. (`<HOME>/.gradle`)

zach-klippenstein:
https://github.com/square/workflow-kotlin-compose/pull/30

PR #1153, opened 2020-05-11
Validate dokka in CI.

zach-klippenstein:
This is a weird way to do it, but it works around #1152.

## Checklist

- [ ] Unit Tests
- [ ] UI Tests

### DIFF
--- a/.github/workflows/kotlin.yml
+++ b/.github/workflows/kotlin.yml
@@ -65,6 +65,12 @@ jobs:
         working-directory: ./kotlin
         run: ./gradlew assemble --build-cache --no-daemon --stacktrace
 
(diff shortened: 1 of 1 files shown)


.github/workflows/kotlin.yml
```diff
@@ -65,6 +65,12 @@ jobs:
         working-directory: ./kotlin
         run: ./gradlew assemble --build-cache --no-daemon --stacktrace
 
+      # This should ideally be done in the Check job below, but until gradle caching is fixed we
+      # need to do it after assembling.
+      - name: Run dokka to validate kdoc
+        working-directory: ./kotlin
+        run: ./gradlew dokka --build-cache --no-daemon --stacktrace
```
zach-klippenstein, line 72: Run `siteDokka` as well.

PR #1154, opened 2020-05-11
Annotate fatalError()s with Workflow type

### LINKED ISSUE #968: `fatalError`s need to be annotated with the associated workflow
In cases where we `fatalError`, we need to attach additional debugging information (e.g. the associated workflow), because the call stack is unhelpful (bunch of closures all the way down). A good example of where this happens is `WorkflowNode.SubtreeManager.EventPipe.enable(with:)`.

AquaGeek:
Fixes #968

I didn't annotate all of the `fatalError` calls, because some of them are effectively unreachable (e.g. `AnyChildWorkflow.enableEvents()`) or are infrastructure problems not consumer problems (i.e. there may not be anything actionable on the side of the library consumer).

### DIFF
--- a/swift/Workflow/Sources/SubtreeManager.swift
+++ b/swift/Workflow/Sources/SubtreeManager.swift
@@ -163,7 +163,7 @@ extension WorkflowNode.SubtreeManager {
             /// If the key already exists in `used`, than a workflow of the same type has been rendered multiple times
             /// during this render pass with the same key. This is not allowed.
             guard usedChildWorkflows[childKey] == nil else {
-                fatalError("Child workflows of the same type must be given unique keys. Duplicate workflows of type \(Child.self) were encountered with the key \"\(key)\"")
+                fatalError("Child workflows of the same type must be given unique keys. Duplicate workflows of type \(Child.self) were encountered with the key \"\(key)\" in \(WorkflowType.self)")
             }
 
             let child: ChildWorkflow<Child>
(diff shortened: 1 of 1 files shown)


swift/Workflow/Sources/SubtreeManager.swift
```diff
@@ -316,19 +316,19 @@ extension WorkflowNode.SubtreeManager {
...
             case .preparing:
-                fatalError("Sink sent an action inside `render`. Sinks are not valid until `render` has completed.")
+                fatalError("[\(WorkflowType.self)] Sink sent an action inside `render`. Sinks are not valid until `render` has completed.")
 
             case .pending:
                 validationState = .queued(event)
 
             case .queued:
-                fatalError("Action sent to pipe while already in the `queueing` state.")
+                fatalError("[\(WorkflowType.self)] Action sent to pipe while already in the `queueing` state.")
```
dhavalshreyas, line 325: This message doesn't make a lot of sense from a developer's perspective. I'm not entirely sure if we can tweak it to make sense. Ideally, this message would indicate what went wrong and what needs to be done to fix it.

AquaGeek: Yeah, I struggled with this one. Not sure what else to do for now…

amorde: Agree this isn't a super helpful message, but adding the Workflow type as least improves it somewhat

bencochran: I should probably know this, but… uh, what leads to this state? Is it sending two events in the same render cycle? Or?

AquaGeek: Um… that's a great question. I'm not entirely sure, TBH. We synchronously force a re-render whenever an event is processed (as of e6938e1?). You can force this to happen by modifying `ConcurrencyTests.test_queuedEvents` to include a second call to `rendering.update()`, though I'm not sure how "possible" this is in normal use.

dhavalshreyas: Thanks @AquaGeek! If this is pretty-much impossible to happen in the real-world, it should be fine to keep this error message as-is.

AquaGeek: I don't think it's _impossible_, but it is highly unlikely.

PR #1156, opened 2020-05-12
Add insecure checksums gradle flag to fix Nexus release.

### LINKED ISSUE #1155: Sonatype releases are broken
It seems that gradle has started publishing sha256 and sha512 checksum files for maven artifacts. As documented in [this Nexus issue](https://issues.sonatype.org/browse/NEXUS-21802), Nexus will fail releasing artifacts when these files are present. The issue is marked as fixed, but we're still experiencing it.

Gradle issue https://github.com/gradle/gradle/issues/11308#issuecomment-554330650 describes a flag we can pass to Gradle to disable generation of these files.

Update: I originally thought this was caused by upgrading to Gradle 6.4, but the 0.27.x release branch is still using gradle 6.3.

zach-klippenstein:
Fixes #1155.

## Checklist

- [ ] Unit Tests
- [ ] UI Tests

### DIFF
--- a/kotlin/gradle.properties
+++ b/kotlin/gradle.properties
@@ -21,6 +21,9 @@ android.enableJetifier=true
 # Uncomment this to diagnose "API $s is obsolete" warnings. Commented out b/c it's pretty noisy.
 #android.debug.obsoleteApi=true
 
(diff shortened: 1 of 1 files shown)


zach-klippenstein:
Verified this removes the files as expected locally, no point waiting for CI to go green.

PR #1157, opened 2020-05-12
Upload android instrumentation test results from CI.

zach-klippenstein:
The upload-archive action now supports globbing, so we can easily
upload all the test reports from all modules.

## Checklist

- [ ] Unit Tests
- [ ] UI Tests

### DIFF
--- a/.github/workflows/kotlin.yml
+++ b/.github/workflows/kotlin.yml
@@ -151,3 +151,9 @@ jobs:
           api-level: ${{ matrix.api-level }}
           arch: x86_64
           script: ./gradlew connectedCheck --build-cache --no-daemon --stacktrace
+
+      - name: Upload results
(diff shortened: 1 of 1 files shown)


.github/workflows/kotlin.yml
```diff
@@ -151,3 +151,9 @@ jobs:
           api-level: ${{ matrix.api-level }}
           arch: x86_64
           script: ./gradlew connectedCheck --build-cache --no-daemon --stacktrace
+
+      - name: Upload results
+        uses: actions/upload-artifact@v2
+        with:
+          name: insrumentation-test-results
+          path: ./**/build/reports/androidTests/connected/**
```
rjrjr, line 159: Where can we see them?

zach-klippenstein: Screenshot_20200512-091610

PR #1162, opened 2020-05-15
Provide default workflowDidChange implementation

dhavalshreyas:
`workflowDidChange` is hardly ever used. Providing a default implementation to reduce unnecessary boilerplate.

### DIFF
--- a/swift/Samples/SampleApp/Sources/DemoWorkflow.swift
+++ b/swift/Samples/SampleApp/Sources/DemoWorkflow.swift
@@ -60,8 +60,6 @@ extension DemoWorkflow {
             subscriptionState: .not
         )
     }
-
-    func workflowDidChange(from previousWorkflow: DemoWorkflow, state: inout State) {}
 }
 
 // MARK: Actions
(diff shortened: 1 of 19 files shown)


AquaGeek:
IIRC (@bencochran keep me honest), the reason we didn't do this before was because if we changed the API (e.g. added a parameter) the compiler wouldn't raise an error on any implementations (as they would use the default). There's no way to mark a function as an `override` of a default implementation. In cases where anything important is being done inside `workflowDidChange`, if we don't have tests in place around that functionality things would silently break.

bencochran:
Yeah, that’s the reason as I remember it. (This came up in Dashboard app specifically when things got renamed, and some stuff stopped working silently).

That said, I might be more inclined to be okay with a default now that the API is settling down towards 1.0? It’s a tricky line to toe for sure.

dhavalshreyas:
That's a valid point. Have there been any other ideas explored for `workflowDidChange`. Seems like 99% of the times we don't need it.

Can we use `Action` for this. Something like: `context.subscribeToWorkflowChanges(action)`

AquaGeek:
> Can we use Action for this. Something like: `context.subscribeToWorkflowChanges(action)`

I'm hesitant to shove anything else into the infrastructure for `Action` s. 🤣

How would we handle cases where `workflowDidChange` would have been called before `render`? (Not entirely sure if this is a valid concern, but there are inherent timing issues.)

The only other approach I can think of would be to create another protocol that inherits from `Workflow` and adds `workflowDidChange`. Or maybe it wouldn't even need to inherit from `Workflow` — we just check in the infrastructure if the running workflow conforms. 🤷

Issue #1046, opened 2020-03-26
Convert Container Screens in Sample to Generic Containers [swift]


dhavalshreyas:
Instead of using `AnyScreen` content, use a generic `ScreenType`.

* `CrossFadeScreen`
*  `BackStackScreen` (#1167)

JustinDSN:
Just to leave a verbal conversation comment trail, we talked about how this also inspires a SwiftUI like IfScreen<TrueScreen, FalseScreen> that could be used to swap out a given screen should a condition evaluates to true.

 Relates to #1164

dhavalshreyas:
`CrossFadeScreen` is meant to handle transitions between different screens. We're not planning on switching that from `AnyScreen` at this time.

PR #1167, opened 2020-05-18
Make BackStackScreen generic

AquaGeek:
Part of #1046

**Pros**
* This makes `BackStackScreen` more consistent with our other containers.
* For back stacks that only have one screen type (e.g. single-screen stacks), there isn't an additional `AnyScreen` wrapping layer anymore.
* Potentially easier testing — you can get at the underlying type without having to cast.

**Cons**
* This creates more work when concatenating multiple back stacks together (which could arguably be pushed into the concatenation API (e.g. `appending<T: Screen>(_ other: BackStack<T>) -> BackStackScreen<AnyScreen>`)).
* More work is required to wrap your screen in an `AnyScreen` before adding it to a `BackStackScreen<AnyScreen>`. ~This could be mitigated by adding helper functions (e.g. in `Screen`: `func asAnyScreen() -> AnyScreen`); otherwise you have to prepend the `AnyScreen(` and reindent.~ The `asAnyScreen()` convenience has been added.

### DIFF
--- a/swift/Samples/TicTacToe/Sources/Game/RunGameWorkflow.swift
+++ b/swift/Samples/TicTacToe/Sources/Game/RunGameWorkflow.swift
@@ -84,19 +84,19 @@ extension RunGameWorkflow {
 // MARK: Rendering
 
 extension RunGameWorkflow {
-    typealias Rendering = AlertContainerScreen<ModalContainerScreen<BackStackScreen>>
+    typealias Rendering = AlertContainerScreen<ModalContainerScreen<BackStackScreen<AnyScreen>>>
 
     func render(state: RunGameWorkflow.State, context: RenderContext<RunGameWorkflow>) -> Rendering {
         let sink = context.makeSink(of: Action.self)
         var modals: [ModalContainerScreenModal] = []
         var alert: Alert?
 
-        var backStackItems: [BackStackScreen.Item] = [BackStackScreen.Item(
-            screen: newGameScreen(
+        var backStackItems: [BackStackScreen<AnyScreen>.Item] = [BackStackScreen.Item(
+            screen: AnyScreen(newGameScreen(
                 sink: sink,
                 playerX: state.playerX,
                 playerO: state.playerO
-            ),
+            )),
             barVisibility: .hidden
         )]
 
@@ -111,7 +111,7 @@ extension RunGameWorkflow {
             )
             .rendered(with: context)
             backStackItems.append(BackStackScreen.Item(
-                screen: takeTurnsScreen,
+                screen: AnyScreen(takeTurnsScreen),
                 barVisibility: .visible(BackStackScreen.BarContent(
                     leftItem: BackStackScreen.BarContent.BarButtonItem.button(BackStackScreen.BarContent.Button(
                         content: .text("Quit"),
@@ -130,7 +130,7 @@ extension RunGameWorkflow {
             )
             .rendered(with: context)
             backStackItems.append(BackStackScreen.Item(
-                screen: takeTurnsScreen,
+                screen: AnyScreen(takeTurnsScreen),
                 barVisibility: .visible(BackStackScreen.BarContent(
                     leftItem: BackStackScreen.BarContent.BarButtonItem.button(BackStackScreen.BarContent.Button(
                         content: .text("Quit"),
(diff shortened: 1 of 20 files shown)


swift/Samples/TicTacToe/Sources/Game/RunGameWorkflow.swift
```diff
@@ -84,19 +84,19 @@ extension RunGameWorkflow {
 // MARK: Rendering
 
 extension RunGameWorkflow {
-    typealias Rendering = AlertContainerScreen<ModalContainerScreen<BackStackScreen>>
+    typealias Rendering = AlertContainerScreen<ModalContainerScreen<BackStackScreen<AnyScreen>>>
```
dhavalshreyas, line 87: 😮

AquaGeek: Yeah…

JustinDSN:
This is more consistent with our other open source containers where we push the wrapping of `AnyScreen` down to the client side code (eg: `SplitScreenContainer`), which at that time we thought was the right direction to take (with that container).

What are the pros & cons of doing it this way?

dhavalshreyas:
Having thought about this more, this definitely feels like the better direction. 

Issues with being able to test `AnyScreen` will also be handled here, since we can use a different comparable wrapper instead.

swift/Samples/TicTacToe/Sources/Game/RunGameWorkflow.swift
rjrjr: What's the 😮 ?

dhavalshreyas: Definitely the right direction as far as type-safety is concerned. Not ideal from the API ergonomics perspective. However, once we bump up to iOS 13 we'll be able to use [opaque types](https://docs.swift.org/swift-book/LanguageGuide/OpaqueTypes.html) and this won't be a concern.

AquaGeek: I think it's a comment on the # of generics `A<B<C<D>>>`.

swift/Samples/TicTacToe/Sources/Game/RunGameWorkflow.swift
```diff
@@ -84,19 +84,19 @@ extension RunGameWorkflow {
...
 
     func render(state: RunGameWorkflow.State, context: RenderContext<RunGameWorkflow>) -> Rendering {
         let sink = context.makeSink(of: Action.self)
         var modals: [ModalContainerScreenModal] = []
         var alert: Alert?
 
-        var backStackItems: [BackStackScreen.Item] = [BackStackScreen.Item(
-            screen: newGameScreen(
+        var backStackItems: [BackStackScreen<AnyScreen>.Item] = [BackStackScreen.Item(
+            screen: AnyScreen(newGameScreen(
```
AquaGeek, line 95: This probably warrants a `asAnyScreen()` helper on `Screen`. The code here would then become:

```swift
screen: newGameScreen(
    sink: sink,
    playerX: state.playerX,
    playerO: state.playerO
).asAnyScreen(),
```

AquaGeek: Or I guess we could add an extension:

```swift
extension BackStackScreen where Content == AnyScreen {
    init<S: Screen>(…) {
        // Do the wrapping into AnyScreen here
    }
}
```

swift/Samples/Tutorial/Frameworks/Tutorial5Complete/Tests/TodoWorkflowTests.swift
```diff
@@ -77,21 +78,21 @@ class TodoWorkflowTests: XCTestCase {
                     // We always expect the TodoListWorkflow
                     ExpectedWorkflow(
                         type: TodoListWorkflow.self,
-                        rendering: BackStackScreen.Item(
+                        rendering: BackStackScreen<AnyScreen>.Item(
```
JustinDSN, line 81: As we talked about in our 1:1, I think we can snip these `<AnyScreen>` and let the compiler infer the generic type.

Issue #702, opened 2019-10-29
Proposal: Unified render testing API for Kotlin and Swift


zach-klippenstein:
First draft of this proposal came out of a discussion between @zach-klippenstein and @bencochran. 

## Background

Swift and Kotlin built completely separate testing APIs. Swift has two APIs: one for testing actions, and one for testing rendering (although it also ends up testing actions). Kotlin also has two APIs, but split on a different axis: one for integration testing whole workflow trees, and one for testing single render passes. The kotlin one has some rough edges, namely #489 and #499, and there's a lot of confusion about how to use it.

#687 is an attempt to build a Kotlin testing API that more closely matched the Swift render testing API. Swift contributors raised some concerns over their API as well though, and so we held off on merging that until we could talk through all the issues and come up with a better solution that met all the requirements.

This proposal is only concerned with the API for unit testing single render passes.

### Kotlin

1. Requires passing in fake child workflows and workers. Since most workflows inject their children, this is required anyway, but the current API requires child workflows to actually return a rendering. Fakes that are required to have some behavior, but not all the behavior (they can't run any grandchildren or anything) is confusing and feels half-baked. Moreover, it's not clear how to actually create these mocks (do you use `MockChildWorkflow` or mockito mocks?).
1. Because child renderings must be returned from the fake workflows, it's not easy to create tests that verify a wide variety of child outputs - the parent workflow itself has to be setup with different fake children every time.

### Swift

1. `ExpectedOutput` is actually an assertion on the result of the render pass. It's confusing because the naming and API looks more similar to the API for setting up expected child workflows/workers.
1. Render tester object is actually stateful, and invoking render event handlers, for example, will actually mutate the state inside of the tester object instead of being pure function calls.
1. `RenderTester.render` is a fluent-style API – it returns `self` so you can chain tests for multiple consecutive render passes. This encourages bloated tests, instead of a 1-to-1 unit test per render pass relation.
1. Because the only way to assert the result of children and events is by asserting on their output (with `ExpectedOutput`), render tests are actually also testing `WorkflowAction`s. Actions should be tested separately.
1. Passing all the expectations and assertions into one giant function call, and requiring manually creating types like `ExpectedWorkflow` and `ExpectedWorker` is clunky and exposes too many types that should probably remain internal.

## Proposed enhancements

Kotlin will get a testing API much more similar to Swifts, using #687 as a starting point. The high-level API usage will look as follows. The testing API is divided into multiple stages.

1. **Stage 1: Setup**
[...]

PR #1174, opened 2020-05-19
Introduce SideEffect #1021

dhavalshreyas:
Introduces a `SideEffect` hook that will be used to perform the functionality currently provided by `Worker`s. 

This is a subset of @bencochran's implementation in the [bc/guwt-prototype](https://github.com/square/workflow/compare/bc/guwt-prototype) branch.

### DIFF
--- a/swift/Workflow/Sources/Lifetime.swift
+++ b/swift/Workflow/Sources/Lifetime.swift
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2020 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+
+/// Represents the lifetime of an object.
+///
+/// Once ended, the `onEnded` closure is called.
+public final class Lifetime {
+    /// Hook to clean-up after end of `lifetime`.
+    public func onEnded(_ action: @escaping () -> Void) {
+        assert(!hasEnded, "Lifetime used after being ended.")
+        onEndedActions.append(action)
+    }
+
+    public private(set) var hasEnded: Bool = false
+    private var onEndedActions: [() -> Void] = []
+
+    deinit {
+        end()
+    }
+
+    func end() {
+        guard !hasEnded else {
+            return
+        }
+        hasEnded = true
+        onEndedActions.forEach { $0() }
+    }
+}
--- a/swift/Workflow/Sources/RenderContext.swift
+++ b/swift/Workflow/Sources/RenderContext.swift
@@ -75,6 +75,24 @@ public class RenderContext<WorkflowType: Workflow>: RenderContextType {
         fatalError()
     }
 
+    /// Execute a side-effect action.
+    ///
+    /// Note that it is a programmer error to run two side-effects with the same `key`
+    /// during the same render pass.
+    ///
+    /// `action` will be executed the first time a side-effect is run with a given `key`.
+    /// `runSideEffect` calls with a given `key` on subsequent renders are ignored.
+    ///
+    /// If after a render pass, a side-effect with a `key` that was previously used is not used,
+    /// it's lifetime ends and the `Lifetime` object's `onEnded` closure will be called.
+    ///
+    /// - Parameters:
+    ///   - key: represents the block of work that needs to be executed.
+    ///   - action: a block of work that will be executed.
+    public func runSideEffect(key: AnyHashable, action: (Lifetime) -> Void) {
+        fatalError()
+    }
+
     final func invalidate() {
         isValid = false
     }
@@ -103,6 +121,11 @@ public class RenderContext<WorkflowType: Workflow>: RenderContextType {
             return implementation.makeSink(of: actionType)
         }
 
+        override func runSideEffect(key: AnyHashable, action: (_ lifetime: Lifetime) -> Void) {
+            assertStillValid()
+            implementation.runSideEffect(key: key, action: action)
+        }
+
         override func awaitResult<W, Action>(for worker: W, outputMap: @escaping (W.Output) -> Action) where W: Worker, Action: WorkflowAction, WorkflowType == Action.WorkflowType {
             assertStillValid()
             implementation.awaitResult(for: worker, outputMap: outputMap)
@@ -122,6 +145,8 @@ internal protocol RenderContextType: AnyObject {
     func makeSink<Action>(of actionType: Action.Type) -> Sink<Action> where Action: WorkflowAction, Action.WorkflowType == WorkflowType
 
     func awaitResult<W, Action>(for worker: W, outputMap: @escaping (W.Output) -> Action) where W: Worker, Action: WorkflowAction, Action.WorkflowType == WorkflowType
+
+    func runSideEffect(key: AnyHashable, action: (_ lifetime: Lifetime) -> Void)
 }
 
 extension RenderContext {
--- a/swift/Workflow/Sources/SubtreeManager.swift
+++ b/swift/Workflow/Sources/SubtreeManager.swift
@@ -35,6 +35,9 @@ extension WorkflowNode {
         /// The current array of workers
         internal private(set) var childWorkers: [AnyChildWorker] = []
 
+        /// The current array of side-effects
+        internal private(set) var sideEffectLifetimes: [AnyHashable: SideEffectLifetime] = [:]
+
         init() {}
 
         /// Performs an update pass using the given closure.
@@ -48,7 +51,8 @@ extension WorkflowNode {
             let context = Context(
                 previousSinks: previousSinks,
                 originalChildWorkflows: childWorkflows,
-                originalChildWorkers: childWorkers
+                originalChildWorkers: childWorkers,
+                originalSideEffectLifetimes: sideEffectLifetimes
             )
 
             let wrapped = RenderContext.make(implementation: context)
@@ -63,6 +67,7 @@ extension WorkflowNode {
             /// as a result of this call to `render`.
             childWorkflows = context.usedChildWorkflows
             childWorkers = context.usedChildWorkers
+            sideEffectLifetimes = context.usedSideEffectLifetimes
 
             /// Captured the reusable sinks from this render pass.
             previousSinks = context.sinkStore.usedSinks
@@ -85,7 +90,7 @@ extension WorkflowNode {
             let queuedEvents = eventPipes.compactMap { pipe in
                 pipe.pendingOutput()
             }
-            if queuedEvents.count > 0 {
+            if !queuedEvents.isEmpty {
                 handle(output: queuedEvents[0])
                 return
             }
@@ -144,7 +149,15 @@ extension WorkflowNode.SubtreeManager {
         private let originalChildWorkers: [AnyChildWorker]
         internal private(set) var usedChildWorkers: [AnyChildWorker]
 
-        internal init(previousSinks: [ObjectIdentifier: AnyReusableSink], originalChildWorkflows: [ChildKey: AnyChildWorkflow], originalChildWorkers: [AnyChildWorker]) {
+        private let originalSideEffectLifetimes: [AnyHashable: SideEffectLifetime]
+        internal private(set) var usedSideEffectLifetimes: [AnyHashable: SideEffectLifetime]
+
+        internal init(
+            previousSinks: [ObjectIdentifier: AnyReusableSink],
+            originalChildWorkflows: [ChildKey: AnyChildWorkflow],
+            originalChildWorkers: [AnyChildWorker],
+            originalSideEffectLifetimes: [AnyHashable: SideEffectLifetime]
+        ) {
             self.eventPipes = []
 
             self.sinkStore = SinkStore(previousSinks: previousSinks)
@@ -154,6 +167,9 @@ extension WorkflowNode.SubtreeManager {
 
             self.originalChildWorkers = originalChildWorkers
             self.usedChildWorkers = []
+
+            self.originalSideEffectLifetimes = originalSideEffectLifetimes
+            self.usedSideEffectLifetimes = [:]
         }
 
         func render<Child, Action>(workflow: Child, key: String, outputMap: @escaping (Child.Output) -> Action) -> Child.Rendering where Child: Workflow, Action: WorkflowAction, WorkflowType == Action.WorkflowType {
@@ -228,6 +244,16 @@ extension WorkflowNode.SubtreeManager {
             } else {
                 let newChildWorker = ChildWorker(worker: worker, outputMap: outputMap, eventPipe: eventPipe)
                 usedChildWorkers.append(newChildWorker)
+            }
+        }
+
+        func runSideEffect(key: AnyHashable, action: (Lifetime) -> Void) {
+            if let existingSideEffect = originalSideEffectLifetimes[key] {
+                usedSideEffectLifetimes[key] = existingSideEffect
+            } else {
+                let sideEffectLifetime = SideEffectLifetime()
+                action(sideEffectLifetime.lifetime)
+                usedSideEffectLifetimes[key] = sideEffectLifetime
             }
         }
     }
@@ -403,7 +429,7 @@ extension WorkflowNode.SubtreeManager {
 
         private var outputMap: (W.Output) -> AnyWorkflowAction<WorkflowType>
 
-        private let (lifetime, token) = Lifetime.make()
+        private let (lifetime, token) = ReactiveSwift.Lifetime.make()
 
         init(worker: W, outputMap: @escaping (W.Output) -> AnyWorkflowAction<WorkflowType>, eventPipe: EventPipe) {
             self.worker = worker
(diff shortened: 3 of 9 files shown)


swift/Workflow/Sources/Lifetime.swift
```diff
@@ -0,0 +1,39 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+
+public final class Lifetime {
+    public func onEnded(_ action: @escaping () -> Void) {
```
JustinDSN, line 20: Does this relate to any existing behavior in RxSwift, ReactiveSwift, or Combine? I'm trying to understand if this is a "Completion" event.

dhavalshreyas: Yup, this is pretty much a completion event. 

This behavior exists in each of those libraries. (`Disposable`, `Cancellable`, `Lifetime`, ...etc)

swift/Workflow/Sources/Lifetime.swift
```diff
@@ -0,0 +1,39 @@
...
+        assert(!hasEnded, "Lifetime used after being ended.")
+        onEndedActions.append(action)
+    }
+
+    public private(set) var hasEnded: Bool = false
+    private var onEndedActions: [() -> Void] = []
```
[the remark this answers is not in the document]
dhavalshreyas: This allows us to add multiple `onEnded` listeners from the same action. I'm not sure if this is needed, however doesn't hurt to support it. cc: @bencochran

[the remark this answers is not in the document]
bencochran: Yeah, mostly to match behavior of similar types from other libraries. And it’s nice to be able to end multiple subscriptions to the outside world in a single side-effect if needed.

swift/Workflow/Sources/SubtreeManager.swift
```diff
@@ -85,7 +90,7 @@ extension WorkflowNode {
             let queuedEvents = eventPipes.compactMap { pipe in
                 pipe.pendingOutput()
             }
-            if queuedEvents.count > 0 {
+            if !queuedEvents.isEmpty {
```
bencochran, line 93: not material to this review, but I’ve personally always had a harder time skimming `!thing.isEmpty` vs `thing.count > 0`

swift/Workflow/Sources/SubtreeManager.swift
```diff
@@ -519,3 +545,19 @@ extension WorkflowNode.SubtreeManager {
...
+extension WorkflowNode.SubtreeManager {
+    internal class SideEffectLifetime {
+        fileprivate let lifetime: Lifetime
+
+        fileprivate init() {
+            self.lifetime = Lifetime()
+        }
+
+        deinit {
+            lifetime.end()
```
bencochran, line 560: Might be good to leave a comment like `// Explicitly end the lifetime in case someone retained it from outside` since that question came up elsewhere and it’s a little subtle.

swift/Workflow/Tests/SubtreeManagerTests.swift
```diff
@@ -294,6 +296,71 @@ final class SubtreeManagerTests: XCTestCase {
...
+                key: "",
+                outputMap: { _ in AnyWorkflowAction.noAction }
+            )
+        }
+
+        XCTAssertEqual(manager.sideEffectLifetimes.count, 1)
+        let sideEffectKey = manager.sideEffectLifetimes.values.first!
+
+        _ = manager.render { context -> TestViewModel in
+            context.runSideEffect(key: "helloWorld") { _ in }
```
bencochran, line 319: We should probably put an `XCTFail()` in that block to ensure that the side-effect isn’t executed the second time. (As is, I think this test would pass if we didn’t correctly reuse side-effects)

swift/Workflow/Tests/SubtreeManagerTests.swift
```diff
@@ -294,6 +296,71 @@ final class SubtreeManagerTests: XCTestCase {
...
+                workflow: TestWorkflow(),
+                key: "",
+                outputMap: { _ in AnyWorkflowAction.noAction }
+            )
+        }
+
+        XCTAssertEqual(manager.sideEffectLifetimes.count, 0)
+        wait(for: [lifetimeEndedExpectation], timeout: 1)
+    }
 }
```
bencochran, line 364: Might be redundant to the two tests, but would be nice to have a test with two different keys between two render passes to make sure we perform both of them?

swift/WorkflowTesting/Sources/RenderExpectations.swift
```diff
@@ -99,6 +102,14 @@ public struct ExpectedWorker {
     }
 }
 
+public struct ExpectedSideEffect {
```
bencochran, line 105: I think right now we don’t have a way to just send a `WorkflowAction` to `RenderTester`. So we can’t write tests that act like a side-effect did something (and we couldn’t convert `ExpectedWorker` to be built on `ExpectedSideEffect`. We should consider adding an action to `ExpectedSideEffect` or a way to send an action to `RenderTester` directly.

swift/WorkflowTesting/Sources/WorkflowRenderTester.swift
```diff
@@ -230,7 +232,7 @@
     fileprivate final class RenderTestContext<T: Workflow>: RenderContextType {
         typealias WorkflowType = T
 
-        private var (lifetime, token) = Lifetime.make()
+        private var (lifetime, token) = ReactiveSwift.Lifetime.make()
```
bencochran, line 235: Do we dislike the need to disambiguate enough to use a different name than `Lifetime`? I couldn’t come up with anything better when I originally typed my version of this, but wanted to raise the question in case others had thoughts.

swift/Workflow/Sources/SubtreeManager.swift
AquaGeek: I think this is mostly a defensive thing — `Collection.isEmpty` is O(1) where `Collection.count` _can_ be O(n). (Granted, we're using an `Array`, which conforms to `RandomAccessCollection` and so `count` is O(1), but 🤷.)

swift/WorkflowTesting/Sources/WorkflowRenderTester.swift
dhavalshreyas: I don't think we need to rename just to disambiguate. As we continue to remove the `ReactiveSwift` dependency from the core library - this will become less of an issue.

swift/WorkflowTesting/Sources/RenderExpectations.swift
dhavalshreyas: Good point @bencochran . 

> or a way to send an action to RenderTester directly.

Any ideas on the best way to do that? 
* Pass in `context`, so that we can `makeSink()` and send event.
* Pass in `inout state` so that we can just modify the state directly.
* Add a generic `ActionType` to `ExpectedSideEffect` and pass in a `sink` of `ActionType`.

dhavalshreyas: @bencochran, @AquaGeek: Please take a look at `ExpectedSideEffect` when you get a chance. I've made it take in a closure which accepts the context.

Currently, the only `init` takes in `key` and `action`, however, I wanted to make sure we can pass in arbitrary block of code in the future to ensure it's in line with the `runSIdeEffect` api.

swift/Workflow/Sources/RenderContext.swift
```diff
@@ -75,6 +75,10 @@ public class RenderContext<WorkflowType: Workflow>: RenderContextType {
         fatalError()
     }
 
+    public func runSideEffect(key: AnyHashable, action: (Lifetime) -> Void) {
```
AquaGeek, line 78: Why `AnyHashable` instead of making the function generic?

```swift
public func runSideEffect<Key: Hashable>(key: Key, action: (Lifetime) -> Void) {
```

swift/Workflow/Tests/SubtreeManagerTests.swift
```diff
@@ -294,6 +296,102 @@ final class SubtreeManagerTests: XCTestCase {
...
+                key: "",
+                outputMap: { _ in AnyWorkflowAction.noAction }
+            )
+        }
+
+        XCTAssertEqual(manager.sideEffectLifetimes.count, 0)
+        wait(for: [lifetimeEndedExpectation], timeout: 1)
+    }
+
+    func test_verifySideEffectsWithDifferentKeysAreExecuted() {
```
AquaGeek, line 367: We're not actually testing that they're executed here, are we? The side effect closures are all empty.

swift/WorkflowTesting/Sources/WorkflowRenderTester.swift
```diff
@@ -288,6 +291,18 @@
             }
         }
 
+        func runSideEffect(key: AnyHashable, action: (Lifetime) -> Void) {
+            guard let sideEffectIndex = expectations.expectedSideEffects.firstIndex(where: { (expectedSideEffect) -> Bool in
```
AquaGeek, line 295: Should expected side effects be a dictionary instead?

swift/Workflow/Sources/RenderContext.swift
dhavalshreyas: Either works. 

Not entirely sure how `AnyHashable` works, but seems like it's able to wrap any `Hashable` type without us having to explicitly do the wrapping.

swift/Workflow/Tests/SubtreeManagerTests.swift
dhavalshreyas: Good point. I'll verify it's actually executed.

swift/WorkflowTesting/Sources/WorkflowRenderTester.swift
dhavalshreyas: Yes, that'd be more correct. However, the API felt weird.

zach-klippenstein:
Been meaning to review this, hopefully will get to it today or Monday since I want to start working on this for Kotlin soon.

swift/Workflow/Sources/RenderContext.swift
AquaGeek: #TIL

I assume it has something to do with [this](https://github.com/apple/swift/blob/master/stdlib/public/core/AnyHashable.swift#L284-L288).

swift/WorkflowTesting/Sources/WorkflowRenderTester.swift
AquaGeek: Which part felt weird? Creating the expected side effects in your tests? (If so, I think it's fine to keep those as an array and just change the storage. :shrug:)

dhavalshreyas: Gotcha! Yes, that makes sense, I'll make that change.

dhavalshreyas: Done.

swift/WorkflowTesting/Sources/RenderExpectations.swift
```diff
@@ -99,6 +104,24 @@ public struct ExpectedWorker {
     }
 }
 
+public struct ExpectedSideEffect<WorkflowType: Workflow> {
+    let key: AnyHashable
+    let action: ((RenderContext<WorkflowType>) -> Void)?
```
AquaGeek, line 109: What's the purpose of this closure?

dhavalshreyas: > Currently, the only init takes in key and action, however, I wanted to make sure we can pass in arbitrary block of code in the future to ensure it's in line with the `runSideEffect` api.

AquaGeek:
The only part unclear to me is the purpose of the closure on `ExpectedSideEffect`. What is the use case for that? Are engineers supposed to use that to e.g. fulfill an `XCTestExpectation`?

swift/WorkflowTesting/Sources/RenderExpectations.swift
zach-klippenstein: But what is the block intended to do? If you're writing a test for a workflow that directly invokes `runSideEffect`, what kinds of things would you actually want to assert on?

1. That `runSideEffect` was called with a given key.
2. That `runSideEffect` was _not_ called with a given key.
3. Given that a side effect was ran with a given key, that the side effect lambda actually performed some side effect (e.g. called `logger.log()` with some message).
4. Given that a side effect was ran with a given key, that the side effect lambda registers the correct logic to be executed in the `onEnded` handler.
5. Given that a side effect was ran with a given key, that the side effect lambda sent some action to the event sink.

1 can definitely be done with this API, presumably 2 can as well (by _not_ expecting the key?). I'm not sure how close the Swift render testing API is to the kotlin one yet, but if the testing infrastructure runs the action immediately, then 3 and 5 can be done with existing APIs (3 by asserting the side effect fake was invoked after the render call, 5 using [`verifyAction(Result)`](https://square.github.io/workflow/kotlin/api/workflow/com.squareup.workflow.testing/-render-test-result/)). The only special case is 4, which requires an API to end the lifetime. Some ways to do this:

1. Pass a flag in `ExpectedSideEffect` that, if set, will cause the `onEnded` handlers to be invoked immediately.
2. An additional API on [`RenderTestResult`](https://square.github.io/workflow/kotlin/api/workflow/com.squareup.workflow.testing/-render-test-result/) called something like `endSideEffect(key: String): RenderTestResult` that executes the end actions for the given side effect and returns a `RenderTestResult` that allows verifying if any actions were sent to the sink in the `onEnded` handler.

I think 1 is the simpler option, although both could be tricky (if the side effect immediately sends an action, and then sends another action when ending, the initial action will always be evaluated first and a second render pass would be needed to evaluate the second, and that's hard to express in the render tester API since it, by definition, tests a single render pass). Perhaps we don't actually need to need to provide the ability to test side effect implementations in this API, and can use the more general "integration" testing API?

In any case, this lambda parameter doesn't seem to address any of these use cases so I'm also not sure what purpose it serves.

swift/Workflow/Sources/RenderContext.swift
zach-klippenstein, line 78: This needs documentation, especially discussing that the side effect is allowed to send actions to the context's `Sink`, that the side effect is expected to clean itself up when the `Lifetime` is ended, that the `Lifetime` won't end until a render pass in which this function is not called with the given key, etc.

swift/WorkflowTesting/Sources/RenderExpectations.swift
dhavalshreyas: This is in response to @bencochran 's feedback here: https://github.com/square/workflow/pull/1174#discussion_r430828163

`ExpectedWorker` allows you to send an `Output` back to the `RenderContext` when executed. Trying to provide the same ability here.

The `action` is actually not public, it's not meant to be used directly. The two public initializers are:

```swift
public init(key: AnyHashable)
public init<ActionType: WorkflowAction>(key: AnyHashable, action: ActionType) where ActionType.WorkflowType == WorkflowType
```

I wanted to make sure `ExpectedSideEffect` is implemented similarly to how `SideEffects` are implemented and hence in the future we can provide additional ways to pass `WorkflowAction`s to the `RenderContext`.

swift/Workflow/Sources/Lifetime.swift
```diff
@@ -0,0 +1,39 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+
+public final class Lifetime {
```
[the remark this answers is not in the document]
dhavalshreyas: This just sets us up for the work that'll let us close those issues. It doesn't close any as-is.

swift/WorkflowTesting/Sources/RenderExpectations.swift
zach-klippenstein: I'm worried we never quite converged our testing APIs, and now they're diverging more. Need to dig into this more to understand where the differences are. I could have sworn we had an issue to track testing APIs but I can't find it.

AquaGeek: > I could have sworn we had an issue to track testing APIs but I can't find it.

#702?

swift/Workflow/Sources/RenderContext.swift
bencochran: Also worth noting that the given closure will not be executed on subsequent runs of the same side-effect key.

swift/Workflow/Sources/SubtreeManager.swift
```diff
@@ -419,7 +445,7 @@ extension WorkflowNode.SubtreeManager {
                 .observe(on: QueueScheduler.workflowExecution)
                 .start { [weak self] event in
                     switch event {
-                    case .value(let output):
+                    case let .value(output):
```
bencochran, line 448: nit: I prefer the `let` inside the case for the reason outlined in the [Google Swift style guide](https://google.github.io/swift/#pattern-matching)

swift/WorkflowTesting/Sources/RenderExpectations.swift
bencochran: I think the public parts of `ExpectedSideEffect` as written here fits with the overall current shape of our `RenderTester`. Though, yeah, I’d prefer we converged in the direction of #702 and what Kotlin has.

I’m not sure of the need internally to hold a `((RenderContext<WorkflowType>) -> Void)?` vs an `AnyWorkflowAction<WorkflowType>?` but that’s implementation detail that I’m less concerned with, I think?

zach-klippenstein: My plan is to work on the Kotlin side of this once I've split the repo out, this week. If yall are happy with this then SGTM, I'll not look too hard, see what I come up with in Kotlin, hopefully it's the same! And if not, I'll be in a much better position to debate 😉

swift/Workflow/Sources/SubtreeManager.swift
dhavalshreyas: This will require a `swiftformat` rules change, will do it in a subsequent PR.

PR #1175, opened 2020-05-20
Fix missing libswiftos.dylib

AquaGeek:
`pod spec lint` was failing with the error:

> xctest encountered an error (Failed to load the test bundle. (Underlying error: The bundle “Workflow-Unit-Tests” couldn’t be loaded because it is damaged or missing necessary resources. The bundle is damaged or missing necessary resources. dlopen_preflight(DerivedData/App-fjnxhlsdrldlhtbvawtgcubjgass/Build/Products/Debug-iphonesimulator/Workflow-Unit-Tests.xctest/Workflow-Unit-Tests): Library not loaded: @rpath/libswiftos.dylib

*Note*: This PR is targeting the `release-v0.28.x` branch, as we should probably tag a `v0.28.1`.

### DIFF
--- a/Workflow.podspec
+++ b/Workflow.podspec
@@ -21,6 +21,7 @@ Pod::Spec.new do |s|
   s.test_spec 'Tests' do |test_spec|
     test_spec.source_files = 'swift/Workflow/Tests/**/*.swift'
     test_spec.framework = 'XCTest'
+    test_spec.library = 'swiftos'
   end
 
 end


AquaGeek:
Hrmph. Looks like `release-v0.28.x` was *not* the branch I wanted.

dhavalshreyas:
> Hrmph. Looks like `release-v0.28.x` was _not_ the branch I wanted.

Why not?

AquaGeek:
`release-v0.28.x` already has the version bump to `v0.29.0`. I cherry-picked the commit onto `master` and cut a new branch with that commit and the `v0.28.1` bumps off the `v0.28.0` tag.

dhavalshreyas:
AFAIK, that must be an error. `release-v0.28.x` should not be bumped up to `v0.29.0`. cc: @zach-klippenstein

PR #1180, opened 2020-05-23
Simplify the entry point into the workflow runtime.

### LINKED ISSUE #463: Kotlin runtime Outputs flow should be a suspending callback
In `WorkflowHost` `outputs` was a `ReceiveChannel`. It then changed to a `Flow` (#426), to mirror `renderingsAndSnapshots`, and became backed by an `ArrayBroadcastChannel`. 

`outputs` is different than `renderingsAndSnapshots` in two important ways:
1. Outputs are _events_, and should never be cached, conflated, or dropped. Renderings are snapshots of state and can be cached + conflated.
2. Because of 1, when there is backpressure consuming outputs, they must either be buffered or propagate backpressure (i.e. pausing the workflow runtime).

#426 actually introduced two changes to outputs, with different implications:
1. Converted backing channel from a `Channel` to an `ArrayBroadcastChannel`.
    - This changed the behavior of `send` to be non-blocking and drop items when there are no subscribers.
    - It also introduced an extra layer of buffering.
2. Converted API from `ReceiveChannel` to `Flow`. This was required for 1, because `Flow` is effectively to `BroadcastChannel` what `ReceiveChannel` is to `Channel` – there is no other way to expose a `BroadcastChannel` as read-only. However, it would still be possible to expose a `Channel` as a `Flow` – doing so would potentially be confusing because the consumer would need to be aware that the flow was not multicasted and that multiple concurrent collectors would never see the same items (similar to the caveat on RxBinding's observables).

### `Channel` Pros

- Very hard to accidentally drop outputs. If the consumer starts consuming outputs late, the runtime will pause.

### `Channel` Cons

- If the consumer (caller of `launchWorkflowIn`) does _not_ care about outputs, and doesn't do anything with the output channel, the runtime will suspend forever.
  - This might be hard to debug, because there are no errors and we can't detect this situation except for maybe using a timeout, which may be too prescriptive (what if the consumer wants to pause the runtime indefinitely)?
  - Mitigated by the fact that we've never seen a use case for ignoring the root outputs, and so if you've forgotten to consume them you probably intended to anyway.

### `ReceiveChannel` Pros

- Makes very clear that the source is not multicasted for you. No need to try to re-document the potentially confusing `ArrayBroadcastChannel` backpressure behavior ourselves.
- Forces consumer to think about multicasting behavior if (but only if) they need it. Just forward the channel to a `BroadcastChannel` (or convert to a single-consumer Flow and use a ([proposed](https://github.com/Kotlin/kotlinx.coroutines/issues/1261)) multicasting operator.
- Hints that backpressure is significant. Documentation would still need to point out that the channel must be consumed for the runtime to progress, but we have not seen a use case yet for the root workflow's outputs being ignored anyway, so this is already the default case.

### `ReceiveChannel` Cons

- Cancellation behavior is ambiguous. Usually cancelling a channel cancels the resource it represents. 
  - However, we already let you cancel the runtime by cancelling the scope receiver, so this would be redundant.
  - Cancellation could also just mean all further outputs are dropped, which treats the channel more like a one-time subscription to a `BroadcastChannel`.
    - This isn't quite "idiomatic", but we would document very clearly and I think cancelling the outputs channel prematurely would be an edge-case anyway.
- If you want to treat outputs as a Flow (e.g. to apply some operators, expose a cleaner external API, etc.), you need one more step. Mitigated because this step forces the consumer to think explicitly about potential issues with doing so and choose which behavior they want to opt-in to, instead of automatically getting a potentially surprising behavior that is easy to forget about.

## Conclusion

We're leaning towards backing `outputs` with a simple, unbuffered/`RENDEZVOUS` `Channel` and exposing a `ReceiveChannel`. It involves less complicated/surprising backpressure behavior and communicates more clearly that `outputs` is hot (whereas `renderingsAndSnapshots` still functions as cold, because it's conflated). For most use cases, the actual usage of both APIs should be very similar, but when multicasting is desired, you're forced to understand the implications of doing so.

### LINKED ISSUE #1179: Kotlin runtime Flow of renderings should be a StateFlow
Since a rendering is always synchronously available, the stream should indicate that with the type system by using the StateFlow type instead of just Flow. StateFlow has the explicit contract that a value is always available, so using that type here communicates the contract of the runtime more clearly.

This also implies that we use a `MutableStateFlow` instead of a broadcast channel to back the flow. I don't think this should be a breaking change but since it's a different implementation of broadcasting, it is slightly riskier than just changing the public API.

Note that while `StateFlow` is currently marked as experimental, according to the docs it is stable for use in libraries, just not inheritance.

This is not relevant for the output flow, which should still be converted to a simple callback (see https://github.com/square/workflow/issues/463#issuecomment-556202723).

zach-klippenstein:
This drops the weird `WorkflowSession` concept, and the even weirder `onBeforeStart` function. The new `launchWorkflowIn2` function (to be renamed) just takes a callback to handle outputs (it's a suspending callback, so we still keep backpressure) and returns a `StateFlow` of renderings and snapshots. `StateFlow` is a new Kotlin specialization of `Flow` that always has a value available. This API is a _lot_ more straightforward and idiomatic. The issue of how to ensure that users don't miss really fast outputs is solved by requiring an output callback function as a parameter. The weird issue where, inside of `onBeforeStart`, you had a `CoroutineScope` receiver but it was different than the one passed into the launch function is gone – there's now only a single scope, the one you pass in.

This change also rewrites the main workflow loop to be cleaner and (hopefully) more readable. Notably, in order to synchronously generate the first rendering, the first render pass is performed synchronously, before even launching the main runtime coroutine. This means the first render pass may run on a different thread than the rest, but I don't think anyone cares since most runtimes are started and executed on the main thread, and `render` functions should all be side-effect-free anyway.

Fixes #463 and fixes #1179.

## Checklist

- [x] Unit Tests
- [x] UI Tests


kotlin/workflow-runtime/src/main/java/com/squareup/workflow/LaunchWorkflow.kt
```diff
@@ -112,6 +122,52 @@ fun <PropsT, OutputT : Any, RenderingT, RunnerT> launchWorkflowIn(
...
 
+@OptIn(ExperimentalCoroutinesApi::class, VeryExperimentalWorkflow::class)
+fun <PropsT, OutputT : Any, RenderingT> launchWorkflowIn2(
+  scope: CoroutineScope,
+  workflow: Workflow<PropsT, OutputT, RenderingT>,
+  props: StateFlow<PropsT>,
+  initialSnapshot: Snapshot? = null,
+  diagnosticListener: WorkflowDiagnosticListener? = null,
+  onOutput: suspend (OutputT) -> Unit
+): StateFlow<RenderingAndSnapshot<RenderingT>> {
```
zach-klippenstein, line 133: This is the main part I'd like early feedback on, the implementation details are still a bit of a WIP.

kotlin/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/WorkflowRunnerViewModel.kt
```diff
@@ -116,3 +106,10 @@ internal class WorkflowRunnerViewModel<OutputT : Any>(
     val BUNDLE_KEY = WorkflowRunner::class.java.name + "-workflow"
   }
 }
+
+@OptIn(ExperimentalCoroutinesApi::class)
+private fun <T, R> StateFlow<T>.mapState(transform: (T) -> R): StateFlow<R> =
+  // map takes a suspend function, so we can't just pass the function reference in.
+  object : StateFlow<R>, Flow<R> by map<T, R>({ transform(it) }) {
+    override val value: R get() = transform(this@mapState.value)
+  }
```
zach-klippenstein, line 115: See https://github.com/Kotlin/kotlinx.coroutines/issues/2081.

zach-klippenstein:
While this API is super elegant and really clearly communicates the stateful semantics of props and renderings, it unfortunately does not play very nicely with Compose.

In Compose, compositions are transactional. This allows them to run across multiple threads, and maybe some other features (not clear exactly what those are, but it's clear this isn't going away). This means that side effects shouldn't be performed directly in Composable functions, but must be deferred until the composition is "committed", via the `onCommit` function (see [this slack thread](https://kotlinlang.slack.com/archives/CJLTWPH7S/p1591565582397500?thread_ts=1591558024.394400&cid=CJLTWPH7S) from Adam Powell). In particular, the Compose support for launching composition-scoped coroutines won't launch the coroutine until the composition is committed. This means the coroutine scope won't be available for the workflow runtime synchronously. With this API as it currently exists, we'd need to defer calling `renderWorkflowIn` until after the first method call, which means returning some "not started yet" sentinel. This is unfortunate because the initial rendering from the workflow is actually available before the coroutine launches, but the current API just doesn't expose a way to get it without also specifying the scope.

One way to solve this is to factor out a separate function that the current `renderWorkflowIn` functions delegate to:
```kotlin
fun prepareWorkflow(
  workflow: Workflow<…>,
  props: …
): Pair<RenderingAndSnapshot, WorkflowStarter>

typealias WorkflowStarter = (CoroutineScope) -> StateFlow<RenderingAndSnapshot>
```

I don't really like this solution because:
1. It is a much weirder and more complex API than what we've got now (although this could be mitigated by only referencing the "nice" `renderWorkflowIn` in samples and most docs).
2. It allows calling the `WorkflowStarter` multiple times. This is not a valid operation, since the runtime is still using a single, stateful root `WorkflowRunner` and so we would need to throw an exception if this method is called more than once, which is awkward (runtime vs compile time error).
3. The only reason for gunking the works up with this is to support weirdness of one particular UI framework that won't even be usable in production for at least a year or two.

That said, this change is only required if we want [`renderAsState()`](https://github.com/square/workflow-kotlin-compose/blob/bfc3ba31e125e3183532166fa9c08f8203fd70ec/core-compose/src/main/java/com/squareup/workflow/ui/compose/RenderAsState.kt#L71-L75) to be able to synchronously return the first rendering. If we're ok with the sentinel solution, we can keep the current core runtime API like it was originally in this PR, but `renderAsState()` would need to either require a non-nullable `RenderingT` or use some sort of other optional type to return the rendering.

Anyone have any better ideas?

PR #1182, opened 2020-05-27
[RFC] [DNM] SideEffect as Type

dhavalshreyas:
## Context 
In #1174, we introduce the `sideEffect` API on `RenderContext`:
```swift
runSideEffect(key: AnyHashable, action: (Lifetime) -> Void)
```

`action` is executed the first time a `key` is encountered and `Lifetime.onEnded` is called after it's no longer needed.

Example usage:
```swift
sink = context.makeSink(Action.self)
context.runSideEffect("entity-" + entityId) { lifetime in 
  fetchEntity(entityId) { sink.send(.gotEntity) }
}
```

## Proposal
Make `SideEffect` a type:

```swift
protocol SideEffect: Hashable {
  func run(sink: Sink<Action>) -> Lifetime
}
```

and the API becomes:
```swift
func run<S: SideEffect>(sideEffect: S)
```

### Harder to mess-up
The `Hashable` contract implies that the `hashValue` will be calculated based on hashing all the contents of the object (`SideEffect`). If the `hashValue`s are the same, we then compare the two objects to make sure they're the same. This contract gives us a strong guarantee that the **block of work** that will be executed by the `SideEffect` will be the same.

If we instead put the responsibility on the consumer of the API to formulate the right `key`, we will be unnecessarily burdening the caller to carefully generate the right `key` and the impact of doing this incorrectly will be subtle. 

In other words, if we want the `key` to represent the same `action` AND the `action` is a function of `(a, b, c)` I believe that the the key should be `f(a, b, c)`

### Flexible
Making `SideEffect` a protocol (instead of having the `key` and an independent `action` closure) will make this more flexible. We will then be able to get any of the `Worker`s to conform to the `SideEffect`. (Instead of having an intermediate `Workflow` to do the `runSideEffect` call.)

### DIFF
--- a/swift/Samples/SampleApp/Sources/DemoWorkflow.swift
+++ b/swift/Samples/SampleApp/Sources/DemoWorkflow.swift
@@ -170,16 +170,15 @@ extension DemoWorkflow {
         case .not:
             subscribeTitle = "Subscribe"
         case .subscribing:
-            context.run(
-                sideEffect: SideEffectPerformer<Self, Action>(key: "Timer", action: { sink, lifetime in
-                    let timer = Timer.scheduledTimer(withTimeInterval: 1, repeats: true) { timer in
-                        sink.send(.titleButtonTapped)
-                    }
+            context.run(SideEffectPerformer<Action>(key: "Timer") { sink, lifetime in
+                let timer = Timer.scheduledTimer(withTimeInterval: 1, repeats: true) { timer in
+                    sink.send(.titleButtonTapped)
+                }
 
-                    lifetime.onEnded {
-                        timer.invalidate()
-                    }
-                })
+                lifetime.onEnded {
+                    timer.invalidate()
+                }
+            }
             )
             subscribeTitle = "Stop"
         }
--- a/swift/Samples/SampleApp/Sources/DemoWorkflow.swift
+++ b/swift/Samples/SampleApp/Sources/DemoWorkflow.swift
@@ -30,7 +30,6 @@ struct DemoWorkflow: Workflow {
 
 extension DemoWorkflow {
     struct State {
-        fileprivate var signal: TimerSignal
         var colorState: ColorState
         var loadingState: LoadingState
         var subscriptionState: SubscriptionState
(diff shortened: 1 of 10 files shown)


dhavalshreyas:
Actually, thinking through this..since we'll be using the `sink` to dispatch changes - this model might not make as much sense. :( 

Back to the drawing board.

JustinDSN:
Why did we close / reopen? 

Did something change?

swift/Workflow/Sources/SideEffect.swift
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+public protocol SideEffect: Hashable {
+    associatedtype Action: WorkflowAction
+
+    func run(sink: Sink<Action>) -> Lifetime
+}
```
AquaGeek, line 21: Squinting, this looks a lot like `Worker`.

dhavalshreyas:
> Why did we close / reopen?
> 
> Did something change?

Yes, we're now passing the `sink` on `run`.

swift/Workflow/Sources/SideEffect.swift
dhavalshreyas: It pretty much is - except we're moving the `ReactiveSwift` specific stuff out of `Worker`. So instead of it returning `SignalProducer<Output>`, we'll be passing in a `Sink` and use an internal concept for `Lifetime` instead.

Oh, we're also adding a `hash` value to it.

swift/Samples/SampleApp/Sources/DemoWorkflow.swift
```diff
@@ -201,20 +205,20 @@ extension DemoWorkflow {
...
+//
+//        let timer = Timer.scheduledTimer(withTimeInterval: 1.0, repeats: true) { [weak observer] _ in
+//            observer?.send(value: ())
+//        }
+//
+//        self.signal = signal
+//        self.observer = observer
+//        self.timer = timer
+//    }
+// }
```
AquaGeek, line 224: Can this be deleted?

dhavalshreyas: Yes!

dhavalshreyas:
This was discussed in #1021, the team has decided to stick with the API as-designed.

PR #1189, opened 2020-06-05
Remove mention of Android template repo from README

zach-klippenstein:


gabriel403:
What was behind the decision to stop supporting the template repo?

Issue #670, opened 2019-10-14
Stock Kotlin Container classes should extend ViewGroup, not FrameLayout


rjrjr:
Extending FrameLayout is just sloppy, implies general use that is not supported.

wugeorgeq:
Looking at `BackStackContainer` in `workflow-ui:backstack-android` and `WorkflowLayout` in `core-android`. Seems like these are intended to display one child view at a time, would it make sense to override `onLayout` like so (it's an abstract function in ViewGroup)? 
```
override fun onLayout(
    changed: Boolean,
    left: Int,
    top: Int,
    right: Int,
    bottom: Int
  ) {
    currentView?.layout(0, 0, 0, 0)
  }
```

rjrjr:
Yes, exactly this. And IIRC you need to implement `onMeasure` too.

I think that's true of every `FrameLayout` subclass in the library, I was just lazy.

rjrjr:
See `ScrimContainer` for an example.

zach-klippenstein:
Should we make a `PassthroughLayout` class?

rjrjr:
My gut is that it would be overkill. The copy / paste is so minimal.

rjrjr:
Fixed in #1193

PR #1193, opened 2020-06-08
Extend ViewGroup in stock Kotlin Container classes

wugeorgeq:
## Checklist

- [ ] Unit Tests
- [ ] UI Tests

Short discussion in this issue:
https://github.com/square/workflow/issues/670

### DIFF
--- a/kotlin/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/WorkflowLayout.kt
+++ b/kotlin/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/WorkflowLayout.kt
@@ -24,7 +24,6 @@ import android.util.SparseArray
 import android.view.View
 import android.view.ViewGroup
 import android.view.ViewGroup.LayoutParams.MATCH_PARENT
-import android.widget.FrameLayout
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.ExperimentalCoroutinesApi
@@ -41,9 +40,9 @@ import kotlinx.coroutines.flow.onEach
 class WorkflowLayout(
   context: Context,
   attributeSet: AttributeSet? = null
-) : FrameLayout(context, attributeSet) {
+) : ViewGroup(context, attributeSet) {
   private val showing: WorkflowViewStub = WorkflowViewStub(context).also {
-    addView(it, ViewGroup.LayoutParams(MATCH_PARENT, MATCH_PARENT))
+    addView(it, LayoutParams(MATCH_PARENT, MATCH_PARENT))
   }
 
   private var restoredChildState: SparseArray<Parcelable>? = null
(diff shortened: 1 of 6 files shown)


kotlin/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/WorkflowLayout.kt
```diff
@@ -41,13 +40,17 @@ import kotlinx.coroutines.flow.onEach
...
-) : FrameLayout(context, attributeSet) {
+) : ViewGroup(context, attributeSet) {
   private val showing: WorkflowViewStub = WorkflowViewStub(context).also {
-    addView(it, ViewGroup.LayoutParams(MATCH_PARENT, MATCH_PARENT))
+    addView(it, LayoutParams(MATCH_PARENT, MATCH_PARENT))
   }
 
   private var restoredChildState: SparseArray<Parcelable>? = null
 
+  private val child: View
```
wugeorgeq, line 50: Is it worth it to make this an internal extension val?

zach-klippenstein: If this is changed to my suggestion, it couldn't be. And if it's not changed, then I still think probably not because keeping these classes self-contained makes them easily-reproducable samples for third party containers. If we trigger the rule of three and are worried about DRY, then I think a better solution than just factoring out this one property is factoring out a base class that also implements the two `ViewGroup` methods as well.

kotlin/workflow-ui/core-android/src/main/java/com/squareup/workflow/ui/WorkflowLayout.kt
```diff
@@ -41,13 +40,17 @@ import kotlinx.coroutines.flow.onEach
...
+    get() = getChildAt(0)
+        ?: error("Child must be set immediately upon creation.")
```
zach-klippenstein, line 52: @rjrjr should confirm, but I think this can be more safely expressed as `showing.actual ?: showing`.

The view is always added to the hierarchy, so there's no need to validate that here.

kotlin/workflow-ui/modal-android/src/main/java/com/squareup/workflow/ui/modal/ModalContainer.kt
```diff
@@ -46,14 +45,18 @@ abstract class ModalContainer<ModalRenderingT : Any> @JvmOverloads constructor(
...
+) : ViewGroup(context, attributeSet, defStyle, defStyleRes) {
 
   private val baseView: WorkflowViewStub = WorkflowViewStub(context).also {
-    addView(it, ViewGroup.LayoutParams(MATCH_PARENT, MATCH_PARENT))
+    addView(it, LayoutParams(MATCH_PARENT, MATCH_PARENT))
   }
 
   private var dialogs: List<DialogRef<ModalRenderingT>> = emptyList()
 
+  private val child: View
```
rjrjr, line 56: Just use `baseView.actual` directly, please.

PR #1200, opened 2020-06-11
Fix deploy_website script to work with extracted repos.

### LINKED ISSUE #1199: Fix deploy_website script to deploy from extracted repos
Extracting libraries into language-specific repos (#1186 and #1187) breaks our website deploy scripts. Once they're both out, the script needs to be updated. I don't think it needs to check this repo out separately anymore, it can just use the current working copy. It does need to check out both swift and kotlin repos, and so it will need a ref for each of those.

zach-klippenstein:
Fixes #1199.

Moved the swift script into a function right in the deploy_website.sh file to live beside the similar kotlin logic.

Updated the Kotlin API doc redirects to use the new package name.

Tested by doing dry runs locally and ensuring that all the API docs are in the right place, and `mkdocs serve` looks good.


AquaGeek:
Swift has been extracted.

zach-klippenstein:
Fixing the Kotlin output directory: https://github.com/square/workflow-kotlin/pull/21

zach-klippenstein:
Finished updating the script. Moved the swift script into a function right in the `deploy_website.sh` file to live beside the similar kotlin logic.

PR #1201, opened 2020-06-11
Remove Swift code

### LINKED ISSUE #1186: Extract Swift code into dedicated repository and remove from this repo.
Swift child of #1129.

https://github.com/square/workflow-swift

_See #1187 for the Swift counterpart._

AquaGeek:
This has been extracted into a separate repository (https://github.com/square/workflow-swift)

Closes #1186


.github/workflows/release.yml
```diff
@@ -7,8 +7,6 @@ on:
 env:
```
amorde, line 7: do we need this at all anymore now that we won't be releasing from this repo? or will we be releasing docs with this?

AquaGeek: We'll be releasing docs from this repo, but what that ultimately looks like is still WIP.

PR #1216, opened 2020-11-02
Create GitHub action to manually publish documentation

AquaGeek:
This creates a new GitHub Action that can be manually run to generate the documentation for Kotlin and Swift and publishes the resulting site to a defined branch. (I didn't want to auto-merge into `gh-pages` quite yet.)

We will need to take some time to reconcile this against the existing `deploy_website.sh` script.

### DIFF
--- a/.github/workflows/update-docs.yml
+++ b/.github/workflows/update-docs.yml
@@ -1,8 +1,6 @@
 name: Publish Documentation Site
 
 on:
-  push:
-    branches: [stromberg/manual-publish]
   workflow_dispatch:
     inputs:
       main-ref:
--- a/.github/workflows/update-docs.yml
+++ b/.github/workflows/update-docs.yml
@@ -1,8 +1,9 @@
 name: Publish Documentation Site
 
 on:
+  push:
+    branches: [stromberg/manual-publish]
   workflow_dispatch:
-    # Inputs the workflow accepts.
     inputs:
       branch:
         description: 'Branch'
--- a/.github/workflows/update-docs.yml
+++ b/.github/workflows/update-docs.yml
@@ -1,0 +1,49 @@
+name: Publish Documentation Site
+
+on:
+  workflow_dispatch:
+    # Inputs the workflow accepts.
+    inputs:
+      branch:
+        description: 'Branch'
+        default: 'main'
+        required: true
+
+jobs:
+  build-docs:
+    runs-on: macos-latest
+
+    steps:
+    - name: Checkout Branch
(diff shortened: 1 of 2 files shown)


.github/workflows/update-docs.yml
```diff
@@ -0,0 +1,50 @@
+name: Publish Documentation Site
+
+on:
+  push:
+    branches: [stromberg/manual-publish]
+  workflow_dispatch:
```
zach-klippenstein, line 6: Haven't seen this one. Could you add a comment documenting why it's needed here?

.github/workflows/update-docs.yml
```diff
@@ -0,0 +1,50 @@
...
+
+    - name: Set up Swift environment
+      run: |
+        # Set global bundle path so it gets used by build_swift_docs.sh running in the nested repo as well.
+        cd "${{ github.event.inputs.name }}"
+        bundle config --global path "$(pwd)/.bundle"
+        bundle check || bundle install
+        # Don't need to run pod gen, the website script does that itself.
+        brew install sourcedocs
+        sudo xcode-select -s /Applications/Xcode_11.4.app
```
zach-klippenstein, line 39: Isn't this a pretty old version of Xcode?

.github/workflows/update-docs.yml
```diff
@@ -0,0 +1,50 @@
+name: Publish Documentation Site
+
+on:
+  push:
+    branches: [stromberg/manual-publish]
```
zach-klippenstein, line 5: I assume this is just here for testing this PR?

zach-klippenstein (approved):

.github/workflows/update-docs.yml
AquaGeek: Yep. Just for testing.

I actually created a separate repo on my personal account to avoid spamming everyone on this repo with build failures. I'll update the PR when all of that is working as intended.

.github/workflows/update-docs.yml
AquaGeek: This allows us to manually kick off the workflow. We may want to (eventually?) move this to automatically happen on commits to `main`, but for now it's a manual process that lets you choose the git ref to use for building the docs.

.github/workflows/update-docs.yml
AquaGeek: Ish. Right now, our language version targets Swift 5.2, which was released with Xcode 11.4. We could probably bump this up, but I'm fine leaving this as-is for now.

PR #1249, opened 2022-05-05
Fixes glossary formatting again.

rjrjr:
Using &lt;dd> botched all the 'graph and 'list markdown in the glossary entries. I think I missed it b/c I previewed as github markdown?

I made no effort at all to format this well -- life is too short for &lt;/p> and &lt;/li>.


rjrjr:
Oops, missed a bunch.

PR #1252, opened 2022-05-05
Adds one-and-done pattern.

### LINKED ISSUE #1251: Add an entry to userguide/core-patterns about using PropsT for one-shot workflows
Use RenderingT instead of OutputT on workflows that might want to return a result immediately, or for one-and-done workflows
A common question is “why can’t I emit OutputT from initialState,” or “what if the workflow realizes it doesn’t actually need to run? The most efficient, and most expressive, way to handle this is to use an optional or conditional RenderingT, and an OutputT of Nothing
 
Imagine a PromptForPermissionMaybeWorkflow, that renders a UI to get a passcode, but only if that permission has not already been granted. If you make its RenderingT nullable (e.g. LayerRendering?), it can return null to indicate that its job is done. Its callers will be synchronously informed that the coast is clear, and can immediately render what they actually care about.
 
Another variation of this pattern is to use a sealed class for RenderingT, with a WorkingState that implements LayerRendering, and a unviewable Finished state that carries the work product.
  
A good rule of thumb for choosing between using RenderingT or OutputT is to remember that OutputT is event-like, and is always asynchronous. A parent waiting for an output must be given something to render in the meantime. Using RenderingT is a great idiom for a one-and-done workflow tasked with providing a single product.

rjrjr:
Closes #1251.

### DIFF
--- a/docs/userguide/core-patterns.md
+++ b/docs/userguide/core-patterns.md
@@ -34,6 +34,16 @@ simply by rendering various children. It may just combine the renderings of mult
 use its props to determine which of a set of children to render. Such workflows can often be
 stateless.
 
+## One-and-done Workflows (RenderingT v. OutputT)
+
+A common question is “why can’t I emit output from `initialState`,” or “what if my Workflow realizes it doesn’t actually need to run? The most efficient, and most expressive, way to handle this is to use an optional or conditional `Rendering` type, and an `Output` of [`Never`](https://nshipster.com/never/)/[`Nothing`](https://medium.com/@agrawalsuneet/the-nothing-type-kotlin-2e7df43b0111).
+
+Imagine a `PromptForPermissionMaybeWorkflow`, that renders a UI to get a passcode, but only if that permission has not already been granted. If you make its `RenderingT` nullable (e.g. `Screen?`), it can return `null` to indicate that its job is done. Its callers will be synchronously informed that the coast is clear, and can immediately render what they actually care about.
+
+Another variation of this pattern is to use a sealed class / enum type for `Rendering`, with a `Working` type that implements `Screen`, and a unviewable `Finished` type that carries the work product.
+
+A good rule of thumb for choosing between using `Rendering` or `Output` is to remember that `Output` is event-like, and is always asynchronous. A parent waiting for an output must be given something to render in the meantime. Using `Rendering` is a great idiom for a one-and-done workflow tasked with providing a single product, especially one that might be available instantly.
+
 ## Props values v. Injected Dependencies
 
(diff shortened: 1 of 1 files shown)


docs/userguide/core-patterns.md
```diff
@@ -34,6 +34,16 @@ simply by rendering various children. It may just combine the renderings of mult
 use its props to determine which of a set of children to render. Such workflows can often be
 stateless.
 
+## One-and-done Workflows (RenderingT v. OutputT)
+
+A common question is “why can’t I emit output from `initialState`,” or “what if my Workflow realizes it doesn’t actually need to run? The most efficient, and most expressive, way to handle this is to use an optional or conditional `Rendering` type, and an `Output` of [`Never`](https://nshipster.com/never/)/[`Nothing`](https://medium.com/@agrawalsuneet/the-nothing-type-kotlin-2e7df43b0111).
+
+Imagine a `PromptForPermissionMaybeWorkflow`, that renders a UI to get a passcode, but only if that permission has not already been granted. If you make its `RenderingT` nullable (e.g. `Screen?`), it can return `null` to indicate that its job is done. Its callers will be synchronously informed that the coast is clear, and can immediately render what they actually care about.
+
+Another variation of this pattern is to use a sealed class / enum type for `Rendering`, with a `Working` type that implements `Screen`, and a unviewable `Finished` type that carries the work product.
```
rjrjr, line 43: @wardellbagby I don't see how you do this in a typesafe way without using a marker type.

RBusarow (approved):

docs/userguide/core-patterns.md
wardellbagby: You do, but it's on a Workflow by Workflow basis instead. Let them have their own distinct rendering types.

rjrjr: Without a common type for "this is something that can become a UI" used across the entire tree, it's a lot harder to pull off.

wardellbagby: To me, whether a rendering is UI or data is only important to the parent Workflow. It's the job of the root Workflow Runner to map renderings to the actual UI system of the platform, so only that Workflow Runner needs to care. Which means it's the job of the root Workflow to turn its child renderings into whatever the root Workflow Runner needs.

I imagine a root Workflow with a LayeredScreen rendering type, running applets that render their own specific AppletRenderings that run their own Workflows with their own rendering types. And a bunch of mapping to make it all work.


Big whirls, little whirls, viscosity, yada yada yada. 

Honestly, it's a lot of work just for the idea of making as many "pure" Workflows as possible that provide opaque renderings as much as possible. I'm not even convinced myself that that's a worthy goal to reach for.

PR #1253, opened 2022-05-18
UI Concepts pt. 1

rjrjr:
"Separation of Concerns," "What's a Screen?" and a TBD for "Composition and Navigation" that at least links to something.

Also deletes `comparison.md` because it would be redundant with existing FAQ entries, unnecessary.

### DIFF
--- a/docs/userguide/ui-concepts.md
+++ b/docs/userguide/ui-concepts.md
@@ -1,3 +1,150 @@
 # Workflow UI
 
-_Coming soon!_
+This page provides a high level overview of Workflow UI, the companion that allows [Workflow Core](../core-workflow) to drive Android and iOS apps.
+
+## Separation of Concerns
+
+Workflow maintains a rigid separation between its core runtime and its UI support.
+The [Workflow Core](../core-workflow) modules are strictly Swift and Kotlin, with no dependencies on any UI framework.
+Dependencies on Android and iOS are restricted to the Workflow UI modules, as you would expect.
+This innate separation naturally puts developers on a path to avoid entangling view concerns with their app logic.
+
+And note that we say "app logic" rather than "business logic."
+In any interesting app, the code that manages navigation and other UI-releated behavior is likely to dwarf that for what we typically think of as model concerns, in both size and complexity.
+
+We're all pretty good at capturing business concerns in tidy OO models of items for sale, shopping carts, payment cards and the like, nicely decoupled from the UI world.
+But the rest of the app, and in particular the bits about how our users navigate it?
+Traditionally it's hard to keep that app-specific logic centralized, so that you can see what's going on; and even harder to keep it decoupled from your view system, so that it's easy to test.
+The strict divide between Workflow UI and Workflow Core leads you to maintain that separation by accident.
+
+## What's a Screen?
+
+Most Workflow implementations produce `struct` / `data class` [renderings](../core-workflow#rendering) that can serve as view models.
+Such a rendering provides enough data to paint a UI, and functions to be called in response to UI events.
+
+These view model renderings implement the `Screen` [protocol](https://github.com/square/workflow-swift/blob/main/WorkflowUI/Sources/Screen/Screen.swift) / [interface](https://github.com/square/workflow-kotlin/blob/main/workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/Screen.kt) to advertise that this is their intended use.
+The core service provided by Workflow UI is to transform `Screen` types into platform-specific view objects, and to keep those views updated as new `Screen` renderings are emitted.
+
+`Screen` is the lynch pin that ties the Workflow Core and Workflow UI worlds together, the basic UI building block for Workflow-driven apps.
+A `Screen` is an object that can be presented as a basic 2D UI box, like an `android.view.View` or a `UIViewController`.
+And Workflow UI provides the glue that allows you to declare (at compile time!) that instances of `FooScreen : Screen` are used to drive `FooViewController`, `layout/foo_screen.xml`, or `@Composable fun Content(FooScreen, ViewEnvironment)`.
+
+=== "iOS"
+    ```Swift
+    struct WelcomeScreen: Screen {
+        var name: String
+        var onNameChanged: (String) -> Void
+        var onLoginTapped: () -> Void
+
+        func viewControllerDescription(environment: ViewEnvironment) -> ViewControllerDescription {
+            return WelcomeViewController.description(for: self, environment: environment)
+        }
+    }
+
+    private final class WelcomeViewController: ScreenViewController<WelcomeScreen> {
+        // ...
+    ```
+
+=== "Android Classic"
+    ```Kotlin
+    data class HelloScreen(
+      val message: String,
+      val onClick: () -> Unit
+    ) : AndroidScreen<HelloScreen> {
+      override val viewFactory: ScreenViewFactory<HelloScreen> =
+        fromViewBinding(HelloViewBinding::inflate) { r, _ ->
+          helloMessage.text = r.message
+          helloMessage.setOnClickListener { r.onClick() }
+        }
+    }
+    ```
+
+=== "Android Compose"
+    ```Kotlin
+    data class HelloScreen(
+      val message: String,
+      val onClick: () -> Unit
+    ) : AndroidScreen<HelloScreen> by ComposeScreen {
+      Button(onClick = onClick)
+      Text(message)
+    }
+    ```
+
+After all the chest-thumping above about [Separation of Concerns](#separation-of-concerns), these code samples probably don't look very separate. That's because, while the libraries themselves are completely decoupled, they don't force that strict rigor on your app code.
+
+If you aren't building, say, a core Workflow module that you want to ship separately from its Android and command line interfaces, you'd probably gain nothing from enforced separation but boilerplate and runtime errors.
+And in practice, your Workflow unit tests won't call `viewFactory` and will build and run just fine against the JVM.
+Likewise, at this point we've been building apps this way for hundreds of engineering years, and so far no one has called `viewControllerDescription()` and stashed a `UIViewController` in their workflow state.
+(This is not a challenge.)
+
+If you're one of the few who truly do need impermeable boundaries between your core and UI modules, they aren't hard to get.
+Your `Screen` implementations can be defined completely separately from their view code and bound later.
+
+=== "iOS"
+    ```Swift
+    // WelcomeScreen.swift
+    struct WelcomeScreen {
+        var name: String
+        var onNameChanged: (String) -> Void
+        var onLoginTapped: () -> Void
+    }
+
+    // WelcomeViewController.swift
+    extension WelcomeScreen: Screen {
+        func viewControllerDescription(environment: ViewEnvironment) -> ViewControllerDescription {
+            return WelcomeViewController.description(for: self, environment: environment)
+        }
+    }
+
+    private final class WelcomeViewController: ScreenViewController<WelcomeScreen> {
+        // ...
+    ```
+
+=== "Android"
+    ```Kotlin
+    // HelloScreen.kt
+    data class HelloScreen(
+      val message: String,
+      val onClick: () -> Unit
+    ) : Screen
+
+    // HelloScreenViewFactory.kt
+    object HelloScreenViewFactory: ScreenViewFactory<HelloScreen>
+    by ScreenViewFactory.fromViewBinding(HelloViewBinding::inflate) { r, _ ->
+          helloMessage.text = r.message
+          helloMessage.setOnClickListener { r.onClick() }
+        }
+    }
+
+    // HelloActivity.kt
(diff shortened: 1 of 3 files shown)


docs/userguide/ui-concepts.md
```diff
@@ -1,3 +1,151 @@
 # Workflow UI
 
-_Coming soon!_
+This page provides a high level overview of Workflow UI, the companion that allows [Workflow Core](../core-workflow) to drive Android and iOS apps.
+
+## Separation of Concerns
+
+Workflow maintains a rigid separation between its core runtime and its UI support.
+The [Workflow Core](../core-workflow) modules are strictly Swift and Kotlin, with no dependencies on any UI framework.
+Those dependencies are restricted to the Workflow UI modules, as you would expect, with Swift UI relying on iOS, and Kotlin UI on Android.
```
0legg, line 9: `Swift UI` could be mistaken for `SwiftUI` — Apple's native declarative UI framework (similar to Compose, or Blueprint).

docs/userguide/ui-concepts.md
```diff
@@ -1,3 +1,151 @@
...
+
+## What's a Screen?
+
+Most Workflow implementations produce pure-Swift / pure-Kotlin renderings that can serve as view models.
+Such a rendering provides enough data to paint a UI, and functions to be called in response to UI events.
+The core service provided by Workflow UI is to transform these renderings into platform-specific UI structures, and to keep those views updated as new renderings are emitted.
+
+The `Screen` [protocol](https://github.com/square/workflow-swift/blob/main/WorkflowUI/Sources/Screen/Screen.swift) / [interface](https://github.com/square/workflow-kotlin/blob/main/workflow-ui/core-common/src/main/java/com/squareup/workflow1/ui/Screen.kt) is the lynch pin that ties the two worlds together, and the basic UI building block for Workflow-driven apps.
+A screen is an object that can be presented as a basic 2D UI box, like an `android.view.View` or an `NSViewController`. And Workflow UI provides the glue that allows you to declare (at compile time!) that instances of `FooScreen : Screen` are used to drive `FooViewController`, `layout/foo_screen.xml`, or `@Composable fun Content(FooScreen, ViewEnvironment)`.
```
rjrjr, line 33: Oops, `UIViewController`, my NeXTstep is showing.

docs/userguide/ui-concepts.md
```diff
@@ -1,3 +1,151 @@
...
+    ) : AndroidScreen<HelloScreen> by ComposeScreen {
+      Button(onClick = onClick)
+      Text("message")
+    }
+    ```
+
+After all the chest-thumping above about [Separation of Concerns](#separation-of-concerns), these code samples probably don't look very separate. That's because, while the libraries themselves are completely decoupled, they don't force that strict rigor on your app code.
+
+If you aren't building, say, a core workflow module that you want to ship separately from its Android app, and an alternative command line interface, you'd probably gain nothing from enforced separation but boilerplate and runtime errors.
+And in practice, your Workflow unit tests won't call `viewControllerDescription()` or `viewFactory` and will build and run just fine against the JVM.
```
rjrjr, line 79:
```suggestion
And in practice, your Workflow unit tests won't call `viewFactory` and will build and run just fine against the JVM. Likewise, at this point we've been building apps this way for hundreds of engineering years, and so far no one has called `viewControllerDescription()` and stashed a `UIViewController` in their workflow state. (This is not a challenge.)
```

docs/userguide/ui-concepts.md
```diff
@@ -1,3 +1,151 @@
...
+
+If you're one of the few who really and trully do need impermeable boundaries between your core and UI modules, that's not hard to get. Your `Screen` implementations can be defined completely separately from their view code and bound later.
```
rjrjr, line 81:
```suggestion
If you're one of the few who really and truly do need impermeable boundaries between your core and UI modules, they aren't hard to get. Your `Screen` implementations can be defined completely separately from their view code and bound later.
```

docs/userguide/ui-concepts.md
```diff
@@ -1,3 +1,151 @@
...
+And note that we say "app logic" rather than "business logic."
+In any interesting app, the code that manages navigation and other UI-releated behavior is likely to dwarf that for what we typically think of as model concerns, in both size and complexity. 
+
+We're all pretty good at capturing business concerns in tidy OO models of items for sale, shopping carts, payment cards and the like, nicely decoupled from the UI world.
+But the rest of the app, and in particular the bits about how our users navigate it? Traditionally it's hard to keep that app logic code centralized, so that you can see what's going on; and even harder to keep it decoupled from your view system, so that it's easy to test.
+The strict divide between Workflow UI and the runtime makes it easy.
+
+## What's a Screen?
+
+Most Workflow implementations produce pure-Swift / pure-Kotlin renderings that can serve as view models.
```
0legg, line 28: What's a Rendering? Like, it'll be awesome to see how rendering and screen relate to each other, and what is the right way to call things depending on context.

docs/userguide/ui-concepts.md
rjrjr: Yikes, thanks.

docs/userguide/ui-concepts.md
rjrjr: How about if I link to the glossary entry for that? Aaand of course there are no automatic bookmarks for DL entries, yay.

https://square.github.io/workflow/glossary/#:~:text=is%20strongly%20discouraged.-,Rendering,-The%20externally%20available

rjrjr: Also note that rendering is already described in some depth in the core docs. Maybe should link to that instead? 

https://square.github.io/workflow/userguide/core-workflow/#workflows-produce-an-external-representation-of-their-state-via-rendering

0legg: Sounds good, and we could explain that Screen could be a derivative of Rendering, tailored to be responsible for UI representation and interaction (while Rendering itself is kind of a public snapshot of the state; but snapshot, and not The Snapshot). Damn, workflow terminology already employed every possible term I could think of :)

docs/userguide/ui-concepts.md
```diff
@@ -1,3 +1,151 @@
...
+## Separation of Concerns
+
+Workflow maintains a rigid separation between its core runtime and its UI support.
+The [Workflow Core](../core-workflow) modules are strictly Swift and Kotlin, with no dependencies on any UI framework.
+Those dependencies are restricted to the Workflow UI modules, as you would expect, with Swift UI relying on iOS, and Kotlin UI on Android.
+This innate separation naturally puts developers on a path to avoid entangling view concerns with their app logic.
+
+
+!!! info "Separate both ways"
+    This strictness goes in both directions: neither UI library depends upon either runtime. They could easily be driven by any other stream of view models, and in view system tests they often are.
```
dhavalshreyas, line 14: I'm not sure if we've honored this on the Swift side. We have an explicit [dependency](https://github.com/square/workflow-swift/blob/main/WorkflowUI.podspec#L21) on `Workflow`.

dhavalshreyas (approved):
This is great! Thanks @rjrjr!

docs/userguide/ui-concepts.md
rjrjr: Oh, that's too bad. Any idea why?

docs/userguide/ui-concepts.md
```diff
@@ -1,3 +1,151 @@
...
+    // WelcomeViewController.swift
+    extension WelcomeScreen: Screen {
+        func viewControllerDescription(environment: ViewEnvironment) -> ViewControllerDescription {
+            return WelcomeViewController.description(for: self, environment: environment)
+        }
+    }
+
+    private final class WelcomeViewController: ScreenViewController<WelcomeScreen> {
+        // ...
+    ```
```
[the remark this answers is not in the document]
mjohnson12: You can declare the struct with it implementing Screen and provide the implementation in an extension but it would have to be in the same swift module.

docs/userguide/ui-concepts.md
```diff
@@ -1,3 +1,151 @@
...
+          workflow = HelloWorkflow.mapRenderings { it.withRegistry(viewRegistry) },
+          scope = viewModelScope,
+          savedStateHandle = savedState
+        )
+      }
+    }
+    ```
+
+!!! faq "Why \"Screen\"?"
+    We chose the name "Screen" because "View" would invite confusion with the like-named Android and iOS classes, and because "Box" didn't occur to us.
```
jbmorgan, line 146: Is it worth calling out that this is also unrelated to UIScreen, which is UIKit's representation of a hardware display?

docs/userguide/ui-concepts.md
rjrjr:
```
Dependencies on Android and iOS are restricted to the Workflow UI modules, as you would expect.
```

docs/userguide/ui-concepts.md
jbmorgan: Take this with a grain of salt as I'm still very new to the code base, but the dependency was introduced in 2de5e98c2 and it looks like it's primarily because ContainerViewController imports Workflow and references WorkflowHost, AnyWorkflowConvertible, etc.

docs/userguide/ui-concepts.md
rjrjr: I'm wary of going too far into the weeds here.

docs/userguide/ui-concepts.md
```diff
@@ -1,3 +1,151 @@
...
+
+We're all pretty good at capturing business concerns in tidy OO models of items for sale, shopping carts, payment cards and the like, nicely decoupled from the UI world.
+But the rest of the app, and in particular the bits about how our users navigate it? Traditionally it's hard to keep that app logic code centralized, so that you can see what's going on; and even harder to keep it decoupled from your view system, so that it's easy to test.
+The strict divide between Workflow UI and the runtime makes it easy.
+
+## What's a Screen?
+
+Most Workflow implementations produce pure-Swift / pure-Kotlin renderings that can serve as view models.
+Such a rendering provides enough data to paint a UI, and functions to be called in response to UI events.
+The core service provided by Workflow UI is to transform these renderings into platform-specific UI structures, and to keep those views updated as new renderings are emitted.
```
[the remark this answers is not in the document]
rjrjr: I was quietly inconsistent on purpose, trying to imply that UI structures are obviously lower-case-V views. Too distracting?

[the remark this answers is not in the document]
rjrjr: Too distracting.

```
The core service provided by Workflow UI is to transform `Screen` types into platform-specific view objects, and to keep those views updated as new `Screen` renderings are emitted.
```

docs/userguide/ui-concepts.md
```diff
@@ -1,3 +1,151 @@
...
+    ```
+
+=== "Android Compose"
+    ```Kotlin
+    data class HelloScreen(
+      val message: String,
+      val onClick: () -> Unit
+    ) : AndroidScreen<HelloScreen> by ComposeScreen {
+      Button(onClick = onClick)
+      Text("message")
```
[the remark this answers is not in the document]
rjrjr: Thanks

docs/userguide/ui-concepts.md
rjrjr: Thanks for the spelunking. I just deleted this block, it didn't seem important.

docs/userguide/ui-concepts.md
rjrjr: Missed this before I merged. Maybe add this?

> (No one seems to have been confused by the fact that `Screen` and iOS's `UIScreen` are unrelated.)

jbmorgan: Yeah, I think so. Thanks!

docs/userguide/ui-concepts.md
```diff
@@ -1,3 +1,150 @@
...
+
+Workflow maintains a rigid separation between its core runtime and its UI support.
+The [Workflow Core](../core-workflow) modules are strictly Swift and Kotlin, with no dependencies on any UI framework.
+Dependencies on Android and iOS are restricted to the Workflow UI modules, as you would expect.
+This innate separation naturally puts developers on a path to avoid entangling view concerns with their app logic.
+
+And note that we say "app logic" rather than "business logic."
+In any interesting app, the code that manages navigation and other UI-releated behavior is likely to dwarf that for what we typically think of as model concerns, in both size and complexity.
+
+We're all pretty good at capturing business concerns in tidy OO models of items for sale, shopping carts, payment cards and the like, nicely decoupled from the UI world.
```
RBusarow, line 15: When I read this, I felt that 'OO' kind of interrupted the flow.

```suggestion
We're all pretty good at capturing business concerns in tidy object-oriented models of items for sale, shopping carts, payment cards and the like, nicely decoupled from the UI world.
```

jbmorgan: Same, good change I think.

rjrjr: Nice, will do this in the next update.

PR #1254, opened 2022-05-18
Another Glossary overhaul, no more HTML

rjrjr:
MarkDown hates &lt;dl> lists, so stop trying to use them. Now the glossary entries are headings, and we're able to use normal markdown everywhere else. Biggest benefit, and the thing that drove me to do this, is that we can link directly to each entry this way.

Also adds some intra-glossary links. Did not go too crazy, e.g. State and Output get mentioned so many times that linking each reference seemed silly.

### DIFF
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -147,7 +147,7 @@ The type of the object that can optionally be delivered to the Workflow’s pare
 
 ## Child Workflow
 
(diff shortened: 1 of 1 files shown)


rjrjr:
Lint is upset, doesn't like me skipping H levels. Inclined to disable that check.

rjrjr:
Eh, H2 for the entries looks okay.

jbmorgan (approved):

PR #1257, opened 2022-05-25
Merges concepts.md and core-workflow.md.

rjrjr:
Deletes stale bits in the process. Also some file renames to keep up w/title changes. And updates all the svg images to have white backgrounds, because geeks tend to force their browsers into dark mode even though it never, ever works.


zach-klippenstein:
It's me. I'm geeks.

zach-klippenstein (approved):

PR #1266, opened 2023-06-12
Fix incorrect workflow parent name used in `Workflow Core` documentation

sssurvey:
Within the `Workflow Core -> Composing Workflows` [documentation](https://square.github.io/workflow/userguide/concepts/#composing-workflows), a minor error can be noticed in the last paragraph. It appears that an incorrect class name has been used to refer to the parent workflow, which is contrary to the documentation context and understanding. Based on my understanding, the parent workflow should be correctly identified as `EmailBrowserWorkflow`, rather than `EmailInboxWorkflow`. 

The paragraph I am referring to is:
> Whenever such a [Workflow Action cascade](https://square.github.io/workflow/glossary#action-cascade) fires, the root Workflow is asked for a new Rendering. Just as before, EmailInboxWorkflow delegates to its two children for their Renderings, this time providing the new value of selection as the updated Props for MessageWorkflow.


rjrjr (approved):
Thanks!

And please stay tuned -- planning a significant update to the tutorials in a few weeks.

sssurvey:
No problem, 👀 out for the update!

PR #1273, opened 2025-07-25
Skippable instead of One and Done

rjrjr:
The One and Done entry was bad advice, or at least incomplete. Updated to describe the Skippable Workflow pattern we've been using with a lot of success lately.


docs/userguide/common-patterns.md
```diff
@@ -34,15 +34,59 @@ simply by rendering various children. It may just combine the renderings of mult
...
+The idea is that a workflow that performs work that may not actually be necessary
+provides an extra function for parents to call.
+The parent can call this during the upstream action
+where it's deciding whether or not to enter the state
+in which the child would be run.
+
+Perhaps it's a simple boolean function indicating whether or not the child actually has any work to do:
+
+```kotlin
+state = if (stepTwoWorkflow.isThisNecessary()) RunStepTwo else RunStepThree
```
amayers, line 54: This only works if the `stepTwoWorkflow` is created and stored in the `state` a head of time, before you get to the `apply(action)` method. If your `stepTwoWorkflow` can't be created until you get into the `apply(action)` with a specific action/state, then you are out of luck.

rjrjr: Typically in Kotlin child workflows are DI injected (available as constructor parameters fo the parent), and actions are inner objects of the parent workflow with access to its injected dependencies.

In Swift, didn't a change just land to make the parent workflow available from the action? cc @jamieQ

steve-the-edwards (approved):

PR #1276, opened 2026-05-05
Fix Kotlin docs output path

steve-the-edwards:
🤖 Fix the Kotlin docs publish workflow after `workflow-kotlin`'s `siteDokka` output moved from `build/dokka/htmlMultiModule` to `build/dokka/workflow`.

The v1.27.0 docs publish failed after `siteDokka` completed because the workflow tried to move the old `htmlMultiModule` directory. This updates both the GitHub Actions workflow and the older `deploy_website.sh` path to use the current `build/dokka/workflow` output.

Validation:
- `zsh -n deploy_website.sh`
- parsed `.github/workflows/update-docs.yml` with Ruby YAML
- `git diff --check`

### DIFF
--- a/.github/workflows/update-docs.yml
+++ b/.github/workflows/update-docs.yml
@@ -59,7 +59,7 @@ jobs:
         ./gradlew siteDokka --build-cache --quiet
 
         mkdir -p ../workflow/docs/kotlin/api
-        mv build/dokka/htmlMultiModule ../workflow/docs/kotlin/api
+        mv build/dokka/workflow ../workflow/docs/kotlin/api
 
     # Generate the mkdocs site
     - name: Generate site with mkdocs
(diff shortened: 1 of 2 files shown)


.github/workflows/update-docs.yml
```diff
@@ -59,7 +59,15 @@ jobs:
         ./gradlew siteDokka --build-cache --quiet
 
         mkdir -p ../workflow/docs/kotlin/api
-        mv build/dokka/htmlMultiModule ../workflow/docs/kotlin/api
+        if [ -d build/dokka/workflow ]; then
+          mv build/dokka/workflow ../workflow/docs/kotlin/api
+        elif [ -d build/dokka/htmlMultiModule ]; then
```
zach-klippenstein: Do we still need both these branches?

zach-klippenstein (approved):
Do we know why this directory changed?

steve-the-edwards:
> Do we know why this directory changed?

It's a dokka default change. I don't think we need the other branches though, will remove.

